PR #63, opened 2020-10-19
✨ @percy/sdk-utils

wwilsman:
## What is this?

Node based SDKs have a few common utils that can be abstracted away into their own module. They also perform common setup steps in tests that can also be abstracted away into a common test helper. 

These common utils include talking to the local running Percy CLI API. Rather than have dependencies on libraries like fetch, we can use a shared request helper using native Node methods. Also, rather than including all of the `winston` dependencies of `@percy/logger`, there is a smaller, simple log util that accomplishes the goal of communicating SDK information to the user. 

The consumable utils exported by this package are `log`, `getInfo`, `isPercyEnabled`, `fetchPercyDOM`, and `postSnapshot`. [Their usage docs can be found in the readme](https://github.com/percy/cli/blob/1eedb03c72cedd2263e51d1621f8be2ac82b84bc/packages/sdk-utils/README.md)

### Other changes

I added an argument to the catch clause in the server helper which is eventually required by the new sdk-utils test helper. While testing these utils, I found that some external libraries (like `testcafe`) will attempt to compile required files even if they are compatible with the current version of Node. That compilation failed when it encountered a catch without arguments.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.0-beta.18",
+  "version": "1.0.0-beta.19",
   "packages": [
     "packages/*"
   ],
--- a/packages/cli-build/package.json
+++ b/packages/cli-build/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@percy/cli-build",
-  "version": "1.0.0-beta.18",
+  "version": "1.0.0-beta.19",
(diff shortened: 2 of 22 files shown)


Robdel12 (approved):
🏁 Love this, shrinks the JS SDKs to like 40 lines or less.

PR #75, opened 2020-11-06
🐛 Fix config merging

wwilsman:
## Purpose

Some aspects of config merging was broken due to either frozen objects, or strange rules around empty objects.

Additionally, the `config:validate` command did not validate core configuration options because the schema is not yet loaded when that command is run.

## Approach

Remove the deepmerge dependency which seemingly will override objects when merging an empty object. This functionality was easy to replace by utilizing the existing `normalize` util's traversal loop to override values while normalizing them. The core loop was refactored out of the normalize function and is also utilized by the `getDefaults()` function to merge overrides while replacing default arrays (rather than concat arrays with the normal merge behavior).

To validate core configuration options with `config:validate`, the core config schema can be added within a hook similar to how snapshot and upload config options are added.

### DIFF
--- a/packages/cli-config/package.json
+++ b/packages/cli-config/package.json
@@ -24,6 +24,9 @@
   "oclif": {
     "bin": "percy",
     "commands": "./dist/commands",
+    "hooks": {
+      "init": "./dist/hooks/init"
+    },
     "topics": {
       "config": {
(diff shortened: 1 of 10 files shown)


Robdel12 (approved):
🏁 Very nice

PR #76, opened 2020-11-06
✨ Retry specific request error codes

wwilsman:
## What is this?

The old `percy-js/client` library had retry handlers for specific request error codes. Unsure how/why they were dropped during the transition, but this PR adds them back with a `shouldRetryRequest` helper that is utilized within request error handlers.


Robdel12 (approved):
🏁

PR #81, opened 2020-11-09
🐛 Start server after launching a browser and creating a build

wwilsman:
## Purpose

When using the `exec:start` command it downloads a browser if needed just like the `exec` command. However, the `exec` command waits for the start process to resolve directly before calling the provided command while the user/SDK is left to check this themselves when using `exec:start`. The healthcheck endpoint resolves when the server is listening, but the server is actually listening before a browser is downloaded and the process is actually ready.

## Approach

Move the server start process to after the browser launch and build creation processes. This ensures that the server, and therefore the healthcheck endpoint, will only resolve when actually ready.

To test this, the various methods are stubbed to capture a timestamp and the resulting timestamps are compared after starting the whole process.

### DIFF
--- a/packages/core/src/percy.js
+++ b/packages/core/src/percy.js
@@ -90,12 +90,12 @@ export default class Percy {
     this.client.getToken();
 
     try {
-      // if there is a server, start listening
-      await this.server?.listen(this.port);
-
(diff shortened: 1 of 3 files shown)


Robdel12 (approved):
🏁

PR #86, opened 2020-11-17
♻ Remove path from cli api address

wwilsman:
# What is this?

Removing the path from the address will make it less error-prone to customize by not requiring that the env var ends in `/percy`


Robdel12 (approved):
🏁

PR #87, opened 2020-11-19
🐛 Fix CSSOM check

wwilsman:
## What is this?

Sometimes `innerText` can be undefined which causes this function to error when it should report true. This check is present in the old serialization script in agent, but it couldn't be confirmed if it was needed since I'm not quite sure how to replicate that behavior to test it.


Robdel12 (approved):
🏁 

Just curious -- what stage is that operator?

Robdel12:
Oh nice, most browsers shipped: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#Browser_compatibility assuming stage 3+ then

wwilsman:
Stage 4, has been shipped to the major browsers already and node 14. But we transpile with babel anyway.

PR #128, opened 2021-01-05
🔥 Remove dependency on Puppeteer

wwilsman:
## Purpose

Percy agent has had a history of troublesome Puppeteer updates that has caused things to break due to introducing either unintentional breaking changes, bugs associated with Puppeteer's version of Chromium, or other factors. In addition, some issues faced by users can be hard to track down because of vague protocol errors bubbling through Puppeteer from Chromium.

This CLI moved to the `puppeteer-core` package to avoid the automatic Chromium install. However, Puppeteer environment variables can still influence Puppeteer's browser fetcher the CLI uses to lazily download Chromium. If decoupled further, the CLI would then only use Puppeteer, an entire browser automation library, to intercept network requests.

## Approach

Use the Chrome DevTools Protocol directly! Using the protocol directly and decoupling from Puppeteer allows us to have a much finer control and insight into the asset discovery process. This is achieved by launching Chromium and watching its debug output for the protocol WebSocket address, similar to puppeteer. Once a socket is connected, we can then issue commands and react to events using the various protocol domains: https://chromedevtools.github.io/devtools-protocol/

The browser download utility was updated to download a version of Chromium directly from a URL as oppose to using Puppeteer's fetcher. By default, it is pinned to a revision of Chromium matching the version used by Percy within our own infrastructure. This revision can be updated alongside the infrastructure's browser version to keep inconsistencies to a minimum.

Besides protocol methods to set page options and navigate to a URL, the network protocol events are utilized directly for asset discovery as oppose to going through several other emitters within Puppeteer's internals. The new `Network` class that coordinates requests was heavily inspired by Puppeteer's own `NetworkManager` class with changes to better fit our use case (rather than being generic for broader automations).

Together, the `Browser`, `Page`, and `Network` classes replace Puppeteer entirely. The `Browser` class is used to launch the browser, connect to it, manage page sessions, and handle clean up of the browser process. The `Page` class represents a page session and is used to drive a single page via protocol commands. It has some helper methods for the discoverer class and capture method to set common options and navigate to URLs. The `Network` class tracks all requests to determine if the network is idle and, when request interception is enabled, it is used to capture request responses for asset discovery.

**@percy/core breaking change**: the `Percy#capture()` method was built around Puppeteer's Page class and its ability to inject scripts and interact with the page via an `execute` option. Rather than exposing our own internal Page class for users to interact with, the `execute` option has changed to evaluate JavaScript within the page's execution context. This is similar to Puppeteer's `Page#evaluate` method. If a user needs to do something more advanced than executing scripts, they can use a more full-fledged SDK with a proper browser automation library.

### Other Changes

**@percy/core breaking change**: the main export was made to be a commonjs module so that `require('@percy/core')` did not need to spread the `default` keyword in order to be used. This should only effect commonjs users importing `@percy/core` directly.

Some options such as `snapshot.requestHeaders` and `discovery.concurrency` were removed from the default config output when creating new config files. I believe only common options should be included by default while other options can be discovered via documentation. Discovery launch options are also now included in types and schema since those are directly in our control as opposed to being passed along through to Puppeteer.

Some files were moved around such as `percy-css.js` moving to `/utils` and `discoverer.js` moving to `/discovery` alongside the new classes. These file movements aren't well diffed and show as a lot of additions and deletions where there wouldn't be as large of a diff normally.

There are probably a few other small things I'm forgetting. I tried my best to leave detailed comments, so if there is something that is confusing or not well documented, let's document it better.


packages/core/src/config.js
```diff
@@ -19,8 +19,15 @@ export const schema = {
...
-        default: {}
+        additionalProperties: { type: 'string' }
+      },
+      authorization: {
+        type: 'object',
+        additionalProperties: false,
+        properties: {
+          username: { type: 'string' },
+          password: { type: 'string' }
+        }
```
wwilsman, line 30: Another exciting change! We could technically implement this before, but I did it here to bump coverage of the new `Network` class. This allows people to specify authentication credentials to use for basic auth requests. They could also do this with a request header before, but this explicitly allows username/password credentials to be used now.

packages/cli-snapshot/README.md
```diff
@@ -147,13 +147,14 @@ module.exports = [{
   waitForSelector: '.some-element',
   snapshots: [{
     name: 'Snapshot two - after click',
-    async execute(page) {
-      await page.click('.button')
+    execute() {
+      document.querySelector('.button').click();
```
Robdel12, line 151: Awh dang, but that's okay.

packages/core/README.md
```diff
@@ -111,9 +112,10 @@ await percy.stop()
 
 ## Advanced
 
-### Puppeteer Executable Path
+### Chromium Executable Path
 
-To avoid downloading the browser used for asset discovery, the local browser executable can be
-defined with an `executablePath` option provided within `discovery.launchOptions`. This options
-falls back to the `PUPPETEER_EXECUTABLE_PATH` environment variable if defined.
+**Use with caution as asset discovery may not work with some versions of Chromium.** To avoid
```
Robdel12, line 117: 🔥 I see what you mean. We can update this then: https://docs.percy.io/docs/skipping-puppeteer-chromium-download

packages/core/src/config.js
Robdel12: 🔥 This is going to make this much easier https://docs.percy.io/docs/capturing-assets-protected-with-authentication

packages/core/src/discovery/browser.js
```diff
@@ -0,0 +1,261 @@
...
+    '--disable-web-security',
+    '--force-color-profile=srgb',
+    '--metrics-recording-only',
+    '--no-first-run',
+    '--no-sandbox',
+    '--enable-automation',
+    '--password-store=basic',
+    '--use-mock-keychain',
+    '--remote-debugging-port=0'
+  ];
```
Robdel12, line 45: This come from somewhere? I feel a little uneasy about managing this (even though deep down I know this list exists in one of the deps lol)

packages/core/src/queue.js
```diff
@@ -27,7 +27,9 @@ export default class Queue {
 
   // Resolves when there are no more queued or pending tasks.
   idle() {
-    return idle(() => this.length);
+    return waitFor(() => this.length === 0, {
+      timeout: 2 * 60 * 1000 // 2 minutes
```
Robdel12, line 31: How did you settle on 2 mins?

packages/core/src/utils/assert.js
```diff
@@ -9,7 +9,6 @@ class PercyAssertionError extends Error {
     'is remote': () => 'Remote resource',
     'no response': () => 'No response',
     'too many bytes': ({ size }) => `Max file size exceeded [${size}]`,
-    'too many redirects': ({ length }) => `Too many redirects [${length}]`,
```
Robdel12, line 12: This not matter anymore? probably not, I think I remember you mentioning a nice way to pop the final redirect of the request?

Robdel12 (approved):
🏁 This is a big step forward for us. The logger PR (#140) that builds on top of this is going to really help us figure out weird issues in asset discovery (as well as have better control over asset discovery).

🔥 work, yo.

packages/core/src/discovery/browser.js
wwilsman: This is coming from puppeteer. Although I would love to have a comment for each flag describing what their purpose is. I know of a handful, and we can do this for all of them in the future (or I can spend some time looking them up for this PR)

packages/core/src/queue.js
wwilsman: It was indefinite before and would just hang if a queued task never resolves for any reason. Figured 2 minutes was long enough to finish uploading any snapshots + resources that weren't uploaded during the rest of the run. We can adjust as needed, but this is really just a fallback in case a queued task hangs somewhere.

packages/core/src/utils/assert.js
wwilsman: Other way around actually. In handling the final request, [we can associate the response with the first request in the chain](https://github.com/percy/cli/pull/128/files#diff-47ebb78ad76028830835f9a5e76db43b8cde14c487e756ce6caa82a9a36b369dR174). So any number of redirects should work so long as the browser is fine with it.

commonwealthEnvoy:
Awesome PR!

Currently trying to install and use percy without puppeteer in my CI to keep things small and avoid a high severity puppeteer security issue. Percy agent [still depends on it](https://github.com/percy/percy-agent/blob/80a59f391bd37794f01c2844611ffb827f8d696c/package.json#L39) though. This also means that my CI breaks because the puppeteer@5.5.0 install fails with
<details><summary>Cannot find module '/root/.npm/_npx/16/lib/node_modules/percy/node_modules/puppeteer/install.js'</summary>
```
Error: Cannot find module '/root/.npm/_npx/16/lib/node_modules/percy/node_modules/puppeteer/install.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Function.Module.runMain (internal/modules/cjs/loader.js:831:12)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:623:3)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! puppeteer@5.5.0 install: `node install.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the puppeteer@5.5.0 install script.
```
</details>

Will this solve itself by percy-agent catching up eventually without anyone doing something or should they be informed?

For now I will try to work around this issue by using [my fork of percy agent](https://github.com/commonwealthEnvoy/percy-agent/pull/1) to simply remove puppeteer from the dependencies, add that fork to package.json and use the beta releases of `percy/core` and `percy/client` which have puppeteer removed already.

Robdel12:
Hey @commonwealthEnvoy! Percy CLI completely replaces percy agent. You can check this doc to see if your SDK has been updated to use Percy CLI: https://docs.percy.io/docs/migrating-to-percy-cli

PR #129, opened 2021-01-07
🐛 @percy/env parallel fixes

wwilsman:
## What is this?

This fixes a few minor issues with parallel total and nonce collection

Parallel total - `parseInt` does not throw but can return NaN, Infinity, or hex values; which should be checked and removed if present.

Parallel nonce - the API does not like it when a nonce is provided without a total, which happens by default when somebody has parallel CI jobs even without parallel Percy builds.

The total collection was moved before the nonce. If there is no total, do no collect the nonce. This will suppress parallel errors when the user is not providing the parallel total option. If they do supply a parallel total and the nonce is missing from their environment, the error will still correctly occur.

### DIFF
--- a/packages/env/test/defaults.test.js
+++ b/packages/env/test/defaults.test.js
@@ -81,7 +81,7 @@ describe('Defaults', () => {
     expect(env).toHaveProperty('ci', 'bitbucket');
     expect(env).toHaveProperty('commit', 'bitbucket-commit-sha');
     expect(env).toHaveProperty('git.sha', 'fully-valid-git-sha');
-    expect(env).toHaveProperty('parallel.nonce', 'bitbucket-build-number');
+    expect(env).toHaveProperty('parallel.nonce', null);
   });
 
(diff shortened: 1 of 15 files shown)


packages/env/test/defaults.test.js
```diff
@@ -132,6 +132,23 @@ describe('Defaults', () => {
...
+
+    expect(env).toHaveProperty('parallel.nonce', null);
+    expect(env).toHaveProperty('parallel.total', null);
+
+    env = new PercyEnvironment({
+      PERCY_PARALLEL_NONCE: 'percy-nonce'
+    });
+
+    expect(env).toHaveProperty('parallel.nonce', null);
+    expect(env).toHaveProperty('parallel.total', null);
```
Robdel12, line 149: This is 😍

Robdel12 (approved):
🏁 Really honing in our parallel / env collection here and I loveeeeeeeeee it. Nice to get this into a better spot

PR #130, opened 2021-01-07
✨ Export CLI commands

wwilsman:
## What is this?

This allows for invoking commands via Node scripts without needing to invoke a subprocess.

These are exported via the commonjs syntax to avoid having to destructure the `default` property when used without ES modules.


Robdel12 (approved):
Well that was easy, now folks can consume even more of the SDK tool chain programmatically 💯

PR #133, opened 2021-01-11
🐛 Use absolute path for image upload root URL

wwilsman:
## Purpose

The API expects resource URLs to either be a fully qualified URL or an absolute path and does not accept relative URLs.

## Approach

Add a leading `/` to the image upload's root URL and add an expectation for it in the existing success test.


Robdel12 (approved):
🏁 🏎️

PR #134, opened 2021-01-14
✨ Add backwards compatability for DOM library fetching

Robdel12:
## What is this?

This allows us to use the new CLI with any non-JS SDK _right now_. This is because the only breaking changes for those non-js SDKs was:

- Change in server path for where to get the DOM library
- Change in function signature for how the dom library works

This PR adds back the `percy-agent.js` server endpoint (with a deprecation log), and also adds a wrapper to match the old function signature.

JS SDKs will still need to be updated, since they carry `@percy/agent` as a dependency (and use the SDK util method to require the DOM JS)

### DIFF
--- a/packages/core/test/server.test.js
+++ b/packages/core/test/server.test.js
@@ -3,6 +3,7 @@ import fetch from 'node-fetch';
 import PercyConfig from '@percy/config';
 import Percy from '../src';
 import pkg from '../package.json';
+import { stdio } from './helpers';
 
 describe('Snapshot Server', () => {
   let percy;
@@ -15,6 +16,7 @@ describe('Snapshot Server', () => {
   });
 
   afterEach(async () => {
+    percy.loglevel('error');
     delete percy.stop; // remove own mocks
     await percy.stop();
   });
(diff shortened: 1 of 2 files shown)


packages/core/test/server.test.js
```diff
@@ -70,6 +72,21 @@ describe('Snapshot Server', () => {
     await expect(response.text()).resolves.toBe(bundle);
   });
 
+  it('serves the legacy percy-agent.js dom bundle', async () => {
+    let bundle = require('fs')
+      .readFileSync(require.resolve('@percy/dom'), { encoding: 'utf-8' })
+      .concat('(window.PercyAgent = class PercyAgent { snapshot(n, o) { return PercyDOM.serialize(o) } });');
+
+    percy.loglevel('warn');
+    await percy.start();
```
wwilsman, line 81: Just in case warnings ever come from start for whatever reason, these should be flipped to make sure we're only warning what we're expecting.
```suggestion
    await percy.start();
    percy.loglevel('warn');
```

wwilsman (changes requested):

wwilsman (approved):

PR #138, opened 2021-01-20
💥 Rename `--disable-asset-cache` flag to `--disable-cache`

Robdel12:
## What is this?

As we hone in on the final v1 of the CLI, we're working out the last breaking changes we want to live with. None of the other flags mention `asset`, so we decided to align this flag with the others (since it's already breaking for `@percy/agent`'s `--cache-responses` flag).


wwilsman (approved):
🎮

PR #140, opened 2021-01-21
✨ Even better logging

wwilsman:
## Purpose

With #128, we have much deeper control and insight into the asset discovery process. While the current logs are pretty good, this PR revamps the internal logger to make them even better so we can utilize them going forward in debugging customer issues.

While Winston is a great logger with _many_ features, we only use a few; and with those few, not to their full potential. For example, formatters and transports – we use a custom formatter to add a percy label and color log levels, and we two transports for console logging and file logging. However, we still patch a few methods to better fit our use cases.

## Approach

Replace Winston entirely with a custom logger. This custom logger follows a familiar loglevel style with `info`, `warn`, `error`, and `debug` log levels, but additionally follows a traditional Node debug style format that allows precise control of which debug logs to show. This enables us to sprinkle more and more logs throughout the codebase while maintaining and easily digestible, granular, format.

The new logger also drops the file logger completely and instead opts to retain in-memory logs so that they may be queried and uploaded later. If necessary, we can still save these full logs to a file using pipes which could then be inspected post-mortem.

The previous logger test helper captured all process IO writes which could cause the runner output itself to be swallowed if something went wrong during capture. The new logger test helper is linked directly to the logger's in-memory store which allows other processes to continue writing to stdio without interruption. This additionally required a slight modification of existing stdio writes to utilize the logger's stdio which is mocked under testing.

The `@percy/sdk-utils` package had it's own logger and log test helper to avoid including Winston's entire ecosystem in individual SDKs. Now that Winston and its dependencies are removed, we can include our own logger and its helper within that package. This does mean that consuming SDKs will need to be updated to the new logging style as well to be compatible with this beta.

As an experiment to test the value of these new logs, I tracked down and fixed a flakey test that consistently failed more often than not on my machine. It required a slight refactor of the capture method, but once additional logs were added, it became clear what the issue was and was substantially easier to fix.


Robdel12:
>  It required a slight refactor of the capture method, but once additional logs were added, it became clear what the issue was and was substantially easier to fix.

🔥 This is exactly what we've been chasing for _ages_. Can't wait

packages/core/src/browser.js
```diff
@@ -0,0 +1,307 @@
+import os from 'os';
```
Robdel12, line 1: 🔪

packages/logger/src/logger.js
```diff
@@ -0,0 +1,142 @@
...
+    return Object.keys(LOG_LEVELS)
+      .reduce((group, level) => Object.assign(group, {
+        [level]: this.log.bind(this, name, level)
+      }), {
+        deprecated: this.deprecated.bind(this, name)
+      });
+  }
+
+  // Ensures that deprecation messages are not logged more than once
+  deprecated(debug, message, meta) {
```
Robdel12, line 75: I love this so much.

packages/logger/src/logger.js
```diff
@@ -0,0 +1,142 @@
...
+    let error;
+
+    // message must be an error object
+    if (typeof message !== 'string' && (level === 'error' || level === 'debug')) {
+      error = message;
+      message = 'stack' in error ? error.stack : error.toString();
+    }
+
+    // timestamp each log
+    let timestamp = Date.now();
```
Robdel12, line 94: It's nice to see my cruft go away 😂

Robdel12 (changes requested):
This is doppe! Super excited for this. I noticed the README wasn't updated, so we'll want to do that for docs/usage/etc

Robdel12 (approved):
🏁

PR #145, opened 2021-01-23
✨ Lazily create sdk-util logger for testing purposes

wwilsman:
## What is this?

A logger is a singleton, but tests teardown and recreate that singleton for a clean environment. When a logging group is created, it is bound to the singleton instance at the time of creation. Creating a group at the root level, like in the sdk-utils, causes it to forever be bound to that initial instance and not be effected by test setup. By lazily creating the logging group, we can be assured that the test environment will be cleaned up without requiring a re-require of the file where the logger is created.

### DIFF
--- a/packages/sdk-utils/index.js
+++ b/packages/sdk-utils/index.js
@@ -1,5 +1,4 @@
 const logger = require('@percy/logger');
-const log = logger('util');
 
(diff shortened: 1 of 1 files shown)


Robdel12 (approved):
🏁

PR #149, opened 2021-01-26
✅ Fix @percy/dom Karma and Webpack issues

wwilsman:
## What is this

Currently, `@percy/dom` is compiled with Webpack. When tests run, a Karma plugin also compiles tests with Webpack. Since Webpack 5.0, some default node polyfills were removed causing test failures and we now have to polyfill them ourselves. Additionally, a recent Karma update prevents it from running when extraneous flags are provided (`--coverage`).

Karma config was updated to use an environment variable rather than rely on the extra flag. Since we have to polyfill node modules anyway, Webpack was removed in favor of Rollup which produces smaller, more readable bundles and is preferred for bundling libraries whereas Webpack is preferred for bundling applications.

Rather than fighting with `expect` and `cheerio` to work in-browser, small alternatives are used in `@percy/dom` tests. `cheerio` is replaced with a simple native DOM parsing script, and `expect` is replaced with a thin polyfill.

### DIFF
--- a/packages/dom/karma.conf.js
+++ b/packages/dom/karma.conf.js
@@ -13,11 +13,15 @@ module.exports = config => {
     ],
 
     files: [
-      { pattern: 'test/index.js', watched: false }
+      { pattern: 'src/index.js', watched: false },
+      { pattern: 'test/helpers.js', watched: false },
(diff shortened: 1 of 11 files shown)


Robdel12 (approved):
🏁 So long, partner (Webpack)

PR #150, opened 2021-01-26
♻ Adjust how the browser download script determines executable path

wwilsman:
## What is this?

This takes the guesswork out of determining the browser executable path. The current guesswork is accurate for Chromium, but when this script is utilized to download other browsers, the executable location can be difficult to determine based on the filename alone. For example, the archive for Firefox on Linux is `firefox 73.0.tar.bz2` and when getting the executable path it guesses `firefox 73.0.tar/firefox` even though the actual archive unzips to `firefox/firefox`.

By removing the guesswork and being explicit, we can avoid OS/Browser archive name inconsistencies.


Robdel12 (approved):
🏁

PR #151, opened 2021-01-29
⬆ Upgrade Chromium 78 to 87

wwilsman:
## What is this?

This updates the version of Chromium downloaded and used by asset discovery to Chrome 87.0.4280, which matches the next version that will be used in Percy's own rendering environment.

With this specific version of Chromium, the revisions are varied across all platforms. To make finding these revisions easier than the manual process, I automated it with a script. I also removed the old release-notes script since this repo uses a release-drafting action.

If authenticated with GitHub, the script could be modified to search a range of several versions to find a close match that has only one revision across all platforms. It could also be improved with GitHub's GraphQL API to search for the tag in a single query rather than paginating all tags for the Chromium repo. But that can be a future improvement for the next update.

Two tests needed to be updated:

1. A test in which the error message thrown by Chromium's Dev Tools Protocol has changed.
2. A test which triggers a page crash by navigating to `chrome://crash`. Scripts are no longer allowed to navigate to URLs which crash the page, so a few private properties were un-privated in order to trigger a page crash by issuing a `Page.crash` command over CDP.

While fixing test 2, I updated some discovery logs as well

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
     "build:watch": "lerna run build --stream -- --watch",
     "bump-version": "lerna version --no-git-tag-version --no-push",
     "clean": "rm -rf packages/**/{dist,.nyc_output,coverage,oclif.manifest.json}",
-    "release-notes": "./scripts/release-notes",
     "lint": "eslint --ignore-path .gitignore .",
     "readme": "lerna run --parallel readme",
     "test": "lerna run --stream --concurrency=1 test -- --colors",
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "build": "lerna run build --stream",
     "build:watch": "lerna run build --stream -- --watch",
     "bump-version": "lerna version --no-git-tag-version --no-push",
+    "chromium-revision": "./scripts/chromium-revision",
     "clean": "rm -rf packages/**/{dist,.nyc_output,coverage,oclif.manifest.json}",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -7,8 +7,8 @@
     "build": "lerna run build --stream",
     "build:watch": "lerna run build --stream -- --watch",
     "bump-version": "lerna version --no-git-tag-version --no-push",
+    "chromium-revision": "./scripts/chromium-revision",
     "clean": "rm -rf packages/**/{dist,.nyc_output,coverage,oclif.manifest.json}",
-    "release-notes": "./scripts/release-notes",
```
Robdel12, line 11: noice

packages/core/src/discovery/network.js
```diff
@@ -8,6 +9,8 @@ export default class Network {
   #intercepts = new Map();
   #authentications = new Set();
 
+  log = logger('core:network');
```
Robdel12, line 12: _fuuuuuuuttttttttttuuureeeeeeeeeeeee_

Robdel12 (approved):
🏁

Robdel12:
Oh, what's this mean for the support tool's version of Chrome? Starts using 87?

Robdel12:
Fix your bootleg emoji? 😂

wwilsman:
> Fix your bootleg ⬆️ emoji? 😂

No, it just gets corrected in some places automatically 😆

PR #158, opened 2021-02-01
✨ Add elapsed time to debug messages

wwilsman:
## What is this?

This adds an additional detail to debug logs which can come in handy when debugging: elapsed time. The time between debug logs can be a useful clue in finding the root cause of issues.

image

The logger test helper will automatically remove the elapsed time so other debug tests do not need to be updated. However, there is an option to leave elapsed time alone in case it is something that needs to be tested (such as within logger tests themselves). The log dump test helper was also updated to include the elapsed time between log entries.


Robdel12 (approved):
SHIP THIS, RIGHT NOW! 🏁

PR #159, opened 2021-02-01
🐛 Fix wait-for helper idle option

wwilsman:
## What is this?

The `waitFor` helper inside of `@percy/core` didn't appear to have any issues, however the option `ensure` was actually meant to be `idle` as evidence by uses of an `idle` option elsewhere in the codebase. This caused those options to be ignored when they should not have been. This was likely missed due to requiring coverage ignore the function as it gets injected into the browser where coverage variables are lost.

Fixing this might help with some flakey tests. 🤔


Robdel12 (approved):
🏁 Tricky

PR #160, opened 2021-02-02
✨ Backwards compatibility with v1 config

wwilsman:
## Purpose

Currently, if a v1 config file is detected, it's configuration options are discarded. In order for config files to work with `@percy/cli`, the config file needs to be in a v2 format which can be migrated to using the `config:migrate` command. This PR breaks up the migration logic from the command and moves it alongside respective config schemas, allowing the `@percy/config` package to migrate configuration options on-the-fly.

## Approach

- The `config:migrate` command no longer houses it's own migration logic. This has been split alongside respective schemas. This allows migration and the migrate command to be extensible just like the schema and validate command is.
- The new migration function can be registered with `PercyConfig.addMigration` which mirrors `PercyConfig.addSchema`. The migration function accepts the `input` config as the first parameter, and a `set` function as the second parameter which can be called with the desired path and value to set on the migrated config.
- For each respective package that now has an additional migration function, a unit test suite was added to account for additional required code coverage.
- When `PercyConfig.load` encounters an older config version, a warning is logged before auto-migrating the config: ``Found older config file version, please run `percy config:migrate` to update to the latest version``

### DIFF
--- a/packages/cli-config/src/commands/config/migrate.js
+++ b/packages/cli-config/src/commands/config/migrate.js
@@ -5,10 +5,6 @@ import Command, { flags } from '@oclif/command';
 import PercyConfig from '@percy/config';
 import logger from '@percy/logger';
 
-function assignOrCreate(obj, key, value) {
-  return Object.assign(obj || {}, { [key]: value });
-}
-
 export class Migrate extends Command {
   static description = 'Migrate a Percy config file to the latest version';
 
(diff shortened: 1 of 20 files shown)


Robdel12 (approved):
🏁 Boom, now non-js SDKs are free to swap agent out for CLI with _no_ issues 🔥

PR #161, opened 2021-02-02
✨ Upgrade to Node 12

wwilsman:
## What is this?

Node 10 will be [EOL on April 30th](https://nodejs.org/en/about/releases/). We can prepare early by specifying a minimum Node version and running our tests on the next maintenance LTS version, Node 12.

### Questions

Should we update the other CI actions to use the current Active LTS, Node 14? I would also encourage development using the Active LTS so we can be prepared next year when Node 12 is EOL and we repeat this process again. Shall I create an `.nvmrc` file specifying Node 14? Are there any other formats or universal way I'm forgetting about to convey development should happen using a specific Node version?


Robdel12:
> Shall I create an .nvmrc file specifying Node 14? 

I think so, I dig that. I don't know if there are any other standards for enforcing that

Robdel12:
Can we change anything about our build? (Are there less things polyfilled from 10 to 12?)

Robdel12 (approved):
🏁 Good to get ahead of this now, before we roll this out to everyone. That way they experience one breaking change (and not another in April, when Node 10 is EOL)

wwilsman:
> Can we change anything about our build? (Are there less things polyfilled from 10 to 12?)

Babel config was updated so `@babel/env` will compile for Node 12 rather than Node 10. After looking at some docs, it looks like the main thing is we'll no longer have to polyfill private class properties. Other updates look to be mostly internal to Node. Once we move to >=14 though, we'll have quite a few other things that will no longer need to be polyfilled.

PR #162, opened 2021-02-02
➖ Remove path-type dependency

wwilsman:
## What is this?

This dependency has many small helpers based on two single sync and async helpers. In both places we use this dependency, we only use it to check if a filepath is a directory synchronously. This is done by the depdendency wrapping `fs.statSync().isDirectory()` in a try-catch to catch `ENOENT` errors. In both places we use this helper, we are already wrapping calls in a try-catch, so we can swap out this dependency entirely with the native method. 

However this dependency will still exist in our tree, as I believe `cosmiconfig` depends on it.


Robdel12 (approved):
🏁 the dep killer

PR #163, opened 2021-02-03
✨ Accept all option syntax styles

wwilsman:
## Purpose

Currently, snapshot options via the CLI API are only accepted as camelCase options. In some languages, this goes against the languages own syntax standards such as `snake_case` variables and options. Allowing the snapshot API endpoint to accept any style of options removes the conversion responsibility from the SDK.

## Approach

The `@percy/config` package already has a `normalize` function that removes empty values and converts kebab-case options to camelCase. This has been modified to also convert snake_case options. The `normalize` function is now exported with the package so that the `@percy/core` server can utilize it for converting snapshot API options.


Robdel12 (approved):
🏁 sweet, another nice fix/addition

PR #164, opened 2021-02-04
🐛 Work around incorrect global context in @percy/dom

wwilsman:
## What is this?

The `@percy/dom` package uses a UMD format which assigns exports to a `globalThis` variable when in a browser environment. However, if the implementation of `globalThis` is incorrect (mozilla/geckodriver#1798), the `PercyDOM` object is not correctly carried forward into other script executions where the DOM is meant to be serialized.

This PR aims to work around this issue by checking if `window` and `globalThis` are not equal to each other (which technically should be impossible under normal circumstances). If they are not equal, the library adds itself to `window` ahead of the eventual UMD dance.

While testing this, I also noticed that babel helpers were inlined multiple times. Updating this setting to `bundled` will ensure they are only bundled once in the final output.

### DIFF
--- a/packages/dom/karma.conf.js
+++ b/packages/dom/karma.conf.js
@@ -45,7 +45,7 @@ module.exports = config => {
         require('@rollup/plugin-node-resolve').default(),
         require('@rollup/plugin-commonjs')(),
         require('@rollup/plugin-babel').default({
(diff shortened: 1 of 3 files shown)


Robdel12 (approved):
🏁 I should have caught this the first time around.

PR #165, opened 2021-02-04
✨ Allow setting multiple client or environment info at a time

wwilsman:
## What is this?

This allows SDKs and other consumers of `@percy/client` to set multiple client or environment info at a time by providing an array instead of requiring an already formatted string. Rather than use arrays and manually checking for uniqueness, we can use Sets which are inherently unique.


Robdel12 (approved):
🏁

PR #166, opened 2021-02-04
🐛 Fix config normalization and migration

wwilsman:
## What is this?

First, the regex used for config key normalization didn't match consecutive words. This new regex will also match mixed usage (`"kebab-and_snek" => "kebabAndSnek"`).

Second, since the `config:migrate` command doesn't require `@percy/core` directly, the config schema and migration are never registered. We were already registering the schema, but also needed to register the migration. **Note**: other schemas and migrations are defined by CLI plugins which are automatically registered.


packages/cli-config/src/hooks/init.js
```diff
@@ -1,7 +1,8 @@
 import PercyConfig from '@percy/config';
-import { schema } from '@percy/core/dist/config';
+import { schema, migration } from '@percy/core/dist/config';
 
-// ensures the core schema is loaded
+// ensures the core schema and migration is loaded
 export default function() {
   PercyConfig.addSchema(schema);
+  PercyConfig.addMigration(migration);
 }
```
wwilsman, line 8: We could technically move this to `@percy/cli-exec` where it might be technically more appropriate? 🤔

packages/config/test/index.test.js
```diff
@@ -495,6 +495,31 @@ describe('PercyConfig', () => {
...
+      expect(PercyConfig.normalize({
+        'foo-bar': 'baz',
+        foo: { bar_baz: 'qux' },
+        'foo_bar-baz': 'qux'
+      })).toEqual({
+        fooBar: 'baz',
+        foo: { barBaz: 'qux' },
+        fooBarBaz: 'qux'
+      });
+    });
```
Robdel12, line 520: 👍🏼 Tests

packages/cli-config/src/hooks/init.js
Robdel12: Hmm, I don't feel strongly either way -- your call 👀

Robdel12 (approved):
🏁

PR #167, opened 2021-02-05
✨ Prefer kebab-case for yaml config files

wwilsman:
## Purpose

The yaml config format typically prefers keys to be kebab-cased. Our docs even show the yaml config files with kebab-cased keys. With config migration moved into `@percy/config`, it always normalizes keys as camelCased. We should always prefer kebab-cased when migrating to a yaml format.

## Approach

Allow normalize to accept options other than `overrides`, including an option to convert keys to kebab-cased rather than camelCase. The new kebab-case function will convert both camelCase and snake_case, similarly to how the camelCase function converts both kebab-case and snake_case. In order to account for edge case camelizations such as `CSS` and `JavaScript`, the edge cases map was utilized to covert keys the other direction as well to avoid accidentally adding dashes where there shouldn't be.


Robdel12 (approved):
🏁

PR #173, opened 2021-02-08
💚 Add @percy/logger to CI test workflow

wwilsman:
## What is this?

It appears that `@percy/logger` tests slipped by without being added to the CI tests workflow. This PR adds the package to the workflow so that its tests are run for each commit.


Robdel12 (approved):
🏁 Oops..

PR #174, opened 2021-02-08
♻️ Migrate to package.json config without string replacement

wwilsman:
## What is this?

Although I haven't noticed any issues with the string replacement method, it feels like it has the potential to be buggy. I think the initial intention was to avoid having to parse the entire json file, but in reality parsing a single package.json file is very quick.

While making this change, I noticed that the parsing and re-stringifying causes the file to lose it's line ending. I decided to add this line ending to the config stringify output so it will be consistent for all formats.


Robdel12 (approved):
🏁 Very nice, I was thinking about this bug this morning. JSON parse 💯

PR #175, opened 2021-02-08
✨ Autoload plugins

wwilsman:
## Purpose

With `@oclif/plugin-plugins`, the user is _required_ to use the various `plugin:*` commands in order to use plugins with the CLI. This means that if the CLI and a CLI plugin are both dependencies, the plugin will **not** automatically be associated with the CLI until the user runs the appropriate `plugin:*` command.

## Approach

Before calling the first OCLIF entry point within `@percy/cli`, we can search through `node_modules` for `@percy/*` and `percy-cli-*` packages to append to `@percy/cli`'s list of plugins. 

This is similar but distinctly different from how the `@oclif/plugins-plugin` handles plugins. With the official plugin, it has its own dependency on `yarn` which manages plugin package, and registers plugins via editing a global data directory's `package.json` file associated with the CLI package. This makes it so all versions of the CLI that may be installed on a single machine in various projects have the same shared registered plugins.

The autoload approach used here will instead edit the local `node_modules/@percy/cli/package.json` file so that only plugins at the same scope as the CLI are registered. Global CLI plugins will be registered with the global CLI package, and individual project plugins will be registered with the project's own CLI dependency.

All `@percy/*` packages are checked for plugin compatibility so official SDKs that extend the CLI do not need to name themselves as CLI plugins. However, rather than checking all `percy-*` dependencies for CLI plugins, only `percy-cli-*` dependencies are checked, requiring custom third party SDKs/plugins to be more explicit.

Finally, the `@oclif/plugins-plugin` package was removed in favor of this autoload approach and letting the users manage Percy CLI plugins using their normal, familiar, package manager. In the future, whenever we require plugins to be installed outside of Node ecosystems, we can leverage [the tarball's included Node binary](https://oclif.io/docs/releasing#standalone-tarballs) to manage plugins.

### DIFF
--- a/packages/cli/index.js
+++ b/packages/cli/index.js
@@ -1,2 +1,43 @@
-// This package is a multi-command oclif CLI composed of plugins that provide
-// various functionalities for taking Percy snapshots.
+const { promises: fs } = require('fs');
+const path = require('path');
+
+// find plugins in a directory, matching a regexp, ignoring registered plugins
+function findPlugins(dir, regexp, registered) {
+  return fs.readdir(dir).then(f => f.reduce((plugins, dirname) => {
+    if (dirname === 'cli' || !regexp.test(dirname)) return plugins;
+
+    let { name, oclif } = require(`${dir}/${dirname}/package.json`);
+
+    if (!registered.includes(name) && oclif && oclif.bin === 'percy') {
+      return plugins.concat(name);
+    } else {
+      return plugins;
+    }
+  }, []));
+}
+
+// automatically register/unregister plugins by altering the CLI's package.json within node_modules
+async function autoRegisterPlugins() {
+  let pkgPath = path.resolve(__dirname, 'package.json');
+  let pkg = require(pkgPath);
+
+  let registered = pkg.oclif.plugins.filter(plugin => {
+    if (pkg.dependencies[plugin]) return true;
+    try { return !!require.resolve(plugin); } catch {}
+    return false;
+  });
+
+  let unregistered = await Promise.all([
+    findPlugins(path.resolve(__dirname, '..'), /^.*/, registered), // @percy/*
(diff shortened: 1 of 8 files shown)


packages/cli/index.js
```diff
@@ -1,2 +1,43 @@
...
+
+  if (unregistered.length || registered.length !== pkg.oclif.plugins.length) {
+    pkg.oclif.plugins = registered.concat(unregistered);
+    await fs.writeFile(pkgPath, JSON.stringify(pkg, null, 2) + '\n');
+  }
+}
+
+// auto register plugins before running oclif
+module.exports.run = () => autoRegisterPlugins()
+  .then(() => require('@oclif/command').run());
```
Robdel12, line 43: Very slick 💯

packages/cli/package.json
```diff
@@ -16,11 +16,15 @@
   },
   "scripts": {
     "lint": "eslint --ignore-path ../../.gitignore .",
-    "postbuild": "oclif-dev manifest"
```
Robdel12, line 19: So the manifest is only for linking with the plugins plugin?

yarn.lock
```diff
@@ -9553,11 +9487,6 @@ yargs@^16.1.1:
     y18n "^5.0.5"
     yargs-parser "^20.2.2"
 
-yarn@^1.21.1:
```
Robdel12, line 9556: Amazing. Just from this, we're dropping Yarn from the dep chain Percy includes.

Robdel12 (approved):
🏁

packages/cli/package.json
wwilsman: The manifest is for telling OCLIF which commands the CLI / plugin contain. This CLI package not only doesn't export any commands itself, but it has no build step so the postbuild step never ran to generate the empty file to begin with.

PR #176, opened 2021-02-08
💚 Add Windows workflow

wwilsman:
## What is this?

This PR adds a Windows CI workflow and fixes a few issues that cause test failures in Windows environments. For the most part, this involves using path helpers rather than hardcoding path separators.

Windows also has issues with atomic file operations. Errors from `rimraf` are now ignored in `@percy/core` tests, and the browser installation test was removed due to it also needing to delete files. Since this test was removed, browser installation coverage became zero which prompted a coverage ignore statement to be added. Ultimately, the tests need the browser to be downloaded in order to work, so I'm confident in removing coverage from this file since it is technically tested before any relevant tests run.

One final _hanging_ issue was that `@percy/cli-exec` tests were hanging, but only in Windows CI. In a local Windows environment, I was unable to reproduce. Debugging showed it to be because of the browser processes's stdio pipes being left open. Even stranger, is `@percy/core` tests were **not** hanging even though the source seemed to be stemming from there. In the end, I found [this internal Node issue](https://github.com/nodejs/node/issues/27097) in which a comment mentions a workaround for ending the piped streams manually after the process exits. Adding that change indeed fixed the hanging issue, and another comment points to the issue being fixed as of Node 13. Since Node 12 still has a year of life left, we should leave this workaround in for the time being (even though I didn't experience it myself locally).

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,0 +1,67 @@
+name: Windows
+on: push
+jobs:
+  build:
+    name: Build
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - name: Get yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - uses: actions/cache@v2
+        with:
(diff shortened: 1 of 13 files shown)


Robdel12 (approved):
🏁 the fight is over, very nice hahaha

PR #177, opened 2021-02-08
✨ Accept browser executable env var

wwilsman:
## What is this?

The browser executable option is typically more useful as a CI environment variable rather than a config option that is carried forward to various environments.


Robdel12 (approved):
🏁

PR #183, opened 2021-02-09
🐛 Improve plugin autoloading

wwilsman:
## What is this?

When the CLI is linked or eventually distributed outside of NPM, it might not always reside within a project or global `node_modules`. In these cases, it should autoload plugins from within its own `node_modules`. This will enable linking plugins to the linked CLI and also allow future iterations to install plugins into the CLIs own `node_modules`.

This is labeled as a bug although it doesn't necessarily fix anything because it would be one if we distributed the CLI outside of NPM.

### DIFF
--- a/packages/cli/index.js
+++ b/packages/cli/index.js
@@ -1,37 +1,57 @@
 const { promises: fs } = require('fs');
 const path = require('path');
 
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
🏁

PR #184, opened 2021-02-09
✅ Fix flakey tests

wwilsman:
## What is this?

Some tests are flakey. This PR aims to unflake them all

- [X] `@percy/dom` frame serialization tests
    
    Sometimes a frame may take longer than expected to load. Even the JS only frame may load after modifying the frame document causing that test to flake. On windows, frames can take even longer to load than 3s, so the wait timeout was upped to 5s and the hook timeout was removed.

- [X] `@percy/core` capture handles closing during network idle

    Expected: `Network error: Page closed`
    Received: `Protocol error (Runtime.callFunctionOn): Page closed`

    Received message suggests that the page is sometimes closing either during the provided script execution or after the second network idle during DOM serialization script execution.

    It's possible for the test to continue after an asset has been accessed but before the request has been intercepted causing it to flake. Delaying when asset access is detected should help the request be intercepted in time for the assertion.

- [X] `@percy/cli-exec` start abort test

    The test sends the process quit signal after running the `exec:start` command, then issues a request to see if the server is down. The server is sometimes still up when that request is sent, but it should be on its way down.

    Sometimes it can take more than half of a second for the server to shutdown when forcefully terminating. Removing the arbitrary wait and making the test more deterministic should show if this was a test timing issue, or if termination itself is flakey.
- [X] `@percy/core` flakey coverage

    Coverage is sometimes missing from [here](https://github.com/percy/cli/blob/master/packages/core/src/discovery/browser.js#L104). Not sure which test normally covers this, but a test was added that should always cover this as long as it passes.

### DIFF
--- a/packages/cli-exec/test/start.test.js
+++ b/packages/cli-exec/test/start.test.js
@@ -24,12 +24,14 @@ describe('percy exec:start', () => {
     ).resolves.toBeDefined();
 
     process.emit('SIGTERM');
-    // wait for event to be handled
-    await new Promise(r => setTimeout(r, 500));
 
-    await expect(
-      fetch('http://localhost:5338/percy/healthcheck', { timeout: 10 })
-    ).rejects.toThrow();
+    // check a few times rather than wait on a timeout to be deterministic
+    await expect(function check(i = 0) {
+      return fetch('http://localhost:5338/percy/healthcheck', { timeout: 10 })
+        .then(r => i >= 10 ? r : new Promise((res, rej) => {
+          setTimeout(() => check(i++).then(res, rej), 100);
+        }));
+    }()).rejects.toThrow();
   });
 
   it('logs an error when percy is already running', async () => {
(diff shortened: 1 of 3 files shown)


packages/cli-exec/test/start.test.js
```diff
@@ -24,12 +24,14 @@ describe('percy exec:start', () => {
     ).resolves.toBeDefined();
 
     process.emit('SIGTERM');
-    // wait for event to be handled
-    await new Promise(r => setTimeout(r, 500));
 
-    await expect(
-      fetch('http://localhost:5338/percy/healthcheck', { timeout: 10 })
-    ).rejects.toThrow();
+    // check a few times rather than wait on a timeout to be deterministic
```
Robdel12, line 28: connnveerrrggeeeeee

wwilsman:
Re-ran latest CI 4-5 times and it didn't fail. This looks good to go.


_inb4 it fails after merging_

Robdel12 (approved):
🏁 🐛 🔨

PR #185, opened 2021-02-10
🐛 Catch autoload directory errors

wwilsman:
## What is this?

If the directories being searched in do not exist, those errors should be silently ignored since no plugins can exist in a non-existent directory.


Robdel12 (approved):
🏁

PR #186, opened 2021-02-12
♻️ Update `agent.js` deprecation message to be user friendly

Robdel12:
## What is this?

When accessing the old DOM library server endpoint (that `@percy/agent` uses) we log a deprecation message. The previous deprecation message is too technical and won't make sense unless you know how the SDKs piece together. The new one aims to be more actionable.

### DIFF
--- a/packages/core/src/server.js
+++ b/packages/core/src/server.js
@@ -113,7 +113,7 @@ export default function createPercyServer(percy) {
       .readFile(require.resolve('@percy/dom'), 'utf-8')
       .then(content => {
         let wrapper = '(window.PercyAgent = class PercyAgent { snapshot(n, o) { return PercyDOM.serialize(o); } });';
-        log.deprecated('`percy-agent.js` is deprecated, please update to the latest SDK version');
+        log.deprecated('It looks like you’re using @percy/cli with an older SDK. Please upgrade to the latest version to fix this warning. See these docs for more info: https://docs.percy.io/docs/migrating-to-percy-cli');
         return [200, 'applicaton/javascript', content.concat(wrapper)];
       }),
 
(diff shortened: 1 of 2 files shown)


packages/core/src/server.js
```diff
@@ -113,7 +113,7 @@ export default function createPercyServer(percy) {
       .readFile(require.resolve('@percy/dom'), 'utf-8')
       .then(content => {
         let wrapper = '(window.PercyAgent = class PercyAgent { snapshot(n, o) { return PercyDOM.serialize(o); } });';
-        log.deprecated('`percy-agent.js` is deprecated, please update to the latest SDK version');
+        log.deprecated('It looks like you’re using @percy/cli with an older SDK. Please upgrade to the latest version to fix this warning.');
```
Robdel12, line 116: How do we feel about hard coding a docs URL in here? 😬 https://docs.percy.io/docs/migrating-to-percy-cli

wwilsman: I really wish the doc could be managed by the repo (`/docs/*.md`) so we can update the link if the doc moves or changes. Probably fine for now for this log, since this endpoint will likely be removed in a future major version anyway

packages/core/src/server.js
```diff
@@ -113,7 +113,7 @@ export default function createPercyServer(percy) {
       .readFile(require.resolve('@percy/dom'), 'utf-8')
       .then(content => {
         let wrapper = '(window.PercyAgent = class PercyAgent { snapshot(n, o) { return PercyDOM.serialize(o); } });';
-        log.deprecated('`percy-agent.js` is deprecated, please update to the latest SDK version');
+        log.deprecated('It looks like you’re using @percy/cli with an older SDK. Please upgrade to the latest version to fix this warning. See these docs for more info: https://docs.percy.io/docs/migrating-to-percy-cli');
```
wwilsman, line 116: This reminds me, we should add a max-len eslint rule for readability 😉

wwilsman (changes requested):

wwilsman (approved):
🎮

PR #201, opened 2021-02-22
✨ Add config search method

wwilsman:
## What is this?

This pattern is used in `@percy/config`, `@percy/cli-config`, and `@percy/migrate`; so making it a helper will DRY up those places.

I also removed a redundant error message as the real error gets printed anyway.


Robdel12 (approved):
🏁

PR #203, opened 2021-02-22
🐛 Fix normalized config request-headers

wwilsman:
## Purpose

Currently the normalize method traverses nested objects, including `request-headers`, and converts the casing of the keys to be predictable. Request header keys however, should not be converted and key casing should be left as-is.

## Approach

Add a `skipMerge` array to the merge helper that will not traverse the object any further if the key is present in the array. Currently, the only key to skip is `request-headers`.


Robdel12 (approved):
🏁

PR #204, opened 2021-02-23
♻ Improve browser process handling

wwilsman:
## What is this?

While chasing an issue on Windows where the `@percy/cli-exec` test suite hangs, some changes were made in an attempt to ensure the browser process was being handled correctly. While this ultimately didn't fix the issue on Windows, I do think these are good changes that should be adopted regardless.

- Add check if browser is already closed
- Use cross-spawn to spawn browsers
- Do not spawn detached browser in Windows
- Always kill browser process when able to


Robdel12 (approved):
🏁

PR #205, opened 2021-02-27
✨ Expose logger's shouldLog method

wwilsman:
## What is this?

It might become more common to verify the log level before performing some logic. Since log levels are well, levels, it's slightly more complicated than an equal operator. Rather than repeating log level logic, we can expose the `shouldLog` instance method to any logger groups. This will allow them to determine if their specific logging group is allowed to log a specific log level. 

For example, `logger('group:id').shouldLog('info')` will only return truthy if both the log level is either `info` or `debug`, and the group id is present in any `PERCY_DEBUG` environment variable. This is much better and more succinct than checking the log level alone.

### DIFF
--- a/packages/logger/src/logger.js
+++ b/packages/logger/src/logger.js
@@ -67,7 +67,8 @@ export default class PercyLogger {
       .reduce((group, level) => Object.assign(group, {
         [level]: this.log.bind(this, name, level)
       }), {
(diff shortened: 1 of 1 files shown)


Robdel12 (approved):
🏁 This will be super helpful for sure

PR #206, opened 2021-02-27
✨ Add ability to download browser during post-install

wwilsman:
## What is this?

This adds a post-install environment variable check for `PERCY_POSTINSTALL_BROWSER` that will trigger the browser download during the post-install step rather than lazily during the start of asset discovery.

The core browser install script was also refactored and moved out of core utils. This script is an integral part of core's asset discovery process and more than a simple utility function. The base install function was made to be more agnostic of what file is being downloaded and extracted. Chromium install defaults were moved to their own function rather than being defaults of the agnostic install function. Unit-like tests were also added for the core install function.

Because of this post-install check, we can remove the `pretest` script entirely since the browser can be download on install by specifying the environment variable during the CI install step. However, since the browser does not install to `node_modules` in development, it would have to be cached in addition to those packages. Because of this, the CI cache setup was changed to cache multiple directories directly rather than caching yarn's own cache.

To allow the post-install step to run in development without access to pre-built files, I adjusted the babel config and added a module resolver to resolve `@percy` packages to their in-repo counterparts. In addition to allowing the post-install script to run in development, this also allows many tests to run without having to also require build files (a compiled `@percy/dom` bundle is still required for relevant tests).


Robdel12 (approved):
🏁

Robdel12 (approved):
🏁 Nice add for the prod env var

PR #218, opened 2021-03-01
🐛 Include post-install script in packaged files

wwilsman:
## What is this?

The addition to include the post-install script in the packaged files was missed in #206


Robdel12 (approved):
🏁 I take blame for that lol dangit

PR #219, opened 2021-03-01
🐛 Check src directory for post-install

wwilsman:
## What is this?

The post-install script checks for a test directory to determine if it is being run in development. This package exports a test helper which means the test directory is also present in production. The src files should never be included, so should be a much safer check.


Robdel12 (approved):
🏁

PR #222, opened 2021-03-03
✨ Add exec partial build flag

wwilsman:
## What is this?

Rather than requiring the user set an environment variable to determine if a build is a partial build, we can set the environment variable automatically when provided a `--partial` flag.

This brought up a question in my mind about the `--parallel` flag's short version, `-p`. Could be confusing now with a second "p" argument; should we remove the parallel shorthand to avoid confusion before leaving beta?


Robdel12 (approved):
🏁 Oh slick, nice! Forgot about this

PR #231, opened 2021-03-10
🔨 Improve repo scripts

wwilsman:
## What is this?

This drastically improves the scripts used to test and build packages in the repo.

### Testing

While browserifying `@percy/logger` along my way to `@percy/sdk-utils`, I needed to copy `@percy/dom` test and build setup and make sure it worked alongside the node scripts already used to compile those packages with babel. Those two configuration files together were already pretty extensive, so just copying and pasting them into the other packages didn't feel like the correct thing to do within a monorepo.

The Node tests already shared a new test script, so it seemed appropriate to expand on that script to run browser tests as well. By default, it will now run browser tests based on the presence of a `browser` field in the appropriate `package.json` file. Node tests will be skipped if both the `browser` and `main` fields are equal to the same file, indicating a browser-only package. When running both types of test, I discovered that the test runners like to assume that they are in control of the process; so if both types are to run, they are run one after another in forked node processes.

I utilized `yargs-parser` (a sub-dependency from `lerna`, `nyc`, and `tsd`) to parse script arguments to allow specifying when to _only_ run certain types of tests (`--node`, `--browsers`). Karma options can also be provided by nesting `--karma.<options>` which can be useful when debugging (e.g. `test --browsers --karma.browsers Chrome --no-karma.single-run`)

### Coverage

The test script presented another opportunity to make coverage collection and reporting better as well. When either type of tests run, they report respective coverage results; Karma, with `karma-coverage`, will report coverage for each respective browser test report as well. Since we currently test in two browsers, testing a polymorphic package would result in three test reports and three coverage reports. The coverage reports would also not be combined, causing "missed" coverage for code that will never run in the browser and vice versa.

First, the `karma-spec-reporter` and `karma-coverage` plugins were removed since they result in multiple reports for each browser. The `karma-mocha-reporter` plugin is now used and is based on the default Mocha reporter and prints a single report for all browsers. Coverage is handled by leveraging Karma's server events and populating a `global.__coverage__` variable which is picked up when running the `nyc` command. While this enables normal usage of the `nyc` command with our test script, I still opted to have the test script itself handle running `nyc` so we can take advantage of other script features.

Coverage can be tested by using the `--coverage` flag when running any type of tests (`yarn test:coverage` was left as an alias for each package). A coverage reporter can then be specified with the `--reporter` flag, which will report the combined coverage of all tests. This is useful for debugging which lines need covering (e.g `test --coverage --reporter html-spa`)

### Building

The inclusion of automatically handing which type of test to perform, as well as the shared rollup config required by the now shared karma config, inspired me to also consolidate the build scripts. This script is much simpler than the test script, but uses similar techniques from the test script for future expansion. Like the test script, the type of build is automatically detected but can be manually defined via a flag (`--node`, `--bundle`).

### Watching

All new scripts can have a `--watch` argument which will run the script and follow it up with [Gaze](https://github.com/shama/gaze) to re-run the main function whenever any package files are modified. In order to prevent watching `node_modules` files, `dist` files which can cause a infinite build loop, or any other undesirable files, the `.gitignore` file is parsed and transformed into a list of negated globs provided to Gaze. This warranted pulling the main watch logic out into its own sharable file. In the future, this could also be modified to accept command arguments and automatically respawn the command on file changes.

### Other Changes

I also removed an outdated Webpack config file and fixed `@percy/dom` tests, which became broke due to the extensive refactoring of the `rollup.config` file made to work for all repo packages.

### Questions

Should we just add `yargs-parser` to dev deps to be explicit? If so, I'm thinking I'll also add `yargs-unparser` to allow Karma options to be passed down through with coverage logic (`--coverage` will currently ignore `--karma.<options>`)


Robdel12 (approved):
🏁 Nice work! Really polished, you would write a little CLI for this 😂

PR #232, opened 2021-03-10
🐛  Catch any errors that arise from `globalThis` in the DOM libary

Robdel12:
## What is this?

Nightmare/electron envs don't have `globalThis` defined, which errors and breaks snapshots. This wraps a try/catch around it to stop that from happening


wwilsman (approved):
🎮

PR #233, opened 2021-03-10
🔨 Fix script watching

wwilsman:
## What is this?

If the first run of a script fails, it never gets to the watch logic. The error handler also called `process.exit`, which it should not do when watching.

Since test runners like to assume they have control over the process, when run together they were forked. They are now always forked since the watch process needs to control its own process.


Robdel12 (approved):
🏁

PR #234, opened 2021-03-10
🐛 Catch undefined globalThis in bundles

wwilsman:
## What is this?

In response to #232, I forgot that `globalThis` is also referenced in build bundles.


Robdel12 (approved):
🏁 I am lol'ing appoving

PR #235, opened 2021-03-10
✨ Add browser and remote support for @percy/logger

wwilsman:
## What is this?

This PR contains two important change to `@percy/logger`.

### Browserification

This takes advantage of all the new test and build scripts to create a browser bundle by simply adding two new fields to the package.json. One to signify a browser bundle, and another to provide package specific rollup options. Of course, the package didn't magically work in a browser. Although it was close! The logger's `log` method would write directly to the process stdio, which doesn't exist in a browser. The test helper was also relying on a native stream object to mock stdio properties for testing.

The write logic was moved into it's own method, as well as some other methods being moved around. A new `BrowserLogger` class was created to extend the normal logger while replacing the new write method with a browser compatible one. Since logs are still colorized via ansi escape codes, and some browsers such as Firefox do not support logging ansi escape sequences, they are instead replaced with the string `%c` and the appropriate CSS is added to another array which is appended to the console log call. This CSS log formatting is supported in all major desktop browsers with the exceptions of IE and versions of Edge prior to 79 (released Jan 2020).

With the help of the bundled `__PERCY_BROWSERIFIED__` environment variable, the logger index file will conditionally use the correct Logger class when creating logger instances. The test helper also uses this variable to spy on the logger's write method and console log usage rather than creating a mock stdio stream.

The singleton logic was also moved to be within the class itself rather than as part of the index helper. As such, the index file was refactored a little. The colors util was also adjusted to allow the browser class to more easily replace ansi codes with CSS colors. The test helper was also change to remove trailing newlines from recorded log messages since the browser logger and new write method both do not receive the newline character and it is only written purposefully in node environments.

### Remote Logging

Part two of this PR doesn't necessarily relate to browser support, but does allow browser loggers to log out to the CLI process when a local Percy server is running. Enabling this in an agnostic way regardless of WebSocket implementation allows this feature to also work across other Node processes. In turn, allowing our main logger to capture and record logs from within the SDKs themselves.

This is made available via two new logger methods. The `#connect` method, to be used by the main logger, will send logging environment info, attach a `message` listener, and allow the connected WebSocket to issue remote logs. The `#remote` method, to be used by SDK loggers, will wait for an open WebSocket connection, silently send previously logged entries, and attach a listener to update local logger environment info. When logging and connected to a remote instance, logs will flow through the websocket and be handled by the remote logger. If a connection cannot be established, a debug error is logged and local logs will continue to be handled locally.

Since the new methods were made to be agnostic of the WebSocket implementation, it will have to be handled in two other places. First, in this PR, the core server was given a WebSocket server via `ws` (already a dependency for CDP connections). This simple server will connect any socket to the logger. In the future, it could be made more robust and allow snapshots and other info to be sent over the established connection. Part two of the WebSocket connect must be made by the SDK. This can be handled for all official JS SDKs via the `@percy/sdk-utils` package (once it is browserified). There is also the potential for other languages to be able to connect to and use the remote logger via their own WebSocket implementations.

### Related changes

Since logger test helpers were updated, that means tests throughout the repo that used the helper needed to be updated. This also means that JS SDKs that use the logger test helper will also need to be updated.

### DIFF
--- a/packages/core/src/server.js
+++ b/packages/core/src/server.js
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import http from 'http';
+import { Server as WSS } from 'ws';
 import PercyConfig from '@percy/config';
 import logger from '@percy/logger';
 import pkg from '../package.json';
@@ -90,7 +91,7 @@ export function createServer(routes) {
 export default function createPercyServer(percy) {
   let log = logger('core:server');
 
-  return createServer({
+  let context = createServer({
     // healthcheck returns meta info on success
     '/percy/healthcheck': () => [200, 'application/json', {
       success: true,
@@ -113,8 +114,8 @@ export default function createPercyServer(percy) {
       .readFile(require.resolve('@percy/dom'), 'utf-8')
       .then(content => {
         let wrapper = '(window.PercyAgent = class PercyAgent { snapshot(n, o) { return PercyDOM.serialize(o); } });';
-        log.deprecated('It looks like you’re using @percy/cli with an older SDK. Please upgrade to the latest version' +
-          ' to fix this warning. See these docs for more info: https://docs.percy.io/docs/migrating-to-percy-cli');
+        log.deprecated('It looks like you’re using @percy/cli with an older SDK. Please upgrade to the latest version' + (
+          ' to fix this warning. See these docs for more info: https://docs.percy.io/docs/migrating-to-percy-cli'));
         return [200, 'applicaton/javascript', content.concat(wrapper)];
       }),
 
@@ -141,4 +142,18 @@ export default function createPercyServer(percy) {
       success: false
     }]
   });
+
+  // start a websocket server
+  context.wss = new WSS({ noServer: true });
+
+  // manually handle upgrades to avoid wss handling all events
+  context.server.on('upgrade', (req, sock, head) => {
+    context.wss.handleUpgrade(req, sock, head, socket => {
+      // allow remote logging connections
+      let disconnect = logger.connect(socket);
+      socket.once('close', () => disconnect());
+    });
+  });
+
+  return context;
 }
--- a/packages/core/test/server.test.js
+++ b/packages/core/test/server.test.js
@@ -162,6 +162,40 @@ describe('Snapshot Server', () => {
     expect(percy.snapshot).not.toHaveBeenCalled();
   });
 
+  it('facilitates logger websocket connections', async () => {
+    await percy.start();
+
+    logger.reset();
+    logger.loglevel('debug');
+
+    // log from a separate async process
+    let [stdout, stderr] = await new Promise((resolve, reject) => {
+      require('child_process').exec('node -e \'' + [
+        'const logger = require("@percy/logger");',
+        // connect to the local websocket server for logging and unref the socket
+        'const ws = new (require("ws"))("http://localhost:1337");',
+        'ws.on("open", () => ws._socket.unref());',
+        // set loglevel to debug failures
+        'logger.loglevel("debug");',
+        // connect and send a remote log
+        'logger.remote(ws).then(() => ',
+        'logger("remote-sdk").info("whoa"));'
+      ].join('') + '\'', (err, stdout, stderr) => {
+        if (!err) resolve([stdout, stderr]);
+        else reject(err);
+      });
+    });
+
+    // child logs are present on connection failure
+    expect(stdout.toString()).toEqual('');
+    expect(stderr.toString()).toEqual('');
+
+    expect(logger.stderr).toEqual([]);
+    expect(logger.stdout).toEqual([
+      '[percy:remote-sdk] whoa'
+    ]);
+  });
+
   describe('when the server is disabled', () => {
     beforeEach(async () => {
       percy = await Percy.start({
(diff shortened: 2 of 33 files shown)


packages/logger/src/logger.js
```diff
@@ -144,8 +119,131 @@ export default class PercyLogger {
...
+  connect(socket) {
+    // send logging environment info
+    let PERCY_DEBUG = process.env.PERCY_DEBUG;
+    let PERCY_LOGLEVEL = process.env.PERCY_LOGLEVEL || this.loglevel();
+    socket.send(JSON.stringify({ env: { PERCY_DEBUG, PERCY_LOGLEVEL } }));
+
+    // attach remote logging handler and return a cleanup function
+    return listen(socket, 'message', ({ data }) => {
+      let { log, logAll } = JSON.parse(data);
+      if (logAll) logAll.forEach(e => this.messages.add(e));
```
wwilsman, line 197: This will silently add previously logged remote logs to this loggers memory store. They have the potential to be added "out of order" since old logs are being added to the current Set order. However, each entry has a timestamp if there is ever confusion. If it's really important (probably not), we could potentially write a small merge util that will order them by timestamp.

packages/logger/src/logger.js
```diff
@@ -144,8 +119,131 @@ export default class PercyLogger {
...
+      if (log) this.log(...log);
+    });
+  }
+
+  // Connects to a remote logger
+  async remote(socket, timeout = 1000) {
```
wwilsman, line 203: In practice, this connection shouldn't be established if the health check fails. So this timeout is more precautionary, although has the potential to add 1 second to test times during the connection handshake. We could make this much lower if we feel the need to. How long is too long for this novel feature?  Since it will fallback to local logs, this failure will really only effect log files and browser loggers.

I think we can utilize this new feature in the SDK utils and do some manual testing to see how quick the connection is typically made within to see what we think we can safely lower this to.

packages/core/test/server.test.js
```diff
@@ -162,6 +162,40 @@ describe('Snapshot Server', () => {
...
+    logger.reset();
+    logger.loglevel('debug');
+
+    // log from a separate async process
+    let [stdout, stderr] = await new Promise((resolve, reject) => {
+      require('child_process').exec('node -e "' + [
+        "const logger = require('@percy/logger');",
+        // connect to the local websocket server for logging and unref the socket
+        "const ws = new (require('ws'))('http://localhost:1337');",
+        "ws.on('open', () => ws._socket.unref());",
```
wwilsman, line 177: This is essentially what the SDK utils will have to do agnostically

packages/core/src/server.js
```diff
@@ -113,8 +114,8 @@ export default function createPercyServer(percy) {
       .readFile(require.resolve('@percy/dom'), 'utf-8')
       .then(content => {
         let wrapper = '(window.PercyAgent = class PercyAgent { snapshot(n, o) { return PercyDOM.serialize(o); } });';
-        log.deprecated('It looks like you’re using @percy/cli with an older SDK. Please upgrade to the latest version' +
-          ' to fix this warning. See these docs for more info: https://docs.percy.io/docs/migrating-to-percy-cli');
+        log.deprecated('It looks like you’re using @percy/cli with an older SDK. Please upgrade to the latest version' + (
+          ' to fix this warning. See these docs for more info: https://docs.percy.io/docs/migrating-to-percy-cli'));
```
Robdel12, line 118: 😂 you're welcome

packages/core/test/server.test.js
```diff
@@ -162,6 +162,40 @@ describe('Snapshot Server', () => {
...
+      });
+    });
+
+    // child logs are present on connection failure
+    expect(stdout.toString()).toEqual('');
+    expect(stderr.toString()).toEqual('');
+
+    expect(logger.stderr).toEqual([]);
+    expect(logger.stdout).toEqual([
+      '[percy:remote-sdk] whoa'
```
Robdel12, line 195: [remember the whoa bowl?](https://scontent-dfw5-1.cdninstagram.com/v/t51.2885-15/e35/s1080x1080/97307038_3197320806966387_5538775687502291611_n.jpg?tp=1&_nc_ht=scontent-dfw5-1.cdninstagram.com&_nc_cat=103&_nc_ohc=Af4LvBwwDwIAX-rXBc0&oh=a4869febf9a4ee4adb29c18e37b7c818&oe=6075ECE2) https://www.whoabbq.com/menu 🤤

packages/logger/src/logger.js
Robdel12: I'd probably half it to start. I bet it lands at 380-420 as a nice default 🤔

packages/logger/src/logger.js
wwilsman: Also just had the idea of having them be sorted by the `query` method so they will be ordered in log files.

Robdel12 (approved):
🏁  This is pretty dang slick. Excited to ship this and then some browser based SDKs 💯

packages/core/src/server.js
wwilsman: Emacs didn't like this line so I added the parens to calm it. 😆

PR #236, opened 2021-03-11
✅ Fix test log debugging

wwilsman:
## What is this?

While debugging some failing tests, I noticed that the failed test check used to produce log dumps did not work due to the recent refactor to Jasmine. Jasmine does not have a way to access the current test from within a hook, but it did feel more proper to add it as a shared reporter in the common test helpers file. Since this file is also loaded in browser environments, I had to add catches around reading the environment variable and also set said variable in an alternate way for karma tests since the common test helpers do not get bundled with rollup.

### DIFF
--- a/karma.config.js
+++ b/karma.config.js
@@ -29,8 +29,12 @@ module.exports = config => config.set({
     'test/**/*.test.js': ['rollupTestFiles']
   },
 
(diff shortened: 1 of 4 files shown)


Robdel12 (approved):
🏁 Ah, fun.

PR #237, opened 2021-03-12
🐛 Force @percy/dom globalization

wwilsman:
## What is this?

Sometimes modules or exports are polyfilled causing the UMD bundle to not attach the DOM script to the global scope. Since this scope issue seems to keep coming up, let's always set the serialize function globally when the window object is defined.

I also updated the bundle intro to use `typeof` rather than `try-catch`.

### DIFF
--- a/packages/dom/src/index.js
+++ b/packages/dom/src/index.js
@@ -1,14 +1,9 @@
 import serialize from './serialize-dom';
 
 /* istanbul ignore next */
-// works around instances where the context has an incorrect global scope
-// https://github.com/mozilla/geckodriver/issues/1798
-try {
-  if (globalThis !== window) {
-    window.PercyDOM = { serialize };
-  }
-} catch (error) {
-  // `globalThis` is probably not defined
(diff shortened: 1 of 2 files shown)


Robdel12 (dismissed):
🏁

packages/dom/src/index.js
```diff
@@ -1,14 +1,9 @@
...
-// https://github.com/mozilla/geckodriver/issues/1798
-try {
-  if (globalThis !== window) {
-    window.PercyDOM = { serialize };
-  }
-} catch (error) {
-  // `globalThis` is probably not defined
+// works around instances where the context has an incorrect scope
+if (typeof window !== 'undefined') {
+  window.PercyDOM = exports;
```
wwilsman, line 6: The test error was because the test bundles use the global `PercyDOM` to reference this module, however setting this manually to a new object caused it to lose other module properties. Setting it to `exports` will set it to be the same value as the module exports (including a `default` property).

Robdel12 (approved):
🏁

PR #249, opened 2021-03-18
✅ Use consistent test helpers name

wwilsman:
## What is this?

Some test helper files were named `helper.js` because they had a single default export. For the sake of consistency, we can name all helpers `helpers.js` which also matches the scenario where helpers need to be broken into `helpers/*.js` files.


.github/.cache-key
```diff
@@ -1 +1 @@
-1
+2
```
Robdel12, line 1: lol

Robdel12 (approved):
🏁

PR #250, opened 2021-03-18
💚 Use CI cache key file for easier cache cleaning

wwilsman:
## What is this?

To clear the cache, the version within the key must be incremented (resulting in a new key). Since this cache key is used in many workflows, it'd be much better for this cache version to live in a single place. This is achieved via `hashFiles` and a new `.github/.cache-key` file.

The yaml "fold" scalar is used (`>`) to concat lines together into a single string. This makes it easier to digest what the cache key is comprised of.


.github/.cache-key
```diff
@@ -0,0 +1 @@
+1
```
Robdel12, line 1: This is basically a "last CI incident" counter 😂

Robdel12 (approved):
🏁

.github/.cache-key
wwilsman: Absolutely 🤣

PR #251, opened 2021-03-18
🔨 Script improvements

wwilsman:
## What is this?

A handful of local script changes:

- Removes an unneeded test babelrc file now that packages are auto-aliased to src files throughout the repo.
- Allows packages to bundle one another's source. This is needed for the future `@percy/sdk-utils` bundle which will also come bundled with `@percy/logger` (without nesting bundles).
- Ignore distributed filepaths in babel's config to hopefully reduce chances of shipping code that works in tests but doesn't work in production (such as certain stage-4 features).
- Forward karma flags to forked test processes; requiring the `flagify` util to reconstruct more complex flags. Any more complex and we'll need a devdep on `yargs-unparser`.
- Use `git clean` to clean the repo rather than `rm`. It works everywhere and uses the .gitignore file when purging ignored files `-X` (see `git help clean`).


Robdel12 (approved):
🏁

PR #253, opened 2021-03-19
✅ Fix test log dumping

wwilsman:
## What is this?

Log dumping was broken in a couple of ways. The actual `dump` method had a member typo which always returned undefined causing the function to abort. Even if it were defined it would log the warning even with zero logs. Finally, the failed test reporter did not work in browser environments since this file does not get bundled. The reporter now references the possible window globals when in a browser context.

Depends on #252


Robdel12 (approved):
🏁

PR #254, opened 2021-03-19
🐛 Fix remote logger cleanup

wwilsman:
## What is this?

The cleanup function for remote logging attempted to remove event listeners, however the events were registered with anonymous functions. By providing the intended function, the argument must be changed along with its usage as the timeout callback.

This bug doesn't necessarily effect anything as the socket lives as long as the process anyway unless it fails to connect. However this code should at least be corrected.

I also experienced an instance where a WebSocket error within the browser may not contain an `error` property. So a fallback message is used for error events without said property (and an accompanying test was added).

Depends on #252 & #253


Robdel12 (approved):
🏁

PR #255, opened 2021-03-19
🐛 Expose version header for CORS

wwilsman:
## What is this?

With most responses, all headers in the request response are included in the response object. In browsers however, the `Access-Control-Expose-Headers` header is strictly respected and if that header does not exist, or the desired header is not defined within it, the response object will not contain all response headers.

Adding this header will allow SDKs within browsers to access the version header exposed by the local Percy server API.

Depends on #252, #253, & #254


Robdel12 (approved):
🏁

PR #256, opened 2021-03-19
♻️ Rename server address method / environment variable

wwilsman:
## Purpose

Initially, the `PERCY_CLI_API` environment variable was named so because it was the CLI that set that variable for SDKs to consume. However, that API address can still be consumed by SDKs without the CLI. In reality, this value comes from the core server, so it's name should reflect that.

`PERCY_SERVER_ADDRESS` was chosen as it is pretty descriptive in that the variable contains the address of the local Percy server. However, I'm open to other suggestions.  bikeshed

Since this variable is only set by the CLI, and official SDKs consume this internally, I wouldn't count this as breaking for our own SDKs. If the user was using the `--port` option (not likely), it will still continue to work once the SDK is updated to account for the new variable name. However this would be considered a breaking change for custom SDKs that do not use sdk-utils, although this package is technically still beta so changes should be expected.

Depends on #252, #253, #254, & #255

### DIFF
--- a/packages/cli-exec/src/commands/exec/index.js
+++ b/packages/cli-exec/src/commands/exec/index.js
@@ -73,7 +73,7 @@ export class Exec extends Command {
 
     // provide SDKs with useful env vars
     let env = {
-      PERCY_CLI_API: this.percy?.apiAddress(),
+      PERCY_SERVER_ADDRESS: this.percy?.address(),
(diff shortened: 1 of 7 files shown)


Robdel12 (approved):
I dig it, `PERCY_SERVER_ADDRESS` is much better 🏁

PR #257, opened 2021-03-19
✅ Fix flakey logger test

wwilsman:
## What is this?

A test logs, waits 100ms to log again, then asserts that the correct elapsed time was printed. Sometimes this logger test shows the elapsed time a little out of range of 100-109ms, so the check is loosened to 90-119ms. We're not really testing speed here, just that the elapsed time prints.

### DIFF
--- a/packages/logger/test/index.test.js
+++ b/packages/logger/test/index.test.js
@@ -265,7 +265,7 @@ describe('logger', () => {
       expect(helpers.stderr).toEqual([
         jasmine.stringMatching('Warn log \\(\\dms\\)'),
         jasmine.stringMatching('Error log \\(\\dms\\)'),
-        jasmine.stringMatching('Debug log \\(10\\dms\\)')
+        jasmine.stringMatching('Debug log \\((9[0-9]|1[01][0-9])ms\\)')
       ]);
     });
   });


packages/logger/test/index.test.js
```diff
@@ -265,7 +265,7 @@ describe('logger', () => {
       expect(helpers.stderr).toEqual([
         jasmine.stringMatching('Warn log \\(\\dms\\)'),
         jasmine.stringMatching('Error log \\(\\dms\\)'),
-        jasmine.stringMatching('Debug log \\(10\\dms\\)')
+        jasmine.stringMatching('Debug log \\((9[0-9]|1[01][0-9])ms\\')
```
wwilsman, line 268: RegExp explanation: matches either `9[0-9]` which matches `90-99`, or `1[01][0-9]` which matches `100-119`.

packages/logger/test/index.test.js
```diff
@@ -265,7 +265,7 @@ describe('logger', () => {
       expect(helpers.stderr).toEqual([
         jasmine.stringMatching('Warn log \\(\\dms\\)'),
         jasmine.stringMatching('Error log \\(\\dms\\)'),
-        jasmine.stringMatching('Debug log \\(10\\dms\\)')
+        jasmine.stringMatching('Debug log \\((9[0-9]|1[01][0-9])ms\\)')
```
wwilsman, line 268: RegExp explanation: matches either `9[0-9]` which matches `90-99`, or `1[01][0-9]` which matches `100-119`.

Robdel12 (approved):
🏁

PR #261, opened 2021-03-22
🐛 Fix cross-env remote logging handlers

wwilsman:
## Purpose

In addition to a WebSocket Server, the node `ws` library provides a way for Node environments to use a `WebSocket` client class similar to browser environments. Even though a `ws` WebSocket has identical event names to its browser counterpart, events in Node typically do not emit Event objects like in browsers. This makes it so our agnostic code must handle both types of event listeners.

## Approach

Rather than checking event listener arguments to determine if a value or event object was provided, we can use `on<eventname>` handler counterparts which are called by `ws` in _exactly_ the same way browser WebSocket handlers are called — with an event object. The main difference between listeners and handlers is that multiple event handlers cannot be registered at the same time. However, we are not registering multiple listeners per event anyway.

The test mock and tests were also updated and the `listen` util was removed since it is no longer needed. An additional cleanup handler test was also added to account for coverage changes with the util removal.

### DIFF
--- a/packages/logger/src/logger.js
+++ b/packages/logger/src/logger.js
@@ -1,4 +1,4 @@
-import { colors, listen } from './util';
+import { colors } from './util';
 
 const URL_REGEXP = /\bhttps?:\/\/[^\s/$.?#].[^\s]*\b/i;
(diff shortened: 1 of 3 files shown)


Robdel12 (approved):
🏁

PR #264, opened 2021-03-25
✅ Make @percy/logger test helper framework agnostic

wwilsman:
## Purpose

We currently use the `spyOn` helper from jasmine to spy on browser logs. This works great for tests within this repo which also uses Jasmine, however it does not work so well outside of Jasmine where `spyOn` isn't typically globally available.

## Approach

Add our own small `spy` helper within this util that takes care of the spying regardless of the testing framework. The expectation API changed and tests updated since we are no longer using a built-in spy, however those changes were small.

### DIFF
--- a/packages/logger/test/helpers.js
+++ b/packages/logger/test/helpers.js
@@ -30,6 +30,26 @@ function TestIO(data, options) {
   }
 }
 
+function spy(object, method, func) {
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
🏁

PR #266, opened 2021-03-25
🐛 Fix @percy/logger test helper unsupported syntax

wwilsman:
## What is this?

The optional chaining syntax (`?.`) is supported in Node 14, but not Node 12. So it was missed when developing `@percy/logger` test helpers. Babel config exists that is meant to ignore distributed files to raise such issues during CI, however it was not configured correctly.

This updates the babel config to use the current working directory in combination with a package's distributed files list. Unfortunately, this meant that development aliases no longer existed for some of these files. To accommodate development aliases, only the base config is not applied to distributed files. This makes it so no ES features are accidentally transpiled in these files and tests will now warn when such a file has been encountered.

Now that this is fixed, the offending `?.` was also fixed in `@percy/logger`'s test helper.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,17 +1,24 @@
-const pkg = require(`${process.cwd()}/package.json`);
+const cwd = process.cwd();
+const path = require('path');
+const pkg = require(`${cwd}/package.json`);
 
 const base = {
-  ignore: pkg.files,
-  presets: [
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
🏁 Next year when we can drop Node 12 😢

PR #279, opened 2021-03-31
✨ Expose used chrome versions

olebedev:
### Background

The `@percy/core` package [manages chromium version](https://github.com/percy/cli/blob/master/packages/core/src/install.js#L18-L23) on its own when no executables provided, so there is no versions mismatch possible in this scenario. However, when the `PERCY_BROWSER_EXECUTABLE` env var is set to point to a chromium executable it becomes difficult to manage the chromium version and keep it aligned with the ones that the library is supposed to work with. Especially, when the dependency definition in `package.json` file is set like:

```
{
  ...
  "peerDependencies": {
    "@percy/core": "*",
  },
  ...
}
```

### In this PR

Expose chromium versions to be able to assert that the local one fits the requirement. For example:

```js
const { chromium } = require('@percy/core/dist/install');
assert.equal(
  '812847',
  chromium.revisions.linux,
  'The local chromium version for @percy/core don\'t match the expected one!',
);
```

By exposing the chromium version, we can provide a guarantee that the versions are matching along with the ability to upgrade the library independently while the local and required versions are the same.

cc @joscha

### DIFF
--- a/packages/core/src/install.js
+++ b/packages/core/src/install.js
@@ -4,7 +4,7 @@ import https from 'https';
 import logger from '@percy/logger';
 import readableBytes from './utils/bytes';
 
-const chromiumRevisions = {
+const revisions = {
   linux: '812847',
   win64: '812845',
   win32: '812822',
@@ -22,7 +22,7 @@ function installChromium({
   // default directory is within @percy/core package root
   directory = path.resolve(__dirname, '../.local-chromium'),
   // default revision corresponds to v87.0.4280.x
-  revision = selectByPlatform(chromiumRevisions),
+  revision = selectByPlatform(revisions),
 } = {}) {
   let extract = (i, o) => require('extract-zip')(i, { dir: o });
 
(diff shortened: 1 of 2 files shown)


wwilsman:
Hey @olebedev! Thanks for the PR!

Only small suggestion would be to set this property as `installChromium.revisions` since these revisions only pertain to this particular function. Then you can remove the additional export and access revisions as `install.chromium.revisions`.

olebedev:
Hi @wwilsman, done in d6e879e4476f834b834740014aeefce7e6ec3a02, please have a look!

wwilsman (approved):
Thanks for the contribution!

Our CI events weren't set up properly for forks at the time you forked, so I'll just go ahead and merge this anyway with pending checks since I don't foresee any issues from this change. 😄

joscha:
Thanks @wwilsman! Can you ping here when this goes into a new version, please?

Robdel12:
Boom went out in 1.0.0-beta.47 (https://github.com/percy/cli/releases/tag/v1.0.0-beta.47) @joscha / @olebedev :D thanks for the PR!

PR #282, opened 2021-04-02
✨ Add AMD support to bundles

wwilsman:
## What is this?

Adds support for AMD format in bundles to allow imports in those environments (such as Ember).


Robdel12 (approved):
🏁 haha I wonder what else uses amd 🤔

PR #283, opened 2021-04-02
✨ sdk-utils improvements

wwilsman:
## What is this?

Two small `@percy/sdk-utils` improvements.

- Grab fetch references before they might be mocked
- Clone the document during mock dom transformations


Robdel12 (approved):
🏁

PR #284, opened 2021-04-02
🐛 Account for logger websocket synchrony

wwilsman:
## What is this?

Sometimes, WebSocket errors can come from the constructor itself which is defined outside the scope of the logger. Rather than having extra error handling, we can have the logger accept a socket creator function and wrap it with the existing error handling.

Usage of the `remote` method was updated in `@percy/sdk-utils` as well as other tests.

### DIFF
--- a/packages/core/test/server.test.js
+++ b/packages/core/test/server.test.js
@@ -172,16 +172,13 @@ describe('Snapshot Server', () => {
     // log from a separate async process
     let [stdout, stderr] = await new Promise((resolve, reject) => {
       require('child_process').exec('node -e "' + [
-        "const logger = require('@percy/logger');",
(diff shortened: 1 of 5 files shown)


Robdel12 (approved):
🏁

wwilsman:
Weird. Looks like automerge kicked in even though there were failing checks

Robdel12:
Not marked required are they?

wwilsman:
No they are required. It must've been a race condition where the checks showed as passing before previously passing checks reran and failed

wwilsman:
Broken test was a silly windows issue fixed with https://github.com/percy/cli/commit/3c94d7e567a849402dff707d8497385ecc0a510f

PR #288, opened 2021-04-05
✨ Better GitHub Action env support

wwilsman:
## What is this?

This adds better support for GH Actions by reading from the event payload file created in workflows by their corresponding webhook events.

In our official GH Actions, we use `@actions/github` to read environment information such as the PR number and branch name. This package (and specifically the properties used in our actions) are actually a thin wrapper around [reading from the file](https://github.com/actions/toolkit/blob/fc005283374f1cea5e03ea1caf959cd9ff12fdc2/packages/github/src/context.ts#L27-L36) located at `GITHUB_EVENT_PATH`.

We can eliminate the need for specific actions, and that library, by reading from this event file directly ourselves.

This is done with a new util function. The existing utils were all `git` based, so the file was named `git.js`. However with the addition of this util, the file exports vary and so was renamed to `utils.js`

### DIFF
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -1,5 +1,5 @@
 import PercyEnvironment from '@percy/env';
-import { git } from '@percy/env/dist/git';
+import { git } from '@percy/env/dist/utils';
 import pkg from '../package.json';
 
 import {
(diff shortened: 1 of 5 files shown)


Robdel12 (approved):
🏁 Hmmmmm, 🔪 our actions now? 🤨

PR #289, opened 2021-04-05
🐛 Fix printed debug timestamps

wwilsman:
## What is this?

The printed debug messages are meant to show the time elapsed between logs. However, when it was implemented by saving the last timestamp accidentally with a logical or assignment, which means that it will only ever be set on the first log and never updated again. Removing the `||` fixes the accumulative timestamp and will show proper elapsed-between timestamps.

This might also relate to this specific test being flakey in CI. Probably because the time between the first log and last log was slow even though we only want to test the time between the last log and the previous log. An additional log was added to test that the timestamp is **not** accumulative.


Robdel12 (approved):
🏁 Slick fix

PR #293, opened 2021-04-07
♻️ Update `build:wait` example output

Robdel12:
## What is this?

The `build:wait` command requires a full org/project slug and the example only gave the project slug (that will error). This PR updates the example to give an org/project slug.

### DIFF
--- a/packages/cli-build/src/commands/build/wait.js
+++ b/packages/cli-build/src/commands/build/wait.js
@@ -46,7 +46,7 @@ export class Wait extends Command {
   };
 
   static examples = [
-    '$ percy build:wait --build 123',
+    '$ percy build:wait --build 2222222',
     '$ percy build:wait --project org-slug/test --commit HEAD'
   ];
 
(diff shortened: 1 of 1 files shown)


packages/cli-build/src/commands/build/wait.js
```diff
@@ -46,8 +46,8 @@ export class Wait extends Command {
   };
 
   static examples = [
-    '$ percy build:wait --build 123',
-    '$ percy build:wait --project test --commit HEAD'
+    '$ percy build:wait --build 2222222',
+    '$ percy build:wait --project org-slug/test --commit HEAD'
```
wwilsman, line 50: Should we do `org/project` or `org-slug/project-slug`?

Robdel12: I think so 💯

wwilsman:
```suggestion
    '$ percy build:wait --project org/project --commit HEAD'
```

wwilsman: Just committed it from the UI since you agree. 😆

wwilsman (approved):
🎮

PR #294, opened 2021-04-08
Set Parallel Total from CircleCI Config

drewmendenhall:
When setting up [`parallelism` in CircleCI](https://circleci.com/docs/2.0/parallelism-faster-jobs/), an environment variable [`CIRCLE_NODE_TOTAL`](https://circleci.com/docs/2.0/env-vars/?section=pipelines#built-in-environment-variables) is set.

This change picks up the above when detecting the parallel build total for the CircleCI environment.


drewmendenhall:
Looks like this was already in place and removed in #51.

More context: https://github.com/percy/percy-cypress/issues/311#issuecomment-815445398

PR #303, opened 2021-04-13
✨ Improved config migrate

wwilsman:
## What is this?

This improves the config migration pattern to support migrating deprecated options within the current config version. To support this, migrations are always performed. As such, they no longer result in a fresh object, but instead mutate the provided config using util functions. Internally, the provided config is never a reference in danger of harmful mutations.

Previously, only a `set` util was provided. Now there is also `map` and `del` which can map and delete config options respectively. A fresh logger is also provided so migrations can log deprecation warnings.

This shouldn't break anything externally unless a 3rd party CLI plugin is already relying on beta config migration patterns.


Robdel12 (approved):
😍 🏁

PR #304, opened 2021-04-13
♻ Move snapshot discovery options

wwilsman:
## What is this?

This moves snapshot specific discovery options to a new `discovery` key. This will hopefully avoid confusion where a user may think that some options are used when rendering when they are actually only used for asset discovery.

For config files, these options were moved over to the existing, top-level, `discovery` key since all config file options are considered global and `snapshot.discovery` options are merged with global `discovery` options anyway.

Deprecation messages were added where appropriate and are planned to be removed once the CLI leaves beta.

I also added a couple missing logs for snapshot options.

## Questions

This opens up a very simple way for us to implement per-snapshot allowed hostnames, network idle timeout, and cache options. Should we (in another PR)?


Robdel12 (approved):
🏁 Much needed. Good to start making it clear which is the rendering env and which is purely for SDKs

PR #306, opened 2021-04-14
✨ Add per-snapshot discovery cache and hostnames options

wwilsman:
## What is this?

This adds `disableCache` and `allowedHostnames` to the new per-snapshot `discovery` options. Some refactoring was done to make future options much simpler to add:

- At the `percy.snapshot` method, specific options (including per-snapshot discovery options) are passed along to the `discoverer.gatherResources` method. This method then explicitly provides other methods with appropriate options; such as providing size/javascript/network options to the `page` method and discovery options to the various discovery handlers.

  Now, `gatherResources` and subsequent intermediary functions will only reference options as needed and then pass all other options along down the call chain. Any option can then be plucked where needed — such as these new discovery options only needing to be added to the discovery handlers, or in the future a user-agent option can be added to the page creation method.

- Page creation was also refactored slightly. All page configuration events were moved to the page init method which is called by the discoverer when requesting a new page from the browser. This is just so all page configuration options can live inside of the page class and not need to be handled by the discoverer instance.

- The `domainMatch` util was made to accept an array of patterns so we can avoid concatenating all hostnames.

### DIFF
--- a/packages/core/src/discovery/discoverer.js
+++ b/packages/core/src/discovery/discoverer.js
@@ -18,39 +18,15 @@ export default class PercyDiscoverer {
 
   #cache = new Map();
 
-  constructor({
-    // asset discovery concurrency
-    concurrency = 5,
-    // additional allowed hostnames besides the root URL hostname
-    allowedHostnames,
-    // how long to wait before the network is considered to be idle and assets
-    // are determined to be fully discovered
-    networkIdleTimeout,
-    // disable resource caching, the cache is still used but overwritten for each resource
-    disableCache,
-    // request headers such as auth tokens
-    requestHeaders,
-    // basic auth credentials
-    authorization,
-    // browser launch options
-    launchOptions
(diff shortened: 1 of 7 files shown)


packages/core/src/discovery/discoverer.js
```diff
@@ -65,72 +41,41 @@ export default class PercyDiscoverer {
...
-      page.send('Emulation.setDeviceMetricsOverride', { deviceScaleFactor, mobile, height, width })
-    ]);
-
-    return page;
+  async page(options) {
+    let { requestHeaders, authorization, networkIdleTimeout } = this.config;
+
+    return this.browser.page({
+      ...options,
+      requestHeaders: { ...requestHeaders, ...options.requestHeader },
```
Robdel12, line 49: Nice on the pluralization

Robdel12 (approved):
🏁

PR #307, opened 2021-04-14
✨ Allow setting discoverer user-agent

wwilsman:
## What is this?

Adds the ability to set a custom user-agent string via a config file or per snapshot option, `discovery.userAgent`.

By default, it will use the normal Chrome user agent without the `Headless` part in an attempt to prevent servers from blocking headless requests from asset discovery. SDKs can gather a more suitable user agent from the browser where the DOM was captured from and provide this user agent to the CLI automatically.

### DIFF
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -51,6 +51,9 @@ export const schema = {
           password: { type: 'string' }
         }
       },
+      userAgent: {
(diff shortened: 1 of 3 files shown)


Robdel12 (approved):
🏁 It'll be very nice to no longer have asset discovery trolled by this

PR #327, opened 2021-05-10
🐛 Use correct Chromium download path for win32

wwilsman:
## What is this?

After a revision somewhere around ~600000, zip files for the win32 platform dropped the 32 from their suffix. Before the recent-ish Chromium upgrade, we actually used a revision from before this change. Now that we have passed this range, we shouldn't need the 32 suffix unless Chromium changes again.


Robdel12 (approved):
🏁 Should we run a nightly cron or something to make sure these paths exist still? Feels not great to have it shift on us and we have no idea until its reported. Not sure tbh

PR #342, opened 2021-05-18
🐛 Do not serialize frames without document elements

### LINKED ISSUE #329: Substantial amount of random errors snapshotting static pages
See [this CI run](https://gitlab.com/mitar/mitarpage/-/jobs/1252371823). I see many errors like:

```
[percy] Encountered an error for page: http://localhost:40611/post/146314072245/do-publicly-owned-planned-economies-work/index.html
[percy] TypeError: Cannot read property 'outerHTML' of null
    at serializeDOM (<anonymous>:206:33)
    at serializeFrames (<anonymous>:87:28)
    at serializeDOM (<anonymous>:187:7)
    at serializeFrames (<anonymous>:87:28)
    at Object.serializeDOM (<anonymous>:187:7)
    at <anonymous>:5:35
    at withPercyHelpers (<anonymous>:9:14)
```

I am surprised because all pages are more or less the same (blog posts) so there is not really a reason that one would fail and another one not.

Is there any way to get those errors down?

wwilsman:
## Purpose

It seems that in some circumstances an iframe's `documentElement` may be null which then causes serialization errors. After some research, it appears this property may be null when accessed before the document has finished loading, if the document does not contain an element such as empty documents, or if a script purposefully unsets the property.

## Approach

Add another check, in addition to the `contentDocument` check, for `documentElement`. Tests are performed by unsetting the property before serialization.

Fixes #329

### DIFF
--- a/packages/dom/src/serialize-frames.js
+++ b/packages/dom/src/serialize-frames.js
@@ -12,13 +12,13 @@ export default function serializeFrames(dom, clone, { enableJavaScript }) {
     if (clone.head.contains(cloneEl)) {
       cloneEl.remove();
 
-    // if the frame document is accessible, we can serialize it
-    } else if (frame.contentDocument) {
+    // if the frame document is accessible and not empty, we can serialize it
+    } else if (frame.contentDocument && frame.contentDocument.documentElement) {
       // js is enabled and this frame was built with js, don't serialize it
-      if (enableJavaScript && builtWithJs) { continue; }
+      if (enableJavaScript && builtWithJs) continue;
 
       // the frame has yet to load and wasn't built with js, it is unsafe to serialize
-      if (!builtWithJs && !frame.contentWindow.performance.timing.loadEventEnd) { continue; }
+      if (!builtWithJs && !frame.contentWindow.performance.timing.loadEventEnd) continue;
 
       // recersively serialize contents
       let serialized = serializeDOM({
--- a/packages/dom/test/serialize-frames.test.js
+++ b/packages/dom/test/serialize-frames.test.js
@@ -22,6 +22,10 @@ describe('serializeFrames', () => {
         this.document.body.innerHTML = '<p>made with js src</p>'
       )"></iframe>
       <iframe id="frame-js-no-src"></iframe>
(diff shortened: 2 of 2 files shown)


Robdel12 (approved):
🏁 Nice, it'll be nice in the future when we can mark and call some of these things out in the UI

mitar:
Would there be a way instead to wait for iframe to load? I worry that this issue shows that `--network-idle-timeout` does not work for iframes. Otherwise all iframes would load, no, before snapshot is made and we could not have the case that "it appears this property may be null when accessed before the document has finished loading" which I think is the main reason for this issue?

wwilsman:
We use Chrome's DevTools Protocol under the hood to monitor requests for the entire page. This means the network idle timeout does indeed work for frames within the page (I even tested it myself just now just in case this was a Chromium bug).

> "it appears this property may be null when accessed before the document has finished loading" which I think is the main reason for this issue?

That quote is actually only the beginning of that sentence. The entire sentence lists three different scenarios in which the document element could be null. In addition to the scenario you quoted, it can also be null if the document does not contain an element such as empty documents, or if a script purposefully unsets the property. In the PR description, and in the tests of this PR, you can see I was able to reproduce this by unsetting the property. Third party scripts loaded within embedded iframes might also do this, intentionally or not.

> And so this PR might lead to flaky screenshots, where sometimes iframe is visible and included and sometimes it is not

Without this PR, the result is no snapshot at all for a page encountering this issue. Furthermore, the frame will still be captured and rendered in our infrastructure, however any flakiness would instead be due to serialization or lack thereof of forms or canvases within the iframe. 

If you're experiencing blank iframes, it could be due to JavaScript being disabled when re-rendering in our own browsers. Network idle won't really help in capturing such iframes in time. Instead you'll have to rely on the other wait options to wait for the iframe to finish rendering before the snapshot is taken.

mitar:
> The entire sentence lists three different scenarios in which the document element could be null. 

Sure. I do not want to argue here. You are right about all other scenarios. But just in my concrete case it looks like things just not being loaded and not that iframes would be removing anything, for example.

In any case, this PR does fix the main problem which can get me going further and I can report back if I do see any issues with flakiness instead of me just speculating. :-)

Thanks!

PR #345, opened 2021-05-21
🐛 Default host port for proxy requests

wwilsman:
## What is this?

While the URL `host` is supposed to contain the `hostname` and `port`, it actually does not always. Similarly, sometimes `port` can be missing as well. In the proxy agent, we were relying on `host`,  but need to actually use `hostname` and `port` separately while also using a default port. This default port does get picked up by coverage since our `localhost` test URLs use specific ports.

This PR also adds proxy agent utilization to browser installation to allow browser downloads to flow through proxies as well.

### DIFF
--- a/packages/client/src/request.js
+++ b/packages/client/src/request.js
@@ -90,10 +90,9 @@ export class ProxyHttpsAgent extends https.Agent {
     };
 
     // write proxy connect message to the socket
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
🏁 very nice

PR #349, opened 2021-05-24
🔊 Fix debug log logging the wrong variable

wwilsman:
## What is this?

A small fix for a debug log logging the wrong variable.


Robdel12 (approved):
🏁 whoops

PR #350, opened 2021-05-24
🐛 Stop core config reloading when already loaded via the CLI

wwilsman:
## What is this?

I've noticed (especially without a config file present) that the config debug logs will log twice. This is because we load the config in the CLI using `this.percyrc()` and core also auto loads the config itself.

By automatically setting the `config: false` option in `this.percyrc()`, when passed to core it will not reload the config file and instead merge defaults with the provided options already loaded from the file.


Robdel12 (approved):
🏁

PR #351, opened 2021-05-24
✨ Add retry options to internal request util

wwilsman:
## What is this?

When utilizing this util outside of client, it can be useful to configure retry options. These options are available, but never used by client itself since the defaults are acceptable.


Robdel12 (approved):
🏁

PR #352, opened 2021-05-24
🐛 Fix asset discovery issue with event streams

wwilsman:
## What is this?

[Server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events) can clog up asset discovery because the request is left open. From my research, it seems like this would only be a problem when JS is enabled. From experience writing the new Storybook SDK, it happens with hot module reloading; which lines up with JS needing to be enabled.

We can handle the [`Network.eventSourceMessageReceived`](https://chromedevtools.github.io/devtools-protocol/tot/Network/#event-eventSourceMessageReceived) devtool event and remove the open request from being tracked. This prevents asset discovery from hanging on network idle while still waiting for at least one event before considering the network to be idle. This also prevents us from capturing these event streams which might lead to flakey screenshots when rerendering.


Robdel12 (approved):
🏁 A good number of fixes coming from the storybook work. Love it!

PR #357, opened 2021-05-27
✨ Add additional retry option to client request util

wwilsman:
## What is this?

In other SDK related packages, we occasionally borrow this request util for its retryability. While using it in the new Storybook SDK to wait for the Storybook server to become available, I discovered that sometimes it can be useful to retry 404s when the file will eventually exist (if the server starts before code finishes compiling).

This PR adds a `retryNotFound` option to the request util to enable this behavior since we don't normally want to retry 404s in most situations. Since this option will not be utilized internally by client, no tests were added here. If we add much more options to this that aren't used internally, we may want to add a unit test (or in the future move shared utils to a shared package).

### DIFF
--- a/packages/client/src/request.js
+++ b/packages/client/src/request.js
@@ -115,9 +115,11 @@ export class ProxyHttpsAgent extends https.Agent {
 }
 
 // Returns true or false if an error should cause the request to be retried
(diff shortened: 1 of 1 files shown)


Robdel12 (approved):
🏁

PR #370, opened 2021-06-17
✨ Allow setting client's API url via an environment variable

wwilsman:
## What is this?

In the world of `@percy/agent`, setting the `PERCY_API` environment variable allowed you to override the URL used by client to communicate with the Percy API. That behavior was lost when moving to the new `@percy/cli` world, but can be useful when manually testing or interacting with a local development API.

With client's proxy agent, we indiscriminately proxy all API requests through HTTPS since our production API server is guaranteed to be HTTPS. However, with the potential to change the API URL to a locally hosted URL, it might not always be HTTPS. So we need to check the protocol of the API URL to disable the HTTPS proxy agent and allow the `request` util to default to the appropriate agent based on the request's protocol.

This shouldn't be used by end users, so `PERCY_API` felt too user friendly. I decided on `PERCY_CLIENT_API_URL` (to be hyper specific) but am open to changing it.

### DIFF
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -6,6 +6,9 @@ import pkg from '../package.json';
 import { sha256hash, base64encode, pool } from './utils';
 import request, { ProxyHttpsAgent } from './request';
 
+// Default client API URL can be set with an env var for API development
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
🏁

djones:
👍🏻 Thanks for putting this together @wwilsman

PR #372, opened 2021-06-17
🐛 Fix unintentionally removing the default queue concurrency

wwilsman:
## What is this?

With recent changes from #369, when `config.discovery.concurrency` is not defined, it is still being set on the queue causing the default of 5 to be removed. This enables infinite concurrency which is not what we want since it causes a huge slowdown as more and more resources are consumed in parallel


Robdel12 (approved):
> This enables infinite concurrency which is not what we want since it causes a huge slowdown as more and more resources are consumed in parallel

🏖️ 🏀 hahahah

🏁

PR #373, opened 2021-06-17
♻️ Post core refactor tweaks

wwilsman:
## What is this?

Following #369 and subsequently #372, I've since found a couple other minor thing that need to be tweaked.

- In agent, up to 10 browser pages were pooled to be running at once. It was lowered in the CLI since moving to a proper queue with a concurrency. Now that queue performance has gotten much better, let's up it to 10 again. Even in previous versions of the CLI up to 10 browser pages could be running when using the capture method, so being able to set this more accurately is a win from #369.

- Rename `queue.pause()` to `queue.stop()` (🚲 🏠). Pause implies that there is still more to run, while stop implies that it won't run regardless if there is more to run. Stop also feels more complimentary to `run()`.

- Add ability for `close()` to abort. This will stop and clear the queue while closing it, exactly what is really needed by the `percy.close()` method.

- Abort queue when force stopping the Percy instance. Rather than clear the queues which will be closed, when force stopping we can utilize `percy.close()` to wholesale abort any currently queued tasks. 

- Catch sync errors when running queued tasks. While wrapping the dequeued callback in a resolve served to catch async errors, sync errors could still happen which would cause pending tasks to never be cleared.

- While testing Storybook and sending a SIGTERM event, I noticed that there were occasional unhandled promise rejections caused by page/browser closed errors bubbling from random event callbacks. The closed errors always bubble and get raised from other methods, so in these few unhandled rejections, we can simply log a debug message to handle it. Coverage was skipped for these catches since they all only occur in specific race conditions.

### DIFF
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -28,7 +28,9 @@ export function createRequestHandler({ disableCache, getResource }, m) {
       log.debug(`Encountered an error handling request: ${url}`, meta);
       log.debug(error);
 
-      await request.abort(error);
+      /* istanbul ignore next: race condition */
+      await request.abort(error)
+        .catch(e => log.debug(e, meta));
     }
   };
 }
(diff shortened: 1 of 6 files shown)


Robdel12 (approved):
No need for 🚲 🏠 , I dig it. Nice 🏁

Issue #296, opened 2021-03-26
Add basic parallel nonce support for GitHub Actions


nicholaschiang:
Right now, there doesn't [seem to be support](https://github.com/percy/percy-js/blob/728244da6815dbdb365ae1c44eebd01fa37d1d8a/src/environment.js#L248-L273) for GitHub Actions without having to configure things ourselves.

Right now, [this is my current configuration](https://github.com/tutorbookapp/tutorbook/blob/develop/.github/workflows/tests.yml#L9-L10) (might be helpful when creating built-in support):

```yaml
env:
  NODE_ENV: test
  PERCY_PARALLEL_NONCE: ${{ github.run_id }}
  PERCY_PARALLEL_TOTAL: 10
```

wwilsman:
For completeness, if you do run into issues with retries you'll have to create a more unique nonce at the start of the workflow, and then persist it throughout your steps/jobs. For example:

```yaml
# ...

env:
  PERCY_PARALLEL_TOTAL: 10

jobs:
  # run this job first
  nonce:
    runs-on: ubuntu-latest
    # persist job results to other jobs in the workflow
    outputs:
      result: ${{ steps.nonce.outputs.result }}
    steps:
    # persist step results to other steps in the job
    - id: nonce
      # adding a timestamp makes the nonce more unique for re-runs
      run: echo "::set-output name=result::${{ github.run_id }}-$(date +%s)"
  
  # other jobs then rely on the first job with `needs`
  run-tests:
    matrix: # ...
    runs-on: ${{ matrix.os }}
    needs: nonce
    env: 
      PERCY_PARALLEL_NONCE: ${{ needs.nonce.outputs.result }}
    steps:
    - # ...
```

(see [output parameters](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-output-parameter) & [`jobs.<job_id>.outputs`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idoutputs))

Robdel12:
This makes me wonder if we want to adopt an automatic nonce -- I don't think so. I think our docs should outline that you'll have to create a (reliable & proper) nonce like your example @wwilsman. 

We have this same issue with TravisCI and it did troll a lot of folks with their parallel builds (they reuse the same ID on rebuilds -- breaks Percy).

wwilsman:
[...]

PR #375, opened 2021-06-18
✨  Add basic parallel nonce support for GitHub actions

Robdel12:
## What is this?

This will at least get the happy path of parallel builds going. We'll need an advanced doc to highlight the issues with this (`GITHUB_RUN_ID` is not unique on rebuilds, which will cause Percy to fail). Touches on #296, but doesn't completely solve it.

### DIFF
--- a/packages/env/test/github.test.js
+++ b/packages/env/test/github.test.js
@@ -22,7 +22,7 @@ describe('GitHub', () => {
 
     env = new PercyEnvironment({
       PERCY_PARALLEL_TOTAL: '-1',
-      PERCY_PARALLEL_NONCE: 'job-id',
+      GITHUB_RUN_ID: 'job-id',
       GITHUB_ACTIONS: 'true',
       GITHUB_EVENT_PATH: ghEventFile
     });
--- a/packages/env/test/github.test.js
+++ b/packages/env/test/github.test.js
@@ -21,6 +21,8 @@ describe('GitHub', () => {
     }));
 
     env = new PercyEnvironment({
+      PERCY_PARALLEL_TOTAL: '-1',
+      PERCY_PARALLEL_NONCE: 'job-id',
       GITHUB_ACTIONS: 'true',
       GITHUB_EVENT_PATH: ghEventFile
     });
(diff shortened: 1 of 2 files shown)


packages/env/test/heroku.test.js
```diff
@@ -23,4 +23,9 @@ describe('Heroku', () => {
     expect(env).toHaveProperty('parallel.nonce', 'heroku-test-run-id');
     expect(env).toHaveProperty('parallel.total', -1);
   });
+
+  it('has the correct parallel nonce in 2.x', () => {
```
wwilsman, line 27: Looks like this is already included in the test above

packages/env/test/github.test.js
```diff
@@ -50,6 +50,7 @@ describe('GitHub', () => {
   describe('without an event payload', () => {
     beforeEach(() => {
       env = new PercyEnvironment({
+        PERCY_PARALLEL_TOTAL: '-1',
```
wwilsman, line 53: Similar to all the other tests, let's just add common env vars such as this and `GITHUB_RUN_ID` to the top-level before each in this entire suite and fix the assertions that are already in the "has the correct properties" test. Especially since the run id will probably always be present.

packages/env/test/github.test.js
```diff
@@ -62,5 +63,10 @@ describe('GitHub', () => {
       expect(env).toHaveProperty('commit', 'gh-env-sha');
       expect(env).toHaveProperty('branch', 'gh-env-branch');
     });
+
+    it('has the correct parallel nonce in 2.x', () => {
```
wwilsman, line 67: What is `2.x` in this context?

wwilsman (changes requested):

packages/env/test/github.test.js
Robdel12: You know when I ripped it from CircleCI, I thought it was referring to the client. But I think it means circleci 2? No tests for 1.x in circle tho. 

But yea, removed, it doesn't make sense.

packages/env/test/github.test.js
```diff
@@ -21,6 +21,8 @@ describe('GitHub', () => {
     }));
 
     env = new PercyEnvironment({
+      PERCY_PARALLEL_TOTAL: '-1',
+      PERCY_PARALLEL_NONCE: 'job-id',
```
wwilsman, line 25:
```suggestion
      GITHUB_RUN_ID: 'job-id',
```

wwilsman (changes requested):

packages/env/test/github.test.js
Robdel12: This is what happens when you get excited to pick your dog up from the vet and rush commit 😆

wwilsman (approved):
🎮

PR #376, opened 2021-06-18
✨ Better progress logging

wwilsman:
## Purpose

There are two logs in particular that would be nice to have some progress associated with them. The processing snapshot queue log and the snapshot uploading queue log.

When SDKs such as Storybook or the snapshot command quickly take and process snapshots via asset discovery, the end result could be waiting for hundreds of snapshots to upload. To prevent clogging up processes with hundreds of in-flight requests at once, uploads are queued just as asset discovery is. 

This results in a single log at the end of snapshot processing, before finalizing the build, when snapshots are uploading: `Waiting for X snapshots to finish uploading`. With #369, the message was changed to be more generic: `Uploading snapshots...`. Which is arguably less helpful, but in either case the CLI could be waiting for up to a minute while snapshots are uploaded. Meanwhile there is no feedback from the user's perspective.

## Approach

During the refactor, while updating the CLI build plugin to work with client changes, I noticed that the code we use to log progress for the wait command was fairly simple and straightforward; and could be adapted to the above use case. The core pieces of progress logging code was moved into a new `logger.progress()` method and exposed for loggers. This method will not capture message to the internal log store since it is meant for visual purposes and not for debugging.

Where the wait command would always clear the line and not log progress before finished or failed logs were printed, we can take a slightly different approach to achieve the same effect. Rather than always clear the entire line, we can clear the line to the right of the cursor after printing a new message over the existing line. This helps to prevent some flicker between when the line was cleared and when new logs are printed. 

For the snapshot queue logs, snapshots can continue to be taken once the queue is awaited one. The build will only be finalized once both the snapshot and upload queues have completely emptied. Because other logs can be printed while waiting on the snapshot queue, a `persist` option can be provided which will re-print the progress log after other logs clear it.

I also updated the install script progress logger to use our new method. Likewise with the build command that inspired the method, and even the browser revision script.

## Preview

https://user-images.githubusercontent.com/5005153/122619596-68dcd580-d056-11eb-8367-1603488d26a1.mp4

(sped up a little for brevity)

### DIFF
--- a/packages/cli-build/src/commands/build/wait.js
+++ b/packages/cli-build/src/commands/build/wait.js
@@ -1,4 +1,3 @@
-import readline from 'readline';
 import Command, { flags } from '@percy/cli-command';
 import PercyClient from '@percy/client';
 import logger from '@percy/logger';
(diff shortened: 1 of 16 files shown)


Robdel12 (approved):
Drops a dependency _and_ makes the UX better? You’re on a roll 🏁

PR #382, opened 2021-06-21
✨ Allow snapshotting a list of pages to utilize the base-url flag

wwilsman:
## What is this?

This PR refactors and tweaks the `percy snapshot` command to be a little more friendly.

First, a page listing no longer needs to be a list of objects containing at least a `url` property. Now the file can contain just a list of URLs in the appropriate format:

```yaml
# snapshots.yml
- https://example.com
- https://example.com/about
- https://example.com/contact
```

Second, the `--base-url` flag was made to work with page listings in addition to the existing static directory usage. Since usage differs between page listings and static directories however, some other small adjustments were necessary.

For static directories, the base URL should be the base **_PATHNAME_** that the site would be hosted at. Most sites are hosted at the domain root (`/`), however some sites might be hosted at other paths (e.g. `/blog`). This requirement has been present since the `@percy/agent` snapshot command and continues to remain a requirement.

For page listings, the base URL should be the _entire base URL_ including the **protocol** and **hostname**. For example, the above list of URLs can be simplified to remove the common base URL:

```yaml
# snapshots.yml
- /
- /about
- /contact
```

```sh
$ percy snapshot --base-url https://example.com snapshots.yml
```

To prevent mixing up the `static.base-url` configuration option with the page listing flag, a couple other changes were also made. The flag is no longer auto-mapped to the configuration option via the `percyrc` flag property. Instead, it is passed along to the `.percyrc()` method which now accepts an initial overrides argument. 

The `static.base-url` configuration option was no longer being checked for the leading forward slash (only the flag is checked by the command now). A pattern requirement was added to the schema, and custom errors messages were added to config validation to print a better error message than what is normally printed for pattern failures. We can also utilize this pattern to do other better config validations in the future such as for certain discovery options.

The `--dry-run` flag will now print a better list of snapshots, however the default `name` in core is only created when calling the snapshot method. The default name was also added here so dry runs can properly print it. When adding the default name here, I discovered a small bug in the core implementation in which `uri.path` needed to be replaced with `uri.pathname`.

### DIFF
--- a/packages/cli-command/src/command.js
+++ b/packages/cli-command/src/command.js
@@ -48,7 +48,7 @@ export default class PercyCommand extends Command {
   // respective `percyrc` parameter. The flag input is then merged with options
   // loaded from a config file and default config options. The PERCY_TOKEN
   // environment variable is also included as a convenience.
-  percyrc() {
+  percyrc(initialOverrides = {}) {
     let flags = Object.entries(this.constructor.flags);
     let overrides = flags.reduce((conf, [name, flag]) => (
       flag.percyrc?.split('.').reduce((target, key, i, paths) => {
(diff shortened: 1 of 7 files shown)


Robdel12 (approved):
Slick! And then we can ship the new snapshot docs 🏁

Issue #374, opened 2021-06-18
Allow setting cookies for asset discovery


wwilsman:
While cookies are sent as headers, it appears as though providing a `Cookie` with the `discovery.requestHeaders` option does not work as expected. 

It looks like there is a [CDP method](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-setCookies) that allows setting cookies. We should add support for this via `discovery.cookies`.

What should the syntax look like? Mirror the headers option and allow `key: value` cookies? Or mirror the CDP API and allow an array of cookie param objects? Both (if object, a, if array, b)?

wwilsman:
Shipped in `1.0.0-beta.53`

PR #383, opened 2021-06-22
✨ Allow setting discovery browser cookies

wwilsman:
## Purpose

Sometimes, cookies sent via a `Cookie` header don't always seem to work (#374). While I couldn't exactly reproduce a scenario where the header wouldn't work, I thought it would be helpful to have a dedicated `cookies` option anyway. The dedicated option can help avoid confusion for people who are unfamiliar with the associated header, and also allow setting cookies for specific domains or paths.

## Approach

The config schema was updated to allow a `discovery.cookies` option that can either be an object containing key-value pairs of cookies, or an array containing [cookie parameters](https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-CookieParam) with `name`, `value`, and other cookie properties.

While it was straightforward to add cookies within page initialization, I discovered that a cookie's `url` or `domain` property is also required. The cookies' object-to-param mapping was moved up into snapshot config merging so we could default a snapshot's cookie `url` to the snapshot's own URL. 

However, I then also discovered that even when used within a page session, cookies are stored for the entire browser session. This fact made it difficult to reason about whether to override clashing cookies, or clear cookies for each page. Either scenario would not be viable for concurrent pages with differing cookies.

It was then decided that cookies could not be made into per-snapshot options. Cookie normalization was moved into the browser class. The browser class was also updated to move launch options into class initialization. This makes it so some SDKs, like Storybook, can launch the browser without needing to provide redundant launch options.

Cookies still cannot be set until a page URL is known, since we need to account for default cookie URLs. If we ignored the url/domain requirement and set cookies eagerly, we would not be able to accept a key-value object and users would also need to know the url/domain in advance in their config; which makes the config less reusable for projects with dynamic staging URLs. Due to this, cookies are set (if provided) whenever a page is navigated to, before actual page navigation occurs.

## Other changes

While cookies were allowed as per-snapshot options, the per-snapshot debug log output would be merged with global config options. While it would be nice to include global config in per-snapshot logs, it can be done in a much better way in a follow up PR if desired (using log filtering). Now only provided options are logged, without merged config.

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -52,6 +52,7 @@ The following options can also be defined within a Percy config file
   - `authorization` — Basic auth `username` and `password` for protected snapshot assets
   - `disableCache` — Disable asset caching (**default** `false`)
   - `userAgent` — Custom user-agent string used when requesting assets
+  - `cookies` — Browser cookies to use when requesting assets
   - `networkIdleTimeout` — Milliseconds to wait for the network to idle (**default** `100`)
   - `concurrency` — Asset discovery concerrency (**default** `5`)
   - `launchOptions` — Asset discovery browser launch options
--- a/packages/core/src/browser.js
+++ b/packages/core/src/browser.js
@@ -59,6 +59,10 @@ export default class Browser extends EventEmitter {
     this.executable = executable;
     this.headless = headless;
     this.args = args;
+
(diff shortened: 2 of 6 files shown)


Robdel12 (approved):
🏁

PR #387, opened 2021-06-22
🐛 Fix percy-specific css resource url

wwilsman:
## What is this?

Resource URLs should include the root origin, otherwise our infrastructure proxy fails to properly proxy the request over secure connections. This behavior is likely due to a default origin somewhere being http attempting to validate against https.

This PR takes advantage of `new URL(input, base)` behavior, where the origin of the second argument is used as the base of the input when the input is an absolute or relative pathname.

### DIFF
--- a/packages/core/src/percy.js
+++ b/packages/core/src/percy.js
@@ -322,14 +322,13 @@ export default class Percy {
         for (let width of widths) await page.resize({ width });
 
         // create and add a percy-css resource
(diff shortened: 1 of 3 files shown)


Robdel12 (approved):
🏁 while I’m happy we no longer have a full suite of only E2E tests on Percy, I think this might be a good argument for a few smokes

wwilsman:
I've always wanted to consolidate our old e2e tests into a single example app at the top level of this repo. Just hasn't been a huge priority since we have 100% test coverage otherwise

PR #390, opened 2021-06-23
🐛 Always ignore existing root resources during asset discovery

### LINKED ISSUE #388: Errors attempting to upload multiple resources with the same resource url
We're using `@percy/cli` (`1.0.0-beta.54`) together with `@percy/puppeteer` (`2.0.0`) and are facing a situation where Percy errors uploading screenshots because resource URLs are identical.

Here's an example scenario where this happens:

```js

await page.goto('http://localhost:8000');

await percySnapshot(page, 'With JS');

await page.setJavaScriptEnabled(false);

await page.goto('http://localhost:8000');

await percySnapshot(page, 'Without JS');

await page.setJavaScriptEnabled(true);
```

But this can of course also happen when taking multiple screenshots in an SPA where the URL never changes:

```js

await page.goto('http://localhost:8000');

await percySnapshot(page, 'Initial state');

await page.click('#button_that_opens_modal');

await percySnapshot(page, 'Modal open');
```

Is this something you can fix? If not, any advice on how to prevent this on our side?

wwilsman:
## What is this?

In asset discovery when the cache is disabled, it doesn't matter if the `getResource` option returns an existing resource. The existing resource will always be overridden with new response. However, root resources aren't normally included with other resources. So when a root resource requests itself (such as a canonical link) and cache is disabled, a new resource is created for the root response (but `is-root` is not set since it happens during normal discovery). Later, when the root resource is added to the discovered resources, it ends up in the list of resources twice.

Fixing this is pretty straight forward. We just need to always skip capturing the root resource, even if the cache is disabled.

Fixes #388

### DIFF
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -52,7 +52,7 @@ export function createRequestFinishedHandler({
       let resource = getResource(url);
 
       // process and cache the response and resource
-      if (!resource || disableCache) {
+      if (!resource?.root && (!resource || disableCache)) {
         let response = request.response;
         let capture = response && hostnameMatches(allowedHostnames, url);
         let body = capture && await response.buffer();
--- a/packages/core/test/discovery.test.js
+++ b/packages/core/test/discovery.test.js
@@ -290,6 +290,41 @@ describe('Discovery', () => {
     );
   });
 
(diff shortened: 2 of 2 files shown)


Robdel12 (approved):
🏁 Solid find after some dog pile debugging 😆

Issue #392, opened 2021-06-24
`execute()` doesn't work in `additionalSnapshots`


chris-dura:
```
"@percy/cli": "^1.0.0-beta.55",
```

The `execute()` function seems to work when in the "parent" snapshot:
```js
// snapshots.js

module.exports = [
	{
		name: 'snapshot one',
		url: 'http://localhost:8080',
		waitForSelector: '.some-element',
		execute() {
			document.querySelector('h1').innerHTML = 'Tilting on windmills...';
		}
	}
]
```
However, trying to run the same `execute()` in an `additionalSnapshot` (like in [this pages.js example](https://github.com/percy/cli/tree/master/packages/cli-snapshot#page-options)) it does _not_ work, and the `<h1>` content stays the same :/

```js
// snapshots.js

module.exports = [
	{
		name: 'snapshot one',
		url: 'http://localhost:8080',
		waitForSelector: '.some-element',
		additionalSnapshots: [
			{
				suffix: '--tilt',
				execute() {
					document.querySelector('h1').innerHTML = 'Tilting on windmills...';
				}
			}
		]
	}
]
```

wwilsman:
Thanks for raising this @chris-dura!

Looks like we need to remove the `execute` from this line: https://github.com/percy/cli/blob/3cc36e7f47f8c37039345a45879594ac83713748/packages/core/src/percy.js#L342

PR out shortly 👍

wwilsman:
End of the work day for us, so we'll merge and release that fix tomorrow! :+1:

wwilsman:
Fix shipped in the latest release!

PR #393, opened 2021-06-24
🐛 Fix additional snapshot options

wwilsman:
## What is this?

@chris-dura helped point out in #392 that additional snapshot options were being passed incorrectly to the `page.snapshot()` method.

The additional options (`opts`) should override parent options (`conf`). We also don't need to explicitly provide `execute` since it is included in `opts` and wasn't being extracted from the additional snapshot anyway.


Robdel12 (approved):
🏁 Ahhh, good stuff.

PR #398, opened 2021-06-28
🥅 Check version of Node, exit if not supported

Robdel12:
## What is this?

Even though NPM/yarn checks on install, due to the number of warnings/errors in a given JS project, it's never noticed. We're going to hard error and exit on any unsupported versions of Node now to prevent support requests based on this.

### DIFF
--- a/packages/cli/index.js
+++ b/packages/cli/index.js
@@ -1,5 +1,5 @@
-/* istanbul ignore if */
-if (parseInt(process.version.split('.')[0].substring(1), 10) <= 11) {
+/* istanbul ignore if: hard to test without actually using an unsupported version */
+if (parseInt(process.version.split('.')[0].substring(1), 10) < 12) {
   console.error(`Node ${process.version} is not supported. Percy only supports the current LTS of Node. Please upgrade to Node v12+`);
   process.exit(1);
 }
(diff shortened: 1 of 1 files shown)


packages/cli/index.js
```diff
@@ -1,3 +1,9 @@
+/* istanbul ignore if */
```
wwilsman, line 1:
```suggestion
/* istanbul ignore if: hard to test without actually using an unsupported version */
```

packages/cli/index.js
```diff
@@ -1,3 +1,9 @@
...
+if (parseInt(process.version.split('.')[0].substring(1), 10) <= 11) {
```
wwilsman, line 2: Should we add a `semver` dep and import `engines.node` from the package.json?

```suggestion
if (!semver.satisfies(process.version, pkg.engines.node)) {
```
<br>

If not, we should at least assert against `12` so it's easy to search for and update

```suggestion
if (parseInt(process.version.split('.')[0].substring(1), 10) < 12) {
```

wwilsman (changes requested):
Small nits

wwilsman (approved):
🎮

PR #400, opened 2021-06-29
✨ Do not wait for remote requests to idle during asset discovery

### LINKED ISSUE #399: Do not wait for network idle when in-flight requests will not be captured
Previously, asset discovery would outright abort remote resource requests that would never be captured. This behavior was removed in #369 so we can capture the DOM of page while simultaneously discovering resources. We don't want to abort requests in this scenario so the page can be accurately rendered before serializing the DOM.

Another reason not to outright abort remote resources, would be in cases of "local" resources originating from remote resources. For example, say a page has a stylesheet at a cross-origin domain, however that stylesheet includes images or fonts that are present at the root domain. By aborting the cross-origin stylesheet that will not be captured, the same-origin images would never be requested. Think along the lines of a CDN that hosts custom stylesheets for a website.

These remote resources though can sometimes take a while to resolve. And when the resource will not be captured anyway, it blocks asset discovery from completing.

To speed up asset discovery when encountering resources that will not be captured, we should not track them at all. Similar to event stream requests, we can untrack resources that will never be captured. This is better than aborting them completely, and should have the same speed improvements.

wwilsman:
## What is this?

As described in #399, when requests will not be captured, they can clog up network idle and cause snapshots to take longer to process. However, if we were to untrack remote requests, as that issue suggests, pages might not yet be ready to snapshot when capturing the DOM ourselves since external resources are sometimes required by sites to load properly.

Instead, this PR adds a new feature to the network idle method to filter requests that are awaited on. This new filter can be used when not capturing our own DOM snapshot to only wait on requests from allowed hostnames.

I added a test for capturing lazy images as a precursor to a new test for not awaiting on remote resources (this test will timeout when failing). Lazy images are used in these tests because non-lazy images are normally awaited on via the document's `load` event, while lazy images are requested after the document's `load` event.

The test diff looks a little weird because I also removed an older test that was included here for coverage of the `hostnameMatches` function. That function now lives in `@percy/client` with its own tests, so the test here was unnecessary.

I'm labelling this as an enhancement as oppose to a bugfix since it's not necessarily a bug that we're waiting on all requests from the network, and this fixes #399 by further enhancing how we handle resources.

### DIFF
--- a/packages/core/src/network.js
+++ b/packages/core/src/network.js
@@ -46,7 +46,7 @@ export default class Network {
   }
 
   // Resolves after the timeout when there are no more in-flight requests.
-  async idle(timeout = this.timeout || 100) {
+  async idle(filter = r => r, timeout = this.timeout || 100) {
     this.log.debug(`Wait for ${timeout}ms idle`, this.page.meta);
 
     await waitFor(() => {
@@ -54,7 +54,8 @@ export default class Network {
         throw new Error(`Network error: ${this.page.closedReason}`);
       }
 
-      return this.#requests.size === 0;
+      return Array.from(this.#requests.values())
+        .filter(filter).length === 0;
     }, {
       timeout: 30 * 1000, // 30 second error timeout
       idle: timeout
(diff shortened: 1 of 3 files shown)


Robdel12 (approved):
🏁

PR #405, opened 2021-07-06
🐛 Serialize iframe document URLs

### LINKED ISSUE #396: Fix serialized iframe relative URLs
When a same-origin frame is serialized, all of its content becomes relative to the frame parent.

In the example below,  `localhost:8080/bar/styles.css` would be proper location of the stylesheet.

```html
<!-- localhost:8080/foo -->
<iframe src="/bar"></iframe>
```

```html
<!-- localhost:8080/bar -->
<link href="styles.css" rel="stylesheet"/>
```

After serializing, the resulting DOM would look like:

```html
<!-- localhost:8080/foo -->
<iframe srcdoc="<link href=\"styles.css\" rel=\"stylesheet\"/>"></iframe>
```

And with that DOM, the stylesheet request will be interpreted as `localhost:8080/foo/style.css`. Which is incorrect. This issue might also be present if the embedded iframe is not same-origin but allows cross-origin resource sharing instead.

We can likely fix this, albeit a little heavy handed, by prepending all nested relative URLs with the iframe's src.

<table>
<thead><tr><td>Frame Source</td><td>Asset URL</td><td align="right">Serialized URL</td></tr></thead>
<tr><td><code>/bar</code></td><td><code>styles.css</code></td><td align="right"><code>/bar/styles.css</code></td>
<tr><td><code>//example.com/bar</code></td><td><code>styles.css</code></td><td align="right"><code>//example.com/bar/styles.css</code></td>
<tr><td><code>//example.com/bar</code></td><td><code>/styles.css</code></td><td align="right"><code>//example.com/styles.css</code></td>
</table>

Open to other solutions as well

wwilsman:
## What is this?

As per #396, when a same-origin frame is serialized, all of its content becomes relative to the frame parent, which can then result in invalid resource URLs. Like the issue suggests, we can transform the URLs within the frame document to be fully qualified URLs based on the frame document's URL.

During the recursive DOM serialization, we can provide a custom transform function to be applied to iframe documents after cloning but before the function returns a string. In this transformation we can then find all relevant elements to transform their respective URLs.

According to HTML 4/5 spec the `href`, `src`, `srcset`, `poster`, and `background` attributes are attributes that may contain URIs (and that we care about transforming). Most attributes' property counterparts are already resolved, however that is not always the case (such as `background`). The `srcset` property is also not resolved, and requires parsing to properly replace URLs within the list of image candidates. 

In addition to those attributes, CSS within `style` elements or inline `style` attributes may also contain URLs using the `url()` CSS function and therefore are also transformed.

In adding a test, Karma complained about assets being missing. I found a workaround by setting a `proxy` option to a semi-valid URL so it doesn't proxy and doesn't log. If logs crop up in the future, we probably want to host a single fake asset to silence them.

Fixes #396

### DIFF
--- a/packages/dom/src/serialize-frames.js
+++ b/packages/dom/src/serialize-frames.js
@@ -2,26 +2,46 @@ import serializeDOM from './serialize-dom';
 
 // List of attributes that accept URIs that should be transformed
 const URI_ATTRS = ['href', 'src', 'srcset', 'poster', 'background'];
-const URI_SELECTOR = URI_ATTRS.map(a => `[${a}]`).join(',');
+const URI_SELECTOR = URI_ATTRS.map(a => `[${a}]`).join(',') + (
+  ',[style*="url("]'); // include elements with url style attributes
 
 // A loose srcset image candidate regex split into capture groups
 // https://html.spec.whatwg.org/multipage/images.html#srcset-attribute
 const SRCSET_REGEX = /(\s*)([^,]\S*[^,])((?:\s+[^,]+)*\s*(?:,|$))/g;
 
+// A loose CSS url() regex split into capture groups
+const CSS_URL_REGEX = /(url\((["']?))((?:\\.|(?!\2).|[^)])+)(\2\))/g;
+
 // Transforms URL attributes within a document to be fully qualified URLs. This is necessary when
 // embedded documents are serialized and their contents become root-relative.
 function transformRelativeUrls(dom) {
+  // transform style elements that might contain URLs
+  for (let style of dom.querySelectorAll('style')) {
+    style.innerHTML &&= style.innerHTML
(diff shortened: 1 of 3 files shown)


packages/dom/src/serialize-frames.js
```diff
@@ -22,6 +72,7 @@ export default function serializeFrames(dom, clone, { enableJavaScript }) {
 
       // recersively serialize contents
       let serialized = serializeDOM({
+        domTransformation: transformRelativeUrls,
```
Robdel12, line 75: [Nice](https://i.kym-cdn.com/photos/images/original/001/246/322/9b2.gif)

Robdel12 (approved):
🏁

PR #406, opened 2021-07-09
♻️ Refactor percy/config utils

wwilsman:
## Purpose

While adding better validations for config options and per-snapshot validations, data must be validated against schemas other than the global config schema. To facilitate this, the validate util was refactored slightly. The merge util is also used within the core snapshot options config function to merge and validate provided options with config options and defaults while also handling deprecated options. Due to the merge util's limitation of only merging two objects at a time while mutating the first object, it had to be used several times in a row in order to merge all of the snapshot options together into a new object. Furthermore, it's options argument was limited in it's abilities and the merge function itself was centered around config normalization, so wasn't generic enough to use outside of said normalization function.

## Approach

The `validate` utils were refactored so default schema `$id`s now start with a forward slash (`/`) to help distinguish them from local schema references. The function will also now only return an array of errors on failure and `undefined` on success. The associated `addSchema` util was updated to accept schemas without automatically adding them to the global Percy config schema (via an explicit `$id`). This allows the validate util to be used with other registered schemas and not just the global Percy config schema.

The `merge` util was refactored to accept an array of objects instead of only two objects. It also always returns a deep clone rather than mutating the first argument. The limited `options` argument was replaced with a `map` callback argument. The callback will be called for each property within each object, recursively, with the path, previous value, and next value of the property. A path-value pair can be returned to override the default merge behavior. Some options previously handled by the old merge function were replaced by the new map callback in the appropriate places.

Other features were also added while refactoring that were found to be useful for better validations: 
- Schemas using `minimum` and `maximum` keywords will be clamped rather than scrubbed. 
- Schemas can use a custom `instanceof` keyword to allow functions or regular expressions. 
- Schemas can also use a custom `disallowed` keyword, based on the built-in `required` keyword (but negated).
- Nested objects that have missing required properties are now scrubbed. 
- Empty properties such as array items that are missing or deleted are now scrubbed.
   e.g. `['foo', <5 empty items>, 'xyzzy'] => ['foo', 'xyzzy']`
- Validation error paths were updated to show as proper array notation.
   before: `snapshots.widths.0: must be <= 2000`
   after: `snapshots.widths[0]: must be <= 2000`

### DIFF
--- a/packages/config/src/defaults.js
+++ b/packages/config/src/defaults.js
@@ -1,21 +1,22 @@
-import { merge } from './normalize';
+import { merge } from './utils';
 import { getSchema } from './validate';
 
+const { isArray } = Array;
 const { assign, entries } = Object;
 
 // Recursively walks a schema and collects defaults. When no schema is provided,
 // the default config schema is used.
-function getDefaultFromSchema(schema) {
+function getDefaultsFromSchema(schema) {
   if (!schema || typeof schema.$ref === 'string') {
     // get the schema from ajv
-    return getDefaultFromSchema(getSchema(schema?.$ref ?? 'config'));
+    return getDefaultsFromSchema(getSchema(schema?.$ref ?? '/config'));
   } else if (schema.default != null) {
     // return the default for this schema
     return schema.default;
   } else if (schema.type === 'object' && schema.properties) {
     // return an object of default properties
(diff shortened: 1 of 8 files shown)


Robdel12 (approved):
🏁

PR #413, opened 2021-07-14
🐛 Fix default proxy port

### LINKED ISSUE #402: Proxy: Port 80 is stripped from URL
When setting `HTTPS_PROXY` to port 80 (`HTTPS_PROXY="http://localhost:80"`), we accidentally strip the port from the URL, which throws an error: 

```
npx percy exec -- node snapshots.js
[percy] Skipping visual tests - Port should be >= 0 and < 65536. Received .
```

wwilsman:
## What is this?

With `new URL()`, the default port is never included in the resulting object, even when explicitly defined.

```js
new URL('http://localhost:80').port
// => ""

new URL('https://localhost:443').port
// => ""
```

In the `@percy/client` proxy, the proxy address' port is passed along to the `(net|tls).connect` method. With the above in mind, this results in an empty string being passed along. Since the `port` option is a required argument of the method, this causes connections to immediately close, showing only a vague error.

This PR adds a small `port` util which will return a default port if one isn't specified, port 443 for HTTPS requests or port 80 otherwise. This util is then used in the two places where a URL port is referenced.

Fixes #402

### DIFF
--- a/packages/client/src/request.js
+++ b/packages/client/src/request.js
@@ -13,6 +13,11 @@ const RETRY_ERROR_CODES = [
   'EHOSTUNREACH', 'EAI_AGAIN'
 ];
 
+// Returns the port number of a URL object. Defaults to port 443 for https
+// protocols or port 80 otherwise.
+const port = ({ port, protocol }) => port ||
+  (protocol === 'https:' && 443) || 80;
+
 // Proxified https agent
 export class ProxyHttpsAgent extends https.Agent {
   // enforce request options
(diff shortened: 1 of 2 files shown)


wwilsman:
Locally, the tests pass since I can freely bind to ports 80 and 443. In CI however, those ports are restricted. I might be able to use `sudo` for linux, but I'm still looking into a workaround for windows (or a GH actions config for either).

wwilsman:
Rather than elevate privileges in CI, which felt wrong and proved to be rather difficult without moving to self-hosted containers, I opted to adjust the tests to allow server creation and request failures while still testing for a default port using a jasmine spy. With the adjustments, coverage dropped and so a rewritten test was added back to take care of it.

Robdel12 (approved):
🏁

Issue #305, opened 2021-04-13
Validate snapshot page listings


wwilsman:
Percy config files get automatically validated so helpful warnings can be shown about possibly non-existent options and also so unexpected options do not make their way into the config objects. We should do the same for page listings when using `percy snapshot`.

PR #416, opened 2021-07-15
✨ Add per-snapshot config validation!

### LINKED ISSUE #389: Validate specific config options
As long as the option name and type is correct it is thought to be valid. We should additionally validate the values themselves to help avoid confusion and footguns.

Examples:

- maximum network idle timeout
- valid allowed hostnames
- widths/height outside the bounds of the API min/max
- etc.

wwilsman:
## What is this?

While working on #305, the decision was made to have the config schema for snapshot options included in `@percy/core` since that is where the options are consumed. Initially however, core only registered the schema while the snapshot command used it to validate snapshot listing options.

With #406 refactoring the `merge` and `validate` utils, the `getSnapshotConfig` helper can now be refactored to utilize their new features to better merge and now validate options! This brings config validation to all of the SDKs as warnings logged whenever `percySnapshot` is called with unrecognized or invalid options.

Speaking of invalid options, this PR also adds more restrictions to existing config options:
- The `widths` and `minHeight` options have min and max values. These values are derived from the upload command clamping, which itself was derived from our API restrictions.
- Hostnames in `allowedHostnames` must be actual hostnames, meaning no protocols and no paths. Including protocols and paths never worked and this validation will help clear up that confusion.
- The `networkIdleTimeout` should never really be larger than 500ms, but is restricted to 750ms to give it some breathing room. This restriction is imposed because long timeouts cause slow tests and don't actually address the underlying network issues.
- Discovery `concurrency` has a minimum value of 1 because having a negative or zero value causes the queue to lock up.

For new per-snapshot validations other restrictions are also imposed:
- Disallowed capture options, like `waitForSelector`, when performing discovery for an existing `domSnapshot`.
- The `waitForTimeout` option must be between `1ms` and `30000ms`.
- The `execute` option must be a string or function.
- For `additionalSnapshots`, a `name`, `prefix`, or `suffix` is required.
- For `additionalSnapshots`, a `prefix` or `suffix` cannot be provided with a `name`.

Tests were updated accordingly and a couple things that were previously taken care of in the config helper are now taken care of where necessary, such as defaulting an additional snapshot name using a prefix/suffix.

This will close #389

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -161,8 +161,8 @@ percy.snapshot({
 
 #### Options
 
-- `name` — Snapshot name (**required**)
 - `url` — Snapshot URL (**required**)
+- `name` — Snapshot name
 - `domSnapshot` — Snapshot DOM string
 - `clientInfo` — Additional client info
 - `environmentInfo` — Additional environment info
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -1,8 +1,8 @@
 import { strict as assert } from 'assert';
+import PercyConfig from '@percy/config';
 import { merge } from '@percy/config/dist/utils';
-import { hostname } from './utils';
 
-// Common options used in Percy commands
+// Common config options used in Percy commands
 export const schema = {
   snapshot: {
     type: 'object',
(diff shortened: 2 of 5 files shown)


Robdel12 (approved):
🏁 I am in love. The network idle timeout & allowed hostnames validations are going to be so good going forward

PR #417, opened 2021-07-15
✨ Validate snapshot command page listings

### LINKED ISSUE #305: Validate snapshot page listings

wwilsman:
## What is this?

This resolves #305 and piggybacks off of #406 & #416 to validate page listings passed to the snapshot command. In addition to validations, page listing files can now contain a top-level `snapshots` key which references the list of snapshots. This enables YAML files to utilize [anchors, references, and extending](https://blog.daemonl.com/2016/02/yaml.html).

Tests were also reorganized a little and cleanup is done with the help of a rimraf transient dep.

### DIFF
--- a/packages/cli-snapshot/src/commands/snapshot.js
+++ b/packages/cli-snapshot/src/commands/snapshot.js
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import path from 'path';
+import PercyConfig from '@percy/config';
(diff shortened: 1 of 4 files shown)


Robdel12 (approved):
🏁

PR #418, opened 2021-07-15
✨ Add static snapshot overrides option

wwilsman:
## What is this?

This allows static snapshots to be configured with per-snapshot options!

``` yaml
static:
  files: **/*.{html,htm}
  overrides:
  - files: /foo-bar.html
    waitForSelector: .is-ready
    execute: |
      document.querySelector('.button').click()
```

I'm not too thrilled with the `overrides` name; could be `options`, `snapshotOptions`, `etc`?

### DIFF
--- a/packages/cli-snapshot/README.md
+++ b/packages/cli-snapshot/README.md
@@ -47,20 +47,6 @@ EXAMPLES
 
 ## Usage
 
-### Static Directory
-
-When providing a static directory, it will be served locally and pages matching the `files` argument
-(and excluding the `ignore` argument) will be navigated to and snapshotted.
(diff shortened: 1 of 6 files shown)


Robdel12 (approved):
🏁 

We have follow ups to update the readme docs right?

wwilsman:
If we're okay with the name or one of the alternatives, we could also allow some options under the `static` keyword itself such as `waitForTimeout`, `execute`, etc

wwilsman:
Upon slack discussion, we're gonna leave this PR as is. The name is okay and we can always circle back for more options.

PR #419, opened 2021-07-15
✨ Add queue concurrency to the upload command

### LINKED ISSUE #391: Parallelize the upload command
Currently, the `percy upload` command uploads a single snapshot at a time. This command uses `@percy/client` directly, and so does not have access to the concurrency used in `@percy/core`.

We can still utilize the core queue class to create an image upload queue and make this process a bit faster for this command.

wwilsman:
## What is this?

This resolves #391 and utilizes core's queue class to parallelize the upload command. It adds a `upload.concurrency` option as well to allow control of the queue size. It also borrows a pattern from core by waiting on uploads at the "end" of the process and clearing the queue on error to allow the process to continue.

### DIFF
--- a/packages/cli-upload/src/commands/upload.js
+++ b/packages/cli-upload/src/commands/upload.js
@@ -5,11 +5,12 @@ import logger from '@percy/logger';
 import globby from 'globby';
 import imageSize from 'image-size';
 import PercyClient from '@percy/client';
+import Queue from '@percy/core/dist/queue';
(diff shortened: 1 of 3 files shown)


Robdel12 (approved):
🏁 I'll never stop loving all of the wins we get from consolidating the SDKs onto this toolchain. 

(nice to see you get bit by the linter too sometimes HAHAHA)

PR #422, opened 2021-07-15
⬆️ Upgrade dependencies

wwilsman:
## What is this?

We have a few open PRs for dependency upgrades that are failing due to strange reasons. I have a hunch that some of the failures are related to upgrading only a single dependency without the others (such as babel dependencies that might conflict with each other).

This PR is the result of running `yarn upgrade-interactive --latest` and fixing any local errors that came up. The only errors that actually did come up were all related to rollup changes which were fixed by adjusting some config options slightly.


Robdel12 (approved):
🏁 yesss, close out those other PRs

Issue #410, opened 2021-07-14
Feature request: Variable expansion in .env


prsdta:
Hello, I see Percy is using `dotenv` to read `.env` variables, in the rails way which is very useful.
However, the ruby `dotenv` implementation also does variable interpolation.
Example:
```
USER=postgres
DATABASE_TEST_URL="postgres://$USER:@localhost/db"
```

Could you add it, e.g. with https://github.com/motdotla/dotenv-expand ?

Thank you.

Robdel12:
Closed with #426

prsdta:
Thank you!

PR #426, opened 2021-07-16
✨ Replace dotenv with custom implementation

wwilsman:
## What is this?

As requested by #410, it's nice to be able to interpolate environment variables within other environment variables. As suggested in the issue, I started with the `dotenv-expand` package and added it as a dependency. Upon a quick review of its internals, I noticed that not only was it small and still did more than we needed, but it was also a bit inefficient with several loops and unnecessary regular expression creations within those loops.

After implementing a more succinct variable expansion helper, I was checking against dotenv-rails, which is what we always intended to mimic. I then noticed there were several other differences with the dotenv package we were using related to variable interpolation, escaping, file parsing, and other features.

The dotenv package we were using was pretty small, so it was fairly straightforward to copy it over and trim away everything we didn't need. For example, we always want to set `process.env` when it isn't already, while dotenv libraries are split up to allow parsing without setting `process.env`. After cutting away the fat, changes were made to align with dotenv-rails, including variable substitution (the original goal of this PR).

Most changes were ported over from dotenv-rails directly, such as the regular expression used to parse the file and the logic behind expanding, unescaping, and interpolating values.

### DIFF
--- a/packages/env/package.json
+++ b/packages/env/package.json
@@ -20,8 +20,5 @@
   },
   "devDependencies": {
     "mock-require": "^3.0.3"
-  },
-  "dependencies": {
-    "dotenv": "^10.0.0"
   }
 }
--- a/packages/env/src/dotenv.js
+++ b/packages/env/src/dotenv.js
@@ -1,25 +1,76 @@
-const dotenv = require('dotenv');
+import fs from 'fs';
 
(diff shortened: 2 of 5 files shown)


Robdel12 (approved):
🏁 Verry nice

PR #430, opened 2021-07-20
♻️ Lockstep sub-package versions

Robdel12:
## What is this?

We should be lock stepping all our sub-packages since they're all released together. This also makes it possible to downgrade the CLI to a specific version


wwilsman (approved):
Excellent point about downgrading 

🎮

PR #431, opened 2021-07-20
🐛 Ensure core schemas are loaded before dependent snapshot schemas

### LINKED ISSUE #429: Error Plugin: @percy/cli: can't resolve reference /snapshot#/properties/name from id /config/static
**Possible bug introduced in beta 59?**

Renovate bot is trying to update to `@percy/cli to v1.0.0-beta.59 ` and our action that runs E2E tests and take screenshots is now falling with the following error:

```
(node:1969) Error Plugin: @percy/cli: can't resolve reference /snapshot#/properties/name from id /config/static
module: @oclif/config@1.17.0
task: runHook init
plugin: @percy/cli
root: /home/runner/work/montevideo/montevideo/node_modules/@percy/cli
```

The same action was running properly on beta.58

Our script just runs: `percy exec -- cypress run` (and some cypress env variables are used too with the --env flag)

wwilsman:
## What is this?

In tests, schemas are loaded where needed. In practice however, core schemas are not loaded in time for references in the snapshot schema. The answer is to register core schemas before they are needed, however by default this causes AJV to error about existing schema `$id`s when core get's around to registering schemas for itself.

The `addSchema` util was updated to replace existing `$id` schemas just as it already does for config file schemas. A test was added for this, but because the array handling was moved, coverage became low. Another test for an array of schemas was added to account for this missing coverage.

Since arrays of schemas are handled by our util now rather than forwarded directly onto AJV, a new pattern was adopted in the appropriate places. Where multiple schemas are defined, the file exports a single `schemas` array that can be passed to the `addSchema` call made when registering schemas. This reduces some boilerplate around adding schemas and also makes it simple to add all possible required core schemas in one go before they are needed, which was the original intent of this PR.

This will fix #429

### DIFF
--- a/packages/cli-command/test/command.test.js
+++ b/packages/cli-command/test/command.test.js
@@ -1,9 +1,10 @@
 import logger from '@percy/logger/test/helpers';
 import PercyConfig from '@percy/config';
+import { configSchema } from '@percy/core/dist/config';
 import PercyCommand, { flags } from '../src';
 
 // add config schema to test discovery flags
-PercyConfig.addSchema(require('@percy/core/dist/config').schema);
+PercyConfig.addSchema(configSchema);
 
 describe('PercyCommand', () => {
   let results;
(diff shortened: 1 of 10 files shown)


Robdel12 (approved):
🏁

PR #432, opened 2021-07-20
📝  Add bug issue template

Robdel12:
## What is this?

This adds an issue template to the repo to help gather the right info from bug reports

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,47 @@
+---
+name: Bug report
+about: Create a report to help us fix the issue
+title: ''
+labels: ''
+assignees: ''
+---
+<!--
+## Reach out to Percy support instead?
+
+If you’re having issues that _aren’t SDK bugs_, it would be best for you to
+reach out to support instead: support@percy.io or
+https://www.browserstack.com/contact#technical-support
+-->
+
+## The problem
+
+Briefly describe the issue you are experiencing (or the feature you want to see
+added to Percy). Tell us what you were trying to do and what happened
+instead. Remember, this is _not_ a place to ask questions. For that, go to
+https://github.com/percy/cli/discussions/new
+
+## Environment
+
+- Node version:
+- `@percy/cli` version:
+- Version of Percy SDK you’re using:
+- If needed, a build or snapshot ID:
+- OS version:
+- Type of shell command-line [interface]:
+
+## Details
+
+If necessary, describe the problem you have been experiencing in more detail.
+
+## Debug logs
+
+If you are reporting a bug, _always_ include logs! Run a Percy build with
+`--verbose` to get the full debug logs. You can also set an environment var
+`PERCY_LOGLEVEL=debug` to get debug logs. For example, `percy exec --verbose --
+[test command]`.  Please include the full complete test output.
+
+## Code to reproduce issue
+
+Given the nature of testing/environment bugs, it’s best to try and isolate the
+issue in a reproducible repo. This will make it much easier for us to diagnose
+and fix.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,35 @@
+---
+name: Bug report
+about: Create a report to help us fix the issue
+title: ''
+labels: ''
+assignees: ''
+
```
wwilsman, line 7: Extra newline?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,35 @@
...
+---
+
+## The problem
+
+Briefly describe the issue you are experiencing (or the feature you want to see added to Percy). Tell us what you were trying to do and what happened instead. Remember, this is _not_ a place to ask questions. For that, go to https://github.com/percy/cli/discussions/new
+
+## Reach out to Percy support
```
wwilsman, line 14: I think we should remove this headline so people don't accidentally leave it in assuming that it's part of the desired issue content. I can easily see this being filled in as if it were a question being asked, such as:

```markdown
## Reach out to Percy support

already done
```

We should also move the below redirection to be first and foremost directly after the first heading. So it's the very first thing people read and don't accidentally skim over it.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,35 @@
...
+
+If you’re having issues that _aren’t SDK bugs_, it would be best for you to reach out to support instead: support@percy.io or https://www.browserstack.com/contact#technical-support
+
+## Environment
+
+- Node version
+- What version of  `@percy/cli`
+- What version of Percy SDK you’re using
+- If needed, a build or snapshot ID
```
wwilsman, line 23: What do you think of adding a colon (`:`) to these so it's more of a "fill in the blank" section?

Also, should these be posed as questions, or as assertions (it's a mix of both right now)?

```markdown
- What version of Node?:
- What version of `@percy/cli`?:
- ...
```

vs

```markdown
- Node version:
- `@percy/cli` version:
- ...
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Details
+
+If necessary, describe the problem you have been experiencing in more detail.
+
+## Debug logs
+
+If you are reporting a bug, _always_ include logs! Run a Percy build with `--verbose` to get the full debug logs. For example, `percy exec --verbose -- [test command]`.  Please include the full complete test output. 
+
+## Code to reproduce Issue
```
wwilsman, line 33: Inconsistent title casing. Based on the other titles, you probably want to lowercase `issue`

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,35 @@
...
+- What version of Percy SDK you’re using
+- If needed, a build or snapshot ID
+
+## Details
+
+If necessary, describe the problem you have been experiencing in more detail.
+
+## Debug logs
+
+If you are reporting a bug, _always_ include logs! Run a Percy build with `--verbose` to get the full debug logs. For example, `percy exec --verbose -- [test command]`.  Please include the full complete test output.
```
wwilsman, line 31: Should we mention `PERCY_LOGLEVEL=debug`? In some CI envs, it might be easier to add an env var from the UI rather than alter the command used to run tests.

wwilsman (changes requested):

.github/ISSUE_TEMPLATE/bug_report.md
Robdel12: GitHub generates that 🤨

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,47 @@
...
+https://github.com/percy/cli/discussions/new
+
+## Environment
+
+- Node version:
+- `@percy/cli` version:
+- Version of Percy SDK you’re using:
+- If needed, a build or snapshot ID:
+- OS version:
+- Type of command line / shell:
```
wwilsman, line 30: `command line / shell / terminal / how you run commands?` 😆 

In all seriousness, I think this is technically correct: `Type of shell command-line [interface]` I put "interface" in optional brackets because I pulled that full term from a wiki but I think `shell command-line` still holds on its own.

wwilsman (approved):
🎮

PR #434, opened 2021-07-20
✨ Add upload option to strip file extensions

### LINKED ISSUE #412: Add an option to remove the file extension from image uploads
When using the `percy upload` command, the file extension is kept in snapshot names. This is because you can have multiple files with the same name but different extensions, i.e. `image.png`, `image.jpg`, `image.jpeg`. If file extensions were automatically removed, only the first upload will work and the following uploads will return a duplicate snapshot API error.

However, having matching filenames with differing extensions probably isn't very common. More likely what is common, is the filenames without extensions are _intended_ to be snapshot names. With intentional names, the duplicate snapshot API error is doing its job.

Rather than change the behavior and suddenly rename all existing snapshots for current users (resulting in "new" snapshots and broken comparisons), we can add an opt-in flag and/or config file option to strip extensions from the resulting snapshot name. This could be a boolean option, or maybe a more complicated config option for more custom behaviors.

wwilsman:
## What is this?

This implements and resolves #412 with a new `--strip-extensions` flag and accompanying config file option.

The upload config options spreading was growing, so was adjusted to reference the config object rather than spread each individual config option.

The extensions are removed from file paths by parsing the path and rejoining without the extension. While doing this, I realized that the `name` parameter does not include parent directories and updated the resource helper accordingly.

The allowed file type regular expression was also updated since it now tests the file extension only and not the entire path.

### DIFF
--- a/packages/cli-upload/src/commands/upload.js
+++ b/packages/cli-upload/src/commands/upload.js
@@ -10,7 +10,7 @@ import createImageResources from '../resources';
 import { schema } from '../config';
 import pkg from '../../package.json';
 
-const ALLOWED_FILE_TYPES = /\.(png|jpg|jpeg)$/i;
+const ALLOWED_FILE_TYPES = /^\.(png|jpg|jpeg)$/i;
 
 export class Upload extends Command {
   static description = 'Upload a directory of images to Percy';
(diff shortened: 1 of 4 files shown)


Robdel12 (approved):
🏁

packages/cli-upload/src/commands/upload.js
```diff
@@ -10,7 +10,7 @@ import createImageResources from '../resources';
 import { schema } from '../config';
 import pkg from '../../package.json';
 
-const ALLOWED_FILE_TYPES = /\.(png|jpg|jpeg)$/i;
+const ALLOWED_FILE_TYPES = /^\.(png|jpg|jpeg)$/i;
```
andressalinero, line 13: @wwilsman Should this also include uppercase for windows users where windows defaults the file type extension to upper case.

wwilsman: @andressalinero no need, as the `i` at the end of the regular expression means it's case insensitive.

andressalinero: That's very elegant!

PR #447, opened 2021-07-27
✨ Add static snapshot clean-urls and rewrites options

### LINKED ISSUE #420: Allow static snapshot pretty paths
For pages such as `blog/index.html`, the desired URL path is `/blog`. Sometimes entire sites are built this way and things on the page might redirect to the pretty URL. We should add an option to automatically support pretty paths for static snapshots. This should be fairly easy as the serve handler we use accepts a few [different options](https://github.com/vercel/serve-handler).

wwilsman:
## What is this?

This adds two new options for static snapshots and refactors small parts of the snapshot command.

Logic that differed between static and non-static usages were moved into their respective implementation specific methods. This includes `base-url` error handling and default/normalized page options. Although the latter is shared, how the server address and non-static `base-url` is determined is specific to each implementation. Tests were also refactored a bit so suite hooks run for each test, and a couple other small test changes were made as well.

The `clean-urls` option is passed along to the static server handler directly, while the `rewrites` option is concatenated to any `base-url` rewrite. The `rewrites` are then also used in conjunction with `overrides` to generate a default page URL from the static file path. Another rewrite override is used to remove indexes and extensions from page URLs when `clean-urls` is true.

For config files, we don't want to normalize nested keys within `rewrites`, so it was added to the ignore-list in `@percy/config`'s normalize module (with tests).

[The snapshot command readme was also updated](https://github.com/percy/cli/blob/ww/static-rewrites/packages/cli-snapshot/README.md#static-options).

Resolves #420

### DIFF
--- a/packages/cli-snapshot/README.md
+++ b/packages/cli-snapshot/README.md
@@ -18,24 +18,19 @@ ARGUMENTS
   PATHNAME  path to a directory or file containing a list of pages
 
 OPTIONS
-  -b, --base-url=base-url                          [default: /] the url path to serve the static directory from
+  -b, --base-url=base-url                          the base url pages are hosted at when snapshotting
   -c, --config=config                              configuration file path
   -d, --dry-run                                    prints a list of pages to snapshot without snapshotting
+  -h, --allowed-hostname=allowed-hostname          allowed hostnames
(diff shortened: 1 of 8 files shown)


Robdel12 (approved):
🏁

PR #452, opened 2021-07-29
🧹 Update release drafter config

Robdel12:
## What is this? 

Updates the release drafter config to match our new standardized labels


wwilsman (approved):
🎮

PR #453, opened 2021-07-29
🐛 Fix enableJavaScript default for snapshot discovery

wwilsman:
## What is this?

During the recent discovery/snapshot refactors, the `Page` class was updated to default `enableJavaScript` to `true` for other packages to borrow a page without explicitly enabling it. However, since `enableJavaScript` did not have a default configuration value for dom snapshots in the `Percy` class, that resulted in all snapshots defaulting to have javascript enabled during asset discovery (this does not affect the API default).

I also noticed that for snapshot captures, JavaScript is always enabled, even if the config option is `false`.

I updated this line to always prefer the `enableJavaScript` option and default based on the presence of `domSnapshot`. When there is no dom provided, the default should be `true`, otherwise it should be `false`. I also added a test so this can be caught in the future if it regresses again.

### DIFF
--- a/packages/core/src/percy.js
+++ b/packages/core/src/percy.js
@@ -295,7 +295,7 @@ export default class Percy {
         // open a new browser page
         page = await this.browser.page({
           networkIdleTimeout: this.config.discovery.networkIdleTimeout,
-          enableJavaScript: domSnapshot ? conf.enableJavaScript : true,
+          enableJavaScript: conf.enableJavaScript ?? !domSnapshot,
           requestHeaders: discovery.requestHeaders,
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
[🏁 Whoops](https://media4.giphy.com/media/26FPImXfDlv4AFbBC/giphy-downsized.gif?cid=6104955e6d787ed505016e59cb05e30c80dbbcd934a07c4b&rid=giphy-downsized.gif&ct=g&cid=6104955e6d787ed505016e59cb05e30c80dbbcd934a07c4b&rid=giphy-downsized.gif&ct=g)

PR #454, opened 2021-07-29
📝 Update upload and snapshot readmes

wwilsman:
## What is this?

A couple options aren't technically documented in the readmes. We should also update the docs site as well.


Robdel12 (approved):
> We should also update the docs site as well.

Yes

🏁

PR #456, opened 2021-07-29
✨ Allow registering multiple config migrations at a time

wwilsman:
## What is this?

With #450, the migrate function was updated to be able to migrate specific schemas. We can now utilize that new feature to register a snapshot config migration for deprecated options. These deprecations previously lived within a merge callback.

The migrate function was updated again in this PR to accept multiple migration at once similar to how we can register multiple schemas at once. The forced version was also moved into a default migration; and to ensure it always runs last, migrations are now registered in reverse order, so they run last-in-first-out.


Robdel12 (approved):
🏁 oh this is a nice win!

PR #457, opened 2021-07-30
⬆️ Update dependencies on and allow github-actions management by dependabot

mmorel-35:
Hi! This pr allows dependabot to handle github actions dependencies management.
I also included the update of the current dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,3 +6,9 @@ updates:
     interval: weekly
   commit-message:
     prefix: ⬆️
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: weekly
+  commit-message:
+    prefix: ⬆️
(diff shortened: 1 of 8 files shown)


wwilsman (approved):
Thanks @mmorel-35!

Weird how dependabot pinned one of these `setup-node` actions to `v2.3.0` but the rest to just `v2`. Looks automated though, so nothing to do about it I suppose.

🎮 🎉

Robdel12:
Hm, I don't know if I want to do this -- unless we go through and update the rest of the SDK repos to match.

wwilsman:
@robdel I definitely think it would be good to have our actions autoupdate. Even for those repos. Could have security issues and such that are fixed

Robdel12:
👍🏼 Let's make sure this makes it to all the other repos so we're not triaging similar-but-different configs/setups in the future

wwilsman:
Created an issue to track automating things like this that should be copied to other repos. I can at least spike something similar to how @Robdel12 spiked mass label management.

mmorel-35:
> Weird how dependabot pinned one of these `setup-node` actions to `v2.3.0` but the rest to just `v2`.
That was me, I modified it and just forgot one.
Some people prefer to have just the major version. Usually, dependabot is using the full semantical version.

PR #465, opened 2021-07-30
✨ Refactor static snapshot logic into utils

wwilsman:
## What is this?

This not only cleans up the main snapshot command class and makes it easier to distinguish between different functionalities, but it also allows us to use these utils for other purposes, such as for static site generator SDKs (currently gatsby, could be more in the future).

For the most part, this PR just moves some stuff around. The `withDefaults` and `serve` methods are now utils, and the path-page mapping was moved into it's own util. Smaller changes were made to replace `this` references with config arguments.


Robdel12 (approved):
🏁

PR #469, opened 2021-08-02
✨ Improve network tracking resilience

wwilsman:
## What is this?

With network tracking, there are a number of events which need to be handled to manage requests properly. While some of these events are intended to happen in a specific order, and handlers are written in a way to handle an unexpected order of events, there aren't actually guards in place to prevent handling seemingly duplicate events.

The changes in this PR were found in Puppeteer after discovering a font-related issue. We have a history of font-related issues, so I thought to give these changes a try here. The specific issue that prompted these safeguards are a result of this Chromium bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1196004

Our network handling isn't an exact 1:1 of Puppeteer's, so I wasn't able to reproduce the issue with a test for the same. One likely reason is because we always have browser caches disabled because we have our own cache. Regardless, after breaking down the changes I decided they would be a good addition to have anyway.

- When requests are handled or forgotten, they should be removed from all trackers.
- Intercepted request mapping was updated to track the entire request event.
- During pending requests, already intercepted requests are always handled.
- During request interception, already pending requests are only handled when it is not actually a redirect.
- During request handling, redirects clean up other existing trackers besides auth.

### DIFF
--- a/packages/core/src/network.js
+++ b/packages/core/src/network.js
@@ -62,6 +62,17 @@ export default class Network {
     });
   }
 
+  // Called when a request should be removed from various trackers
+  _forgetRequest({ requestId, interceptId }, keepPending) {
+    this.#requests.delete(requestId);
+    this.#authentications.delete(interceptId);
+
+    if (!keepPending) {
+      this.#pending.delete(requestId);
+      this.#intercepts.delete(requestId);
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
🏁  This was a reallly good find!

PR #479, opened 2021-08-05
🔊 Add log to warn when SDK user agent is missing

### LINKED ISSUE #455: Log warning if no env or client info are passed
We use those two strings for debugging SDKs. I think we should log a warning if at any point that info is missing from an implemented SDK. Without the user agent info, it's pretty hard to know which SDK / test framework the snapshots are coming from.

Robdel12:
## What is this?

This implements a warning log if the SDK sends a snapshot without any user agent info (`environmentInfo` & `clientInfo`). Closes #455 

Before releasing this, we will need to update the Storybook SDK

### DIFF
--- a/packages/cli-build/src/commands/build/finalize.js
+++ b/packages/cli-build/src/commands/build/finalize.js
@@ -1,6 +1,7 @@
 import Command, { flags } from '@percy/cli-command';
 import PercyClient from '@percy/client';
 import logger from '@percy/logger';
+import pkg from '../../../package.json';
 
 export class Finalize extends Command {
   static description = 'Finalize parallel Percy builds where PERCY_PARALLEL_TOTAL=-1';
@@ -26,7 +27,10 @@ export class Finalize extends Command {
       process.env.PERCY_PARALLEL_TOTAL = '-1';
     }
 
-    let client = new PercyClient();
+    let client = new PercyClient({
+      clientInfo: `${pkg.name}/${pkg.version}`,
+      environmentInfo: ''
+    });
 
     // ensure that this command is not used for other parallel totals
     if (client.env.parallel.total !== -1) {
--- a/packages/cli-build/src/commands/build/finalize.js
+++ b/packages/cli-build/src/commands/build/finalize.js
@@ -29,7 +29,7 @@ export class Finalize extends Command {
 
     let client = new PercyClient({
       clientInfo: `${pkg.name}/${pkg.version}`,
-      environmentInfo: ''
+      environmentInfo: `node/${process.version}`
     });
 
     // ensure that this command is not used for other parallel totals
--- a/packages/cli-build/src/commands/build/wait.js
+++ b/packages/cli-build/src/commands/build/wait.js
@@ -59,7 +59,7 @@ export class Wait extends Command {
 
     await new PercyClient({
       clientInfo: `${pkg.name}/${pkg.version}`,
-      environmentInfo: ''
+      environmentInfo: `node/${process.version}`
     }).waitForBuild(this.flags, data => {
       this.status(data);
     });
--- a/packages/cli-build/src/commands/build/wait.js
+++ b/packages/cli-build/src/commands/build/wait.js
@@ -1,6 +1,7 @@
 import Command, { flags } from '@percy/cli-command';
 import PercyClient from '@percy/client';
 import logger from '@percy/logger';
+import pkg from '../../../package.json';
 
 export class Wait extends Command {
   static description = 'Wait for a build to be finished. Requires a full access PERCY_TOKEN';
@@ -56,10 +57,12 @@ export class Wait extends Command {
       return this.log.info('Percy is disabled');
     }
 
-    await new PercyClient()
-      .waitForBuild(this.flags, data => {
-        this.status(data);
-      });
+    await new PercyClient({
+      clientInfo: `${pkg.name}/${pkg.version}`,
+      environmentInfo: ''
+    }).waitForBuild(this.flags, data => {
+      this.status(data);
+    });
   }
 
   // Log build status and maybe exit when failed
--- a/packages/cli-snapshot/src/commands/snapshot.js
+++ b/packages/cli-snapshot/src/commands/snapshot.js
@@ -95,6 +95,7 @@ export class Snapshot extends Command {
       }),
 
       clientInfo: `${pkg.name}/${pkg.version}`,
+      environmentInfo: `node/${process.version}`,
       server: false
     });
 
@@ -123,11 +124,7 @@ export class Snapshot extends Command {
           this.log.debug(`-> url: ${snap.url}`);
         }
       } else {
-        this.percy.snapshot({
-          ...snap,
-          clientInfo: `${pkg.name}/${pkg.version}`,
-          environmentInfo: `node/${process.version}`
-        });
+        this.percy.snapshot({ ...snap });
       }
     }
   }
--- a/packages/cli-snapshot/src/commands/snapshot.js
+++ b/packages/cli-snapshot/src/commands/snapshot.js
@@ -123,7 +123,11 @@ export class Snapshot extends Command {
           this.log.debug(`-> url: ${snap.url}`);
         }
       } else {
-        this.percy.snapshot(snap);
+        this.percy.snapshot({
+          ...snap,
+          clientInfo: `${pkg.name}/${pkg.version}`,
+          environmentInfo: `node/${process.version}`
+        });
       }
     }
   }
--- a/packages/cli-snapshot/src/commands/snapshot.js
+++ b/packages/cli-snapshot/src/commands/snapshot.js
@@ -124,7 +124,7 @@ export class Snapshot extends Command {
           this.log.debug(`-> url: ${snap.url}`);
         }
       } else {
-        this.percy.snapshot({ ...snap });
+        this.percy.snapshot(snap);
       }
     }
   }
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -33,6 +33,7 @@ export default class PercyClient {
   env = new PercyEnvironment(process.env);
   clientInfo = new Set();
   environmentInfo = new Set();
+  envWarning = false;
 
   constructor({
     // read or write token, defaults to PERCY_TOKEN environment variable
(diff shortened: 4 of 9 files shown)


packages/cli-build/src/commands/build/wait.js
```diff
@@ -56,10 +57,12 @@ export class Wait extends Command {
       return this.log.info('Percy is disabled');
     }
 
-    await new PercyClient()
-      .waitForBuild(this.flags, data => {
-        this.status(data);
-      });
+    await new PercyClient({
+      clientInfo: `${pkg.name}/${pkg.version}`,
+      environmentInfo: ''
```
Robdel12, line 62: @wwilsman 🚌 Strugglin' to think of what should go here. The node version is passed by the client always, so I'm not sure what would be useful here.

Also feels not great an empty string is all it takes to get around this lol

wwilsman: You're right, empty environment info doesn't feel correct, but also in these instances it's already handled since client adds Node env info automatically.

I was going to suggest adding node info anyway since a `Set` is used and values can only ever be unique. However, I realized that when constructing the user-agent, node info is always added regardless if it already exists (which feels like a bug). This comment is starting to stray from this line of code, so I'll move it to the proper place. 😆

packages/client/src/client.js
```diff
@@ -31,13 +31,14 @@ function validateProjectPath(path) {
...
+  envWarning = false;
 
   constructor({
     // read or write token, defaults to PERCY_TOKEN environment variable
     token,
     // initial user agent info
-    clientInfo = '',
-    environmentInfo = '',
+    clientInfo = [],
+    environmentInfo = [],
```
wwilsman, line 41: (I can't comment on the unchanged folded code below this, but I'll comment here because it's close enough)

Rather than initialize these properties in this constructor, we can move the `new Set()`s to the same place you added the `envWarning`. Then in the constructor, utilize the `this.add*Info()` methods to handle empty strings (which they already do).

packages/client/src/client.js
```diff
@@ -67,6 +68,11 @@ export default class PercyClient {
 
   // Stringifies client and environment info.
   userAgent() {
+    if (!this.envWarning && (this.clientInfo.size === 0 || this.environmentInfo.size === 0)) {
```
wwilsman, line 71: By moving the initial values and utilizing the existing methods, at this point we should be certain that these Sets only contain truthy and unique values. This way empty strings will no longer work around the warning.

packages/client/src/client.js
```diff
@@ -67,6 +68,11 @@ export default class PercyClient {
...
+      this.envWarning = true;
+      this.log.warn('Warning: Missing `clientInfo` and/or `environmentInfo` properties');
+    }
+
     let client = [`Percy/${/\w+$/.exec(this.apiUrl)}`]
       .concat(`${pkg.name}/${pkg.version}`, ...this.clientInfo)
       .filter(Boolean).join(' ');
```
wwilsman, line 78: The forced client info here is safe to assume will always be unique since no other SDK should be providing API or `@percy/client` info. However, if we wanted to be sure it is unique, we'd have to do the same as my suggestion below, only because of the order of forced client info, we would also need to reverse the resulting array.

(Again, I can't comment on unchanged folded code below)

For environment info below, we can add the forced info to the Set so it's always unique. I would suggest cloning the set to avoid mutation, but considering the whole point of Sets is to have unique values, it's safe to just keep adding the forced info blindly. We still want the variable to be an array though, so we can just move the `concat` into `[...envInfo.add(...).add(...)]`. (I really wish GH would let me comment on folded code)

packages/cli-build/src/commands/build/finalize.js
```diff
@@ -26,7 +27,10 @@ export class Finalize extends Command {
       process.env.PERCY_PARALLEL_TOTAL = '-1';
     }
 
-    let client = new PercyClient();
+    let client = new PercyClient({
+      clientInfo: `${pkg.name}/${pkg.version}`,
+      environmentInfo: ''
```
wwilsman, line 32: Now back to these. With those changes, we should be safe to repeat node info. It might help rule out any crazy environments if multiple node versions show up.

wwilsman (changes requested):
I agree with your empty string workaround, but my suggested fix revealed a bug. So read on to see my suggestion to fix it. 😆

packages/client/src/client.js
wwilsman: Also, if we forcibly add both client and env info, the `envWarning` won't be needed since we know an empty value would only ever happen the first time

packages/client/test/client.test.js
```diff
@@ -3,12 +3,15 @@ import { mockgit } from '@percy/env/test/helpers';
...
 import { sha256hash, base64encode } from '../src/utils';
 import mockAPI from './helpers';
+import logger from '@percy/logger/test/helpers';
 
 describe('PercyClient', () => {
   let client;
 
   beforeEach(() => {
     mockAPI.start();
+    logger.mock();
```
Robdel12, line 13: ✂️  dead

packages/client/test/client.test.js
```diff
@@ -3,12 +3,15 @@ import { mockgit } from '@percy/env/test/helpers';
 import PercyClient from '../src';
 import { sha256hash, base64encode } from '../src/utils';
 import mockAPI from './helpers';
+import logger from '@percy/logger/test/helpers';
```
Robdel12, line 6: v

packages/core/src/percy.js
```diff
@@ -235,6 +236,11 @@ export default class Percy {
       throw new Error('Not running');
     }
 
+    if (!this.envWarning && (!options.clientInfo || !options.environmentInfo)) {
+      this.envWarning = true;
+      this.log.warn('Warning: Missing `clientInfo` and/or `environmentInfo` properties');
+    }
+
```
wwilsman, line 243: Thinking about this and seeing other changes, it doesn't quite feel like the right place to warn.

For example, some SDKs iterate over a list and use the `snapshot` method directly. They correctly provide env and client info, but they don't provide it with snapshots. So this warning would incorrectly trigger. 

As another example, the tests that you've changed previously had valid use cases where you can use the snapshot method without client/env info. If that info were to be required for every snapshot, this whole PR would be made invalid by adding the properties to the validation schema as required properties.

Now that I'm noticing this warning, I believe my previous comment about the env info being in the wrong place in the snapshot command was due to this warning. In that command, the env/client info should be passed on init and not with every snapshot. This also explains why you needed to duplicate the client info that was already present.

A better place for this would be in client. Although when discussing this offline, we discovered it was too eager to do so at the construction of the UA header because we didn't want to trigger the warning when a build is created. I think we're on the right path moving it to the same place snapshots are handled, but instead, we should move it to the `client.createSnapshot` method. That way we avoid the check for builds, but still check at the last moment before it should be there.

wwilsman (changes requested):

packages/client/src/client.js
```diff
@@ -31,19 +31,27 @@ function validateProjectPath(path) {
...
-    this.clientInfo = new Set([].concat(clientInfo));
-    this.environmentInfo = new Set([].concat(environmentInfo));
+    // set our env info
+    this.addClientInfo(`Percy/${/\w+$/.exec(this.apiUrl)}`);
+    this.addClientInfo(`${pkg.name}/${pkg.version}`);
+    this.addEnvironmentInfo(`node/${process.version}`);
+    this.addEnvironmentInfo(this.env.info);
+    // set info passed from the SDK
+    this.addClientInfo([].concat(clientInfo).filter(Boolean));
+    this.addEnvironmentInfo([].concat(environmentInfo).filter(Boolean));
```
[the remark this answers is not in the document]
wwilsman: With my other review comment below, that check wouldn't work since we're adding our own info as soon as possible here. We'll have to go back to adding this all lazily when the UA is created. To avoid duplicates (which was the goal of moving these here), we can duplicate the sets immutably with our information appended first.

packages/cli-snapshot/src/commands/snapshot.js
```diff
@@ -123,7 +124,7 @@ export class Snapshot extends Command {
           this.log.debug(`-> url: ${snap.url}`);
         }
       } else {
-        this.percy.snapshot(snap);
+        this.percy.snapshot({ ...snap });
```
wwilsman, line 127: ✂️  Can revert this change

packages/client/src/client.js
```diff
@@ -292,6 +296,11 @@ export default class PercyClient {
     this.addClientInfo(clientInfo);
     this.addEnvironmentInfo(environmentInfo);
 
+    if (!this.envWarning && (this.clientInfo.size === 0 || this.environmentInfo.size === 0)) {
+      this.envWarning = true;
+      this.log.warn('Warning: Missing `clientInfo` and/or `environmentInfo` properties');
+    }
```
wwilsman, line 302: Small nit
```suggestion
    if (!this.envWarning && (!this.clientInfo.size || !this.environmentInfo.size)) {
      this.log.warn('Warning: Missing `clientInfo` and/or `environmentInfo` properties');
      this.envWarning = true;
    }
```

AND/OR, how annoying do we want this warning / what do we think of logging this for every snapshot (getting rid of the `envWarning` variable)? I'm wondering how easy this might be to miss if only logged once (plus this shouldn't be an issue for our own SDKs (and if so, we need to fix them anyway))

packages/core/test/percy.test.js
```diff
@@ -356,7 +362,9 @@ describe('Percy', () => {
       await percy.snapshot({
         name: 'test snapshot',
         url: 'http://localhost:8000',
-        domSnapshot: '<html></html>'
+        domSnapshot: '<html></html>',
+        clientInfo: 'client-info',
+        environmentInfo: 'env-info'
       });
 
       await expectAsync(percy.stop()).toBeResolved();
```
wwilsman, line 370: ✂️  Can revert this file

packages/core/test/snapshot.test.js
```diff
@@ -546,7 +581,9 @@ describe('Snapshot', () => {
         execute: {
           beforeResize: domtest('beforeResize', () => window.innerWidth),
           afterResize: domtest('afterResize', () => window.innerWidth)
-        }
+        },
+        clientInfo: 'client-info',
+        environmentInfo: 'env-info'
       });
 
       await percy.idle();
```
wwilsman, line 589: ✂️  This file too

wwilsman (changes requested):

packages/core/test/percy.test.js
wwilsman: Oh, maybe we need to add these to the `new Percy` init in the before each?

packages/core/test/snapshot.test.js
wwilsman: Same comment here about moving to `new Percy` 😆

wwilsman (approved):
🎮

PR #486, opened 2021-08-10
✨ Install the correct Chromium binary for arm Macs

wwilsman:
## What is this?

For a while, Chromium for Mac was only built on x64 architecture. With the advent of new M1 arm Macs, it wasn't clear if the Percy CLI even worked on these new macs since we rely on Chromium for asset discovery. I looked back into the Google storage APIs and found that there were recent additions that added Mac arm support. This PR adds the appropriate changes so our install script will download the correct version of Chromium when run on arm based Macs.

### DIFF
--- a/scripts/chromium-revision
+++ b/scripts/chromium-revision
@@ -34,6 +34,7 @@ const GH_HEADERS = {
 const G_STORAGE_API_URL = 'https://www.googleapis.com/storage/v1/b/chromium-browser-snapshots/o';
 const G_STORAGE_PREFIXES = {
   darwin: 'Mac',
+  darwinArm: 'Mac_Arm',
   linux: 'Linux_x64',
   win64: 'Win_x64',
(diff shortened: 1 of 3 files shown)


scripts/chromium-revision
```diff
@@ -105,8 +106,9 @@ async function printVersionRevisions(version) {
     async function(state) {
       let platform = state.platforms[state.i - 1];
       if (!platform) return state.value;
+      let rev = state.range[1];
 
-      for (let rev = state.range[1]; rev >= state.range[0]; rev--) {
+      for (; rev >= state.range[0]; rev--) {
```
wwilsman, line 111: I got an error with this variable using the latest LTS node with this script. Apparently, `let` hoisting has changed under the hood. The change makes sense though, as we're referencing `rev` outside of the loop down below.

Robdel12 (approved):
🏁

Issue #478, opened 2021-08-05
1.0.0-beta.63 outputs timeout errors


agrobbin:
## The problem

Since upgrading to Beta 63 this morning, we have started seeing significant degradation in Percy snapshot retrieval:

```
[percy] Encountered an error taking snapshot: [snapshot name redacted]
[percy] Error: Navigation failed: Timeout of 30000ms exceeded.
```

Weirdly the builds in the Percy UI itself seem to be OK, but test run times have skyrocketed due to these timeouts.

## Environment

- Node version: 14.17.3
- `@percy/cli` version: 1.0.0-beta.63
- Version of Percy SDK you’re using: percy-capybara 5.0.0
- OS version: Ubuntu 20.04.2
- Type of shell command-line [interface]: `yarn percy exec -- bundle exec rake`

Robdel12:
Hey @agrobbin! In order to debug this kind of issue we'll need more (and preferably unedited) logs. This usually means there's something hanging in asset discovery for that snapshot

agrobbin:
Happy to provide it! What kind of logs am I looking for? A Percy log that's stored on a CI node while snapshots are retrieved? a Percy build number?

My hunch is that this is somehow related to #469, though I don't know specifically *why* it would've caused this problem suddenly.

Robdel12:
Whoops, I quickly typed a reply before a series of meetings 😅 If you want to attach it to an email, you can send it to robert (at) percy.io.

You can get the verbose logs by adding `--verbose` (I see you're using yarn, so it'd be `yarn percy exec --verbose -- -- bundle exec`). Or setting the env var `PERCY_LOGLEVEL=debug`

Robdel12:
Thanks for the email -- I did get the logs. We dug in and started working on a fix Friday. We'll have a fix out this week (the next day or so 🤞🏼) Basically, when we capture the DOM, we serialize iframes (if possible). When we take that serialized DOM and run asset discovery over it, it hangs because the of iframe CORS requests (we get one response and the rest are forever inflight). That's the gist of it, at least :p 

Anyways, this should be fixed by mid-week. Super big thanks for the logs! That was really helpful for us to crack this one open 😃

agrobbin:
Happy to have been able to help @Robdel12! Let me know if there's anything else I can do. Once a fix is out I will do my best to test it quickly.

Robdel12:
[...]

PR #490, opened 2021-08-11
🐛 Fix situations where asset discovery might hang

wwilsman:
## What is this?

This PR addresses a couple of networking issues in asset discovery.

### Improved timeout logs

Foremost, while debugging a very minimal reproduction of asset discovery hanging, it was very helpful to know which requests were causing the issue. This resulted in an improved network timeout log, which includes a list of the offending request URLs in debug logs.

With the improved logs, it was easy to see that there was an issue with requests for some frame documents. With some additional, very noisy, event logs (not included), it revealed that once the document's response was received, the subsequent request finished events were never observed.

### Site isolation handling

This prompted an investigation into how events are handled with embedded documents, and it was then discovered that Chromium's [site-isolation](https://www.chromium.org/Home/chromium-security/site-isolation) was somehow responsible for the lack of events once the frame was created. Disabling this feature indeed fixed the issue with frame requests hanging, however it felt wrong disabling a security feature like this.

Unrelated, I also found that the `TranslateUI` feature flag changed to just `Translate`. Unfortunately, Chromium will change flags/features between versions like this, and in this particular instance we don't happen to have an explicit test for the translate popup that this was initially meant to suppress. I still didn't add a test in this PR since it mostly focuses on network issues and I just happened to mess with this flag during the above investigation.

To avoid disabling site-isolation, we need to be able to observe subframe requests. This is possible with a couple of very small changes. While pages are already automatically created for newly attached targets, those pages were not automatically initialized. If we want to automatically attach and initialize a subframe for request interception, we do not want to resize the subframe when initialized, so page resizing was moved out of page initialization. To automatically attach and initialize subframes: the `Target.setAutoAttach` command is sent and when a page is created with a parent, it is initialized with inherited options (including network interception options).

### Hanging frame requests

This still leaves the issue of the initial frame request not properly getting cleaned up due to never receiving request finished events. The subframe network also does not see this request since the request is already complete at the time of target creation. To clean up such frame requests, they are registered during the response phase if the resulting response has a different frame. The `Page.frameDetached` event is also watched, and if a frame request has not been handled at the time the event is received, it is assumed to be finished and the appropriate handlers are called.

### Hanging redirects

This concludes the bulk of the changes this PR focuses on, however while testing I also noticed a flakey test related to redirects. This flake also resulted in asset discovery hanging, so it felt appropriate to include a fix with other changes related to the same undesired outcome. 

While debugging this issue, I found that page navigation was sometimes flaking, and that was fixed by a slight refactor to how lifecycle events are handled during page navigation. Once that was fixed, the real issue revealed itself to be related to recent network additions (#469). With some of the logic that was added, there was a gap in which an intercepted redirect might not be properly registered as being intercepted. This is fixed by simply removing the gap in the if-else conditionals. The redirect test was made more robust as a result, although it always seems related to race-y events rather than complex redirect rules.

### DIFF
--- a/packages/core/test/discovery.test.js
+++ b/packages/core/test/discovery.test.js
@@ -965,6 +965,24 @@ describe('Discovery', () => {
         })
       );
     });
+
+    it('does not hang waiting for embedded isolated pages', async () => {
+      server.reply('/', () => [200, {
+        'Content-Type': 'text/html',
+        'Origin-Agent-Cluster': '?1' // force page isolation
+      }, testDOM]);
+
+      server2.reply('/', () => [200, 'text/html', [
+        '<iframe src="http://embed.localhost:8000"></iframe>'
+      ].join('\n')]);
+
+      await percy.snapshot({
+        name: 'test cors',
+        url: 'http://test.localhost:8001'
+      });
+
+      await expectAsync(percy.idle()).toBeResolved();
+    });
   });
 
   describe('with launch options', () => {
--- a/packages/core/test/discovery.test.js
+++ b/packages/core/test/discovery.test.js
@@ -41,7 +41,8 @@ describe('Discovery', () => {
     server = await createTestServer({
       '/': () => [200, 'text/html', testDOM],
       '/style.css': () => [200, 'text/css', testCSS],
-      '/img.gif': () => [200, 'image/gif', pixel]
+      '/img.gif': () => [200, 'image/gif', pixel],
+      '/font.woff': () => [200, 'font/woff', '<font>']
     });
 
     percy = await Percy.start({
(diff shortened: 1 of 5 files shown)


packages/core/test/discovery.test.js
```diff
@@ -556,6 +574,66 @@ describe('Discovery', () => {
...
+      });
+
+      expect(logger.stderr).toContain(jasmine.stringMatching([
+        '^\\[percy:core] Error: Timed out waiting for network requests to idle.',
+        '',
+        '  Active requests:',
+        '  -> http://localhost:8000/img.gif',
+        '',
+        '(?<stack>(.|\n)*)$'
+      ].join('\n')));
```
wwilsman, line 633: 💜

packages/core/test/discovery.test.js
```diff
@@ -897,14 +975,32 @@ describe('Discovery', () => {
...
+            'resource-url': 'http://ex.localhost:8001/img.gif'
           })
         })
       );
     });
+
+    it('does not hang waiting for embedded isolated pages', async () => {
+      server.reply('/', () => [200, {
+        'Content-Type': 'text/html',
+        'Origin-Agent-Cluster': '?1' // force page isolation
```
wwilsman, line 990: This was the smoking gun to reproduce this on localhost

Robdel12 (approved):
🏁 Amazing work, as usual! 

> we don't happen to have an explicit test for the translate popup that this was initially meant to suppress

Should we create an issue to follow up? This one is likely on me -- I should have really checked the changes between Chrome 87 & 91.

Robdel12:
Related issues: #477 #478

PR #491, opened 2021-08-12
🐛 Wait for isolated networks to idle

wwilsman:
## What is this?

Thanks to #490, it now makes it possible to properly capture resources from isolated pages. However, during the network idle check, only the top-level network method is used which only references its own requests. This PR adds tracking isolated pages to the parent page to create a inverse relationship. And also adds a new network method to retrieve all nested page requests during the idle check. These changes make it so network idle will check for nested requests in addition to its own requests.

While here, I also came across a familiar race condition in the page constructor as a result of the page closing during an async call within the constructor. This same race condition exists in the network constructor, however with a check to prevent intentional page closed errors from bubbling. Rather than repeat the pattern, it was improved slightly to make debug logs less confusing. Now, when one of these async constructor calls happen with the page closing, the error is disregarded. All other errors will be logged rather than let bubble since an error from these locations usually stems from an error elsewhere.

### DIFF
--- a/packages/core/src/page.js
+++ b/packages/core/src/page.js
@@ -38,7 +38,7 @@ export default class Page extends EventEmitter {
 
     if (this.parent) {
       this.parent.frames.add(this);
-      this.init(this.parent.options);
+      this._handleCloseRace(this.init(this.parent.options));
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/page.js
```diff
@@ -96,6 +101,8 @@ export default class Page extends EventEmitter {
   async close() {
     if (!this.browser) return;
 
+    this.log.debug('Page closing', this.meta);
```
wwilsman, line 104: I also moved this from the close handler since the handler will be called on every automatically created page while this method is only called when intentionally closing a page. Having it remain in the former location would show several close logs in a row for pages with many frames.

Robdel12 (approved):
🏁

PR #492, opened 2021-08-12
✨ Add --debug discovery flag

wwilsman:
## What is this?

Asset discovery issues are typically debugged with the `--verbose` flag, or by setting the environment variable `PERCY_LOGLEVEL=debug`. While this option prints detailed logs throughout the toolchain, including asset discovery, it only does just that — prints detailed logs. However, when debugging asset discovery issues, the constant uploading of any snapshots results in a hit to screenshot usage. It would be beneficial while debugging asset discovery issues to simply not send any snapshots at all, or not even create a build to begin with.

This PR introduces a `--debug` flag, which not only enables debug logs just like the `--verbose` flag, but additionally disables snapshot uploads via a new core `skipUploads` option. This option mirrors the existing (but unused) `deferUploads` option. The two options do the same thing, with the addition of the `skipUploads` option skipping the deferred uploads altogether. The `--debug` flag could also be an entrypoint for future asset discovery debugging features.

The `start()` method was adjusted slightly to remove the Percy token check. This token doesn't need to be set when not uploading snapshots or creating builds. The token is also checked before any API call by client, such as the first task of the `start()` method, creating a build; this check is also already deferred when uploads are deferred. The creation of the deferred task was also moved outside of the try-catch block since there is nothing to catch until the task is run and awaited on.

### DIFF
--- a/packages/logger/src/index.js
+++ b/packages/logger/src/index.js
@@ -14,7 +14,8 @@ Object.assign(logger, {
   connect: (...args) => new Logger().connect(...args),
   remote: (...args) => new Logger().remote(...args),
   loglevel(level, flags = {}) {
-    if (flags.verbose) level = 'debug';
+    if (flags.debug) level = 'debug';
+    else if (flags.verbose) level = 'debug';
     else if (flags.quiet) level = 'warn';
     else if (flags.silent) level = 'silent';
     return new Logger().loglevel(level);
--- a/packages/logger/test/logger.test.js
+++ b/packages/logger/test/logger.test.js
@@ -171,6 +171,8 @@ describe('logger', () => {
     });
 
     it('can be controlled by a secondary flags argument', () => {
+      logger.loglevel('info', { debug: true });
+      expect(logger.loglevel()).toEqual('debug');
       logger.loglevel('info', { verbose: true });
       expect(logger.loglevel()).toEqual('debug');
       logger.loglevel('info', { quiet: true });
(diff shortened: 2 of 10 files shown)


packages/logger/src/index.js
```diff
@@ -14,7 +14,8 @@ Object.assign(logger, {
   connect: (...args) => new Logger().connect(...args),
   remote: (...args) => new Logger().remote(...args),
   loglevel(level, flags = {}) {
-    if (flags.verbose) level = 'debug';
+    if (flags.debug) level = 'debug';
+    else if (flags.verbose) level = 'debug';
```
wwilsman, line 18: Kept these as separate checks in case we ever decide that one of these should produce even more logs

packages/logger/test/logger.test.js
```diff
@@ -267,7 +269,7 @@ describe('logger', () => {
       expect(helpers.stderr).toEqual([
         jasmine.stringMatching('Warn log \\(\\dms\\)'),
         jasmine.stringMatching('Error log \\(\\dms\\)'),
-        jasmine.stringMatching('Debug log \\((9[0-9]|1[01][0-9])ms\\)'),
+        jasmine.stringMatching('Debug log \\(\\d{2,3}ms\\)'),
```
wwilsman, line 272: 🚩 Flake

Robdel12 (commented):
🏁 This makes my heart happy -- now we can start writing some docs on how to debug. 💯

Robdel12 (approved):

PR #493, opened 2021-08-13
🐛 Fix small config util bugs

wwilsman:
## What is this?

This PR fixes a couple small bugs I found in `@percy/config`:

- When validating arrays as the root subject, the resulting message would contain a path like `0].example`. This was fixed by only stripping leading periods from joined paths rather that stripping the first character blindly.

- When normalizing objects that are not plain objects or arrays (Date, RegExp, etc), the object was inadvertently removed from the result. This was fixed by explicitly testing for the correct type of objects we're looking for. The util was named `hasKeys` in case we ever need to account for other types of iterable objects (but not all of them).

### DIFF
--- a/packages/config/src/utils.js
+++ b/packages/config/src/utils.js
@@ -7,6 +7,11 @@ function create(array) {
   return array ? [] : {};
 }
 
+// Returns true or false if a subject has iterable keys or not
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
🏁

PR #494, opened 2021-08-13
✨ Add deprecate config migrate util

wwilsman:
## What is this?

We currently only have a single place where config migrations log a few deprecation warnings, and that is in `@percy/core`. We also had/have deprecations in the Storybook SDK, although that SDK recently left beta and so the deprecations are/will be removed. In the future, other config deprecations might come up, and so it would be helpful to consolidate how deprecations are handled into a single migrate util passed along just like the other current migrate utils.

This PR also updates `@percy/core` migration deprecations to use the new util.

Depends on #493 as these changes are how I discovered those bugs.

### DIFF
--- a/packages/core/test/unit/config.test.js
+++ b/packages/core/test/unit/config.test.js
@@ -3,13 +3,13 @@ import { configMigration } from '../../src/config';
 
 describe('Unit / Config Migration', () => {
   let mocked = {
+    deprecate: (...a) => mocked.migrate.deprecate.push(a),
     map: (...a) => mocked.migrate.map.push(a),
-    del: (...a) => mocked.migrate.del.push(a),
-    log: require('@percy/logger')('test')
+    del: (...a) => mocked.migrate.del.push(a)
   };
 
   beforeEach(() => {
-    mocked.migrate = { map: [], del: [] };
+    mocked.migrate = { deprecate: [], map: [], del: [] };
     logger.mock();
   });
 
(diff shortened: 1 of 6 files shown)


packages/core/test/unit/config.test.js
```diff
@@ -57,16 +57,9 @@ describe('Unit / Config Migration', () => {
...
-    ]);
-
-    expect(logger.stderr).toEqual([
-      '[percy] Warning: The config option `snapshot.authorization` ' +
-        'will be removed in 1.0.0. Use `discovery.authorization` instead.',
-      '[percy] Warning: The config option `snapshot.requestHeaders` ' +
-        'will be removed in 1.0.0. Use `discovery.requestHeaders` instead.'
+    expect(mocked.migrate.deprecate).toEqual([
+      ['snapshot.authorization', { map: 'discovery.authorization', type: 'config', in: '1.0.0' }],
+      ['snapshot.requestHeaders', { map: 'discovery.requestHeaders', type: 'config', in: '1.0.0' }]
```
Robdel12, line 62: Wonder if we should open an issue like "v1.0.0 deprecations" and list the various deprecations we've made/make sure to remove?

Robdel12 (approved):
🏁

packages/core/test/unit/config.test.js
wwilsman: That's a good idea.

PR #495, opened 2021-08-13
♻ Small revision to the deprecate migrate util api

wwilsman:
## What is this?

The option to deprecate `until: version` reads more accurately than `in: version` and reflects that the user has _until_ that version to move away from the deprecation.


Robdel12 (approved):
hahah 🚲🏠

🏁

PR #496, opened 2021-08-13
✨ Allow automatically deprecating and mapping CLI flags

wwilsman:
## What is this?

At some point (soon), we might want to deprecate some flags. This PR draws inspiration from #494 with a similar API to deprecate and map CLI flags.


Robdel12 (approved):
🏁 closer and closer to v1

PR #497, opened 2021-08-13
🐛 Do not replace initial overrides with empty flags

wwilsman:
## What is this?

The `initialOverrides` value of the shared CLI `#percyrc()` method does not work as intended. Flags always override the initial overrides, even when the flag is undefined, resulting in desired values being removed from the resulting config object.

This PR fixes the issue by aborting the override if the flag is null or undefined. I also borrowed the `set` util from `@percy/config` to set the overridden flag using a much more tested method. An existing test was also updated to remove unnecessary set up.


Robdel12 (approved):
🏁

PR #498, opened 2021-08-13
💥✨ Add include/exclude cli-snapshot options & deprecate files/ignore 

wwilsman:
## What is this?

This PR brings more robust filtering options to the `percy snaphot` command!

Building off of the recent work around adding deprecation utils and fixing various small bugs, both the previous `files` and `ignore` flags are now deprecated and will log as such. The preferred options are now `include` and `exclude`.

These new options are more than just a name change. The flags also work for filtering lists of snapshots when using a single file with the snapshot command. Plus, in addition to the existing glob support, the new options also allow filtering using a regular expression, a regex-like string pattern, or even a filter function. Snapshot list files can now also accept top-level `include`, `exclude`, and `base-url` options.

The static snapshots method previously passed all `files` and `ignore` options to globby. Now, it will only pass `include` and `exclude` to globby if said options only contain strings. All `include` and `exclude` options will continue to be passed along to the static mapping util where new filter logic is used. 

The tests for the snapshot command were also split up into dedicated files covering each functionality of the command. A base test suite for common features, a directory suite for static features, and a file suite for snapshot listing features.

Slightly related — while refactoring a portion of the snapshot command, references to "pages" were replaced with references to "snapshots" instead. Since the merger of core `#capture()` and `#snapshot()` methods, the distinction is moot. Other small naming changes were made as well in reference to internal methods and utils.


Robdel12:
Is the todo for this PR or a follow up task? We have to update docs.percy.io too

Robdel12:
The amount of times I type that is too much. We need to build docs from source 😩

wwilsman:
Both. I need to update the readme for this PR and also the docs site

wwilsman:
Readme here updated. Docs site will be updated along with the next release

Robdel12 (approved):
🔥 🏁 Just keeps getting better

PR #506, opened 2021-08-18
🐛 Fix `CSS_URL_REGEX` for serializing relative URLs in iframes

Robdel12:
## What is this?

Turns out when serializing relative URLs in iframes, the regex we're using _could_ cause the browser to crash/hang. 
For example, the following inline styles from a google maps iframe:

```
position: absolute; z-index: 0; left: 0px; top: 0px; height: 100%; width: 100%; padding: 0px; border-width: 0px; margin: 0px; cursor: url("https://maps.gstatic.com/mapfiles/openhand_8_8.cur"), default; touch-action: pan-x pan-y;
```

Would crash the browser. When plugging this into a regex tester, we saw this error:

image


Robdel12:
Going to take a different direction to eliminate the regex's

PR #512, opened 2021-08-19
🐛 Wait for network idle for responsive asset discovery

### LINKED ISSUE #511: Percy CLI Snapshot should wait for network to be idle between resizes
Reposting the issue from https://github.com/percy/cli/discussions/510#discussioncomment-1208299 (/CC @Robdel12)

when the percy cli iterates over the widths:
https://github.com/percy/cli/blob/feecfde20f8eb0dbc5f645771e0549f58185cb1a/packages/core/src/percy.js#L327-L330
it will cause the browser to at first request a different image from the sourceset / based on CSS but then cancel that request when the image is no longer needed. You may need to make a `await this.network.idle()` call between sizes.

wwilsman:
## What is this?

Thanks to @cancan101, it was pointed out that the browser may cancel responsive image requests between resizes. While we have a test for this, the test wasn't quite comprehensive enough. Adding a third width to the test reveals that intermediary responsive asset requests between the first and the last are indeed cancelled because the viewport is resized too quickly. The first asset in this test likely only succeeds due to page navigation waiting for the `load` event.

This PR adjusts the test so it accurately fails, then duplicates discovery's idle check before resizing. Unfortunately, this fix has a drawback that causes snapshots to take slightly longer than they used to, by a magnitude of `networkIdleTimeout * (widths.length - 1)`. So a network idle timeout of `100ms` across 3 widths would result in the snapshot taking `200ms` longer than it previously did without this change.

Fixes #511

### DIFF
--- a/packages/core/src/percy.js
+++ b/packages/core/src/percy.js
@@ -288,9 +288,6 @@ export default class Percy {
           root = createRootResource(url, domSnapshot);
         }
 
-        // copy widths to prevent mutation later
-        let widths = conf.widths.slice();
-
         // open a new browser page
         page = await this.browser.page({
           networkIdleTimeout: this.config.discovery.networkIdleTimeout,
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
🏁

PR #515, opened 2021-08-20
🐛 Fix reference error when static config options are not defined

wwilsman:
## What is this?

This PR adds a safety check to the migrate function to avoid errors when there are no `static` config options (which usually happens when using any other command).

More evidence that a root-level integration test suite is needed. This error only happens with other commands but each command is tested in isolation (mostly).


Robdel12 (approved):
🏁

PR #516, opened 2021-08-20
✨ Accept sitemaps with the snapshot command!

wwilsman:
## What is this?

This makes it possible to use `percy snapshot <sitemap-url>` to take snapshots of all of the URLs listed by a sitemap.

The changes in the PR leverage the existing `mapStaticSnapshots` util, thus allowing `include`, `exclude`, and `overrides`. Technically, this util also allows a couple of other static options, however those options (such as rewrites and base-url) are not necessary as URLs within a sitemap are typically canonical URLs.

If a fetched sitemap is not an XML document, an error will be thrown. In the future, it's possible this can be relaxed a little since our regexp parser doesn't necessarily need valid XML to work (it simply matches URLs between XML `<loc>` tags).

I also discovered that some sitemaps list a URL twice, with and without a trailing forward slash (`/`). This prompted a filter to remove such duplicate URLs.

Not many tests were necessary for this since it leverages existing utils.

### DIFF
--- a/packages/cli-snapshot/src/commands/snapshot.js
+++ b/packages/cli-snapshot/src/commands/snapshot.js
@@ -2,6 +2,7 @@ import fs from 'fs';
 import path from 'path';
 import PercyConfig from '@percy/config';
 import Command, { flags } from '@percy/cli-command';
+import request from '@percy/client/dist/request';
 import Percy from '@percy/core';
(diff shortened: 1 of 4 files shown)


Robdel12 (approved):
🏁 super fun times! Great stuff

PR #517, opened 2021-08-20
🐛 Fix sitemap unique URL filtering bug

wwilsman:
## What is this?

This fixes a bug in the sitemap URL filtering logic where if a URL ends in a forward slash and does not exist in the sitemap without one, it will be removed from the sitemap.


Robdel12 (approved):
🏁 Classic URL bug eh

PR #518, opened 2021-08-20
📝 Add docs for snapshotting sitemaps

wwilsman:
## What is this?

This adds forgotten docs for the new `percy snapshot <sitemap-url>` feature from #516.

### DIFF
--- a/packages/cli-snapshot/README.md
+++ b/packages/cli-snapshot/README.md
@@ -4,18 +4,18 @@ Snapshot a list or static directory of web pages.
 
 ## Commands
 <!-- commands -->
-* [`percy snapshot PATHNAME`](#percy-snapshot-pathname)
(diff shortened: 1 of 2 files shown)


packages/cli-snapshot/README.md
```diff
@@ -265,3 +265,43 @@ static:
...
         document.querySelector('.button').click()
   ```
+
+### Sitemap URL
+
+When providing a sitemap URL, the document must be an XML document. For sitemap URLs the `--include` and
+`--exclude` flags can be used to filter snapshots. With a Percy config file, the `overrides` option
+is also accepted.
+
+> Tip: Sitemaps can contain **a lot** of URLs, so its best to always start with the `--dry-run` flag
```
Robdel12, line 275: **A LOT** 💯

Robdel12 (approved):
🏁

Issue #507, opened 2021-08-09
Serializing relative URL in iframes: Percysnapshot hang on iframe page with map


yen-tt:
Hello!
I'm using `percySnapshot()` function from the @percy/puppeteer package to create snapshots in the test scripts. Pages with iframe page that contains map, e.g. from Google or Mapbox API, seem to hang indefinitely and so the percy build is never finalized. There's no error in the terminal. I look through the code, and the issue seems to be specifically on the second page.evaluate call in `percySnapshot()`?

I have isolated the general page and test setup in this [repo](https://github.com/yen-tt/test-puppeteer) (map api key is omitted). when running `npm run percy`, the first snapshot for index.html is taken successfully. The second snapshot for iframe_test.html hangs.

PR #519, opened 2021-08-20
🐛 Set serialized iframe base URL for realtive URLs using `<base>`

Robdel12:
## What is this?

Discovered in #507, the regex that was used to parse and covert relative URLs was quite expensive. In that specific issue, it would hang the browser indefinitely. 

Instead of trying to use a regex to serialize the full URL into the document, we can skip all of that serialization work by [adding a `<base>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) to the iframes `<head>`. This will set the base URL for all relative paths within the iframe.

### DIFF
--- a/packages/dom/src/serialize-frames.js
+++ b/packages/dom/src/serialize-frames.js
@@ -1,53 +1,14 @@
 import serializeDOM from './serialize-dom';
 
-// List of attributes that accept URIs that should be transformed
-const URI_ATTRS = ['href', 'src', 'srcset', 'poster', 'background'];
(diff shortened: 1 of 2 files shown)


wwilsman (approved):
Two small changes (I made them for you since you're on vacation).

1. We don't actually have to use the `body` element to get the `baseURI`, we can just pull it off the element passed in (the frame document's root element).
2. There was some leftover test setup for the removed test. I also changed the tests to use the respective iframe's `baseURI`.

🎮

PR #530, opened 2021-08-26
🐛 Do not track or capture requests from service workers

wwilsman:
## What is this?

Our network interception techniques are designed for web pages, but not workers, who do not share the same devtools protocol namespaces. When auto-attaching targets, we do so indiscriminately for all targets. This causes page-network instances to be associated with workers which then causes asset discovery to hang since said instances are not designed for workers.

When creating page instances from auto-attached targets, we can first check against the target type to ensure it is a page target. This has the effect of causing requests from within service workers to never be tracked or captured. This is actually the desirable behavior since JavaScript is normally disabled by default in our infrastructure. If we decide to support capturing service worker requests in the future, we will likely need to approach it differently to how network requests within pages are currently handled.

I also added Network.setBypassServiceWorker which will bypass workers for network requests to avoid custom resource caching implementations since we have our own asset discovery caching.

### DIFF
--- a/packages/core/src/browser.js
+++ b/packages/core/src/browser.js
@@ -256,7 +256,8 @@ export default class Browser extends EventEmitter {
   _handleMessage(data) {
     data = JSON.parse(data);
 
-    if (data.method === 'Target.attachedToTarget') {
+    if (data.method === 'Target.attachedToTarget' &&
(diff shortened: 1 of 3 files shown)


wwilsman:
Superseded by #549

PR #549, opened 2021-09-09
♻️ Refactor page, network, and session handling

wwilsman:
## Purpose

With #530, it was found that worker requests cause the network to hang. The approach in that PR was to ignore requests from workers. After new tests failed on Windows enough times, it was discovered that the test itself was a bit faulty. Once it was made deterministic via the worker updating DOM state, it consistently failed no matter the platform.

After a day of debugging, I discovered that workers inherit the `Fetch` domain of their parent session, but not the `Network` domain. Since both domains need to be enabled in order for request interception to work, requests within workers were being paused but never continued. In order to attempt to disable the inherited `Fetch` domain, we would have to alter how pages are initialized since workers do not have many of the domains that the page class references. Also, if we have to explicitly disable the `Fetch` domain, we could alternatively enable the `Network` domain properly. This alternative fix to the network hanging would also enable capturing requests sent by service workers.

## Approach

All browser targets (pages, frames, workers) are associated with a debugging session when auto-attached. However, the page class only works with page and frame targets due to certain domains used during set up. All page methods that work for any session were moved into a dedicated session class. How the browser handles auto-attached targets now creates session instances rather than page instances. Page instances are now an abstraction on top of specific session instances that also get a dedicated network instance. These single page and network instances are responsible for setting up domains and handlers for all related auto-attached sessions.

These changes resulted in the ability to capture requests from any auto-attached target, including workers and frames. Because of this, the `Page.frameDetached` network handler workaround was no longer needed since the network was now aware of all child session network events as soon as they are available. **However**, this resulted in a regression with capturing isolated frame's root resource. The request for isolated frames starts in the parent session, but once resolved, a new session is attached and all future events come from the new session. Because the parent session never receives a finished event it does not have access to the resource contents, but since the request does not originate from the frame session it complains that the resource's requestId does not exist.

I couldn't find any workaround this time to capture the frame's root resource. It seems like the previous workaround worked because the network events were enabled after fetching the frame resource contents. With these changes however, all events are enabled and attached as soon as possible. The only solution I could find, was to disable site-isolation with a flag, `--disable-site-isolation-trials`. This flag is intended for diagnosing issues with out-of-process iframes by disabling site-per-process isolations. Once disabled, our isolated frame tests passed without workarounds. I suspect this might also help with the flakiness this test sees in CI.

### Unrelated

While researching flags, I documented all of our existing flag args.

While refactoring how the browser handles sessions and pages, I also refactored how args are handled once I realized that `defaultArgs` was per-instance and not a static that needed to avoid being mutated. How browser closing is handled was also refactored once I realized that the CDP method of closing was never actually encountered.

While testing #530 on Windows locally, I found that the browser would frequently close unexpectedly while headed. Turns out, this is because when headed, closing the automatically opened blank page causes the whole browser to close (Windows only). This auto-closing was removed to make debugging on Windows easier.

### DIFF
--- a/packages/core/src/browser.js
+++ b/packages/core/src/browser.js
@@ -7,11 +7,12 @@ import WebSocket from 'ws';
 import rimraf from 'rimraf';
 import logger from '@percy/logger';
 import install from './install';
+import Session from './session';
 import Page from './page';
 
 export default class Browser extends EventEmitter {
   log = logger('core:browser');
-  pages = new Map();
+  sessions = new Map();
   closed = false;
 
   #callbacks = new Map();
@@ -124,14 +125,14 @@ export default class Browser extends EventEmitter {
       }));
     }
 
-    // trigger rejecting pending page callbacks
-    for (let page of this.pages.values()) {
-      page._handleClose();
+    // trigger rejecting pending session callbacks
+    for (let session of this.sessions.values()) {
+      session._handleClose();
     }
 
-    // clear callback and page references
+    // clear own callbacks and sessions
     this.#callbacks.clear();
-    this.pages.clear();
+    this.sessions.clear();
 
     // resolves when the browser has closed
     let closed = new Promise(resolve => {
@@ -181,11 +182,12 @@ export default class Browser extends EventEmitter {
     }));
   }
 
-  async page(options) {
-    // create and attach to a new page target returning the resulting page instance
-    let { targetId } = await this.send('Target.createTarget', { url: 'about:blank' });
+  async page(options = {}) {
+    let { targetId } = await this.send('Target.createTarget', { url: '' });
     let { sessionId } = await this.send('Target.attachToTarget', { targetId, flatten: true });
-    return this.pages.get(sessionId).init(options);
+    let page = new Page(this.sessions.get(sessionId), options);
+    await page._handleAttachedToTarget();
+    return page;
   }
 
   async send(method, params) {
@@ -257,19 +259,20 @@ export default class Browser extends EventEmitter {
     data = JSON.parse(data);
 
     if (data.method === 'Target.attachedToTarget') {
-      // create a new page reference when attached to a target
-      this.pages.set(data.params.sessionId, new Page(this, data));
+      // create a new session reference when attached to a target
+      let session = new Session(this, data);
+      this.sessions.set(session.sessionId, session);
     } else if (data.method === 'Target.detachedFromTarget') {
-      // remove the old page reference when detached from a target
-      let page = this.pages.get(data.params.sessionId);
-      this.pages.delete(data.params.sessionId);
-      page?._handleClose();
+      // remove the old session reference when detached from a target
+      let session = this.sessions.get(data.params.sessionId);
+      this.sessions.delete(session.sessionId);
+      session?._handleClose();
     }
 
     if (data.sessionId) {
-      // message was for a specific page that sent it
-      let page = this.pages.get(data.sessionId);
-      page?._handleMessage(data);
+      // message was for a specific session that sent it
+      let session = this.sessions.get(data.sessionId);
+      session?._handleMessage(data);
     } else if (data.id && this.#callbacks.has(data.id)) {
       // resolve or reject a pending promise created with #send()
       let callback = this.#callbacks.get(data.id);
--- a/packages/core/src/browser.js
+++ b/packages/core/src/browser.js
@@ -129,11 +129,18 @@ export default class Browser extends EventEmitter {
     if (this._closed) return this._closed;
 
     // resolves when the browser has closed
-    this._closed = new Promise(resolve => {
-      /* istanbul ignore next: race condition paranoia */
-      if (!this.process || this.process.exitCode) resolve();
-      else this.process.on('exit', resolve);
-    }).then(() => {
+    this._closed = Promise.all([
+      new Promise(resolve => {
+        /* istanbul ignore next: race condition paranoia */
+        if (!this.process || this.process.exitCode) resolve();
+        else this.process.on('exit', resolve);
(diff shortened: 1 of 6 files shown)


packages/core/src/browser.js
```diff
@@ -94,28 +113,51 @@ export default class Browser extends EventEmitter {
...
 
   async close() {
-    if (this.closed) return;
-    this.closed = true;
+    if (this._closed) return this._closed;
+
+    // resolves when the browser has closed
+    this._closed = Promise.all([
+      new Promise(resolve => {
+        /* istanbul ignore next: race condition paranoia */
```
Robdel12, line 134: The life of automating browsers 😅

packages/core/src/browser.js
```diff
@@ -124,68 +166,37 @@ export default class Browser extends EventEmitter {
...
-    // trigger rejecting pending page callbacks
-    for (let page of this.pages.values()) {
-      page._handleClose();
+    // trigger rejecting pending session callbacks
+    for (let session of this.sessions.values()) {
+      session._handleClose();
     }
 
-    // clear callback and page references
+    // clear own callbacks and sessions
```
Robdel12, line 174: ? nitpick
```suggestion
    // clear our own callbacks and sessions
```

packages/core/src/browser.js
```diff
@@ -124,68 +166,37 @@ export default class Browser extends EventEmitter {
...
-    }));
+    // wait for the browser to close
+    return this._closed;
   }
 
-  async page(options) {
-    // create and attach to a new page target returning the resulting page instance
-    let { targetId } = await this.send('Target.createTarget', { url: 'about:blank' });
+  async page(options = {}) {
+    let { targetId } = await this.send('Target.createTarget', { url: '' });
```
Robdel12, line 195: I swear the about:blank page trolled windows on Chrome 91 (before we opted to go 92)

packages/core/src/network.js
```diff
@@ -6,70 +6,87 @@ import {
...
-    this.page.on('Fetch.authRequired', this._handleAuthRequired);
-    this.page.on('Fetch.requestPaused', this._handleRequestPaused);
-    this.page.on('Network.requestWillBeSent', this._handleRequestWillBeSent);
-    this.page.on('Network.responseReceived', this._handleResponseReceived);
-    this.page.on('Network.eventSourceMessageReceived', this._handleEventSourceMessageReceived);
-    this.page.on('Network.loadingFinished', this._handleLoadingFinished);
-    this.page.on('Network.loadingFailed', this._handleLoadingFailed);
-    this.page.on('Page.frameDetached', this._handleFrameDetached);
-    this.page._handleCloseRace(this.page.send('Network.enable'));
+    this.timeout = options.networkIdleTimeout ?? 100;
```
Robdel12, line 23: Does this differ from our config defaults?

packages/core/src/page.js
```diff
@@ -1,124 +1,40 @@
...
+  // Resize the page to the specified width and height
+  async resize({ width, height }) {
     this.log.debug(`Resize page to ${width}x${height}`);
 
-    await this.send('Emulation.setDeviceMetricsOverride', {
-      deviceScaleFactor,
-      mobile,
+    await this.session.send('Emulation.setDeviceMetricsOverride', {
+      deviceScaleFactor: 1,
+      mobile: false,
```
Robdel12, line 37: One day, we'll probably need to make this configurable (probably when we capture many DOM snapshots per-snapshot). Until then...! 🏃🏼‍♂️

packages/core/src/page.js
```diff
@@ -201,14 +120,15 @@ export default class Page extends EventEmitter {
...
-    let { result, exceptionDetails } = await this.send('Runtime.callFunctionOn', {
-      functionDeclaration: fnbody,
-      arguments: args.map(value => ({ value })),
-      executionContextId: this.contextId,
-      returnByValue: true,
-      awaitPromise: true,
-      userGesture: true
-    });
+    let { result, exceptionDetails } =
+      await this.session.send('Runtime.callFunctionOn', {
```
Robdel12, line 124: _enter the bikeshed zone_

I have very loose feels on this but I liked the function call on the same line as the var 🤨 

_leave the bikeshed zone_

Robdel12 (commented):
So far this is looking 🔥 🔥 I know you've got more coming so I'll wait to 🏁 until then

packages/core/src/browser.js
wwilsman: Nitpick to your nitpick

```suggestion
    // clear this instance's own callbacks and sessions
```

FWIW, "\<verb> \<pronoun>" (without a proper noun) is grammatically correct. 😜

packages/core/src/network.js
wwilsman: No, it's the same. However, our config default will always take precedence over this default. This default is used when borrowing a browser page in other SDKs that don't provide their own idle timeout.

packages/core/src/page.js
wwilsman: On the same line causes it to extend too close to the 100 char range which typically causes overflow and scrolling unless you have a high res monitor or smaller font sizes. If we don't like this syntax on the next line, I'd propose adding a rule to the linter and I can wrap it in parenthesis like we do elsewhere

packages/core/src/browser.js
wwilsman: `about:blank` still loads when passing an empty URL. This was left over from me testing that theory and never cleaned up since it doesn't actually change anything

packages/core/src/browser.js
Robdel12: For that, I’m rejecting the PR

Robdel12 (approved):

PR #556, opened 2021-09-14
🐛 Do not capture scripts or XHR requests when JS is not enabled

wwilsman:
## What is this?

Currently, when using the snapshot command (or any SDK that does not provide their own DOM snapshots), JavaScript is enabled for the local page context during capture to ensure the page is loaded properly, but it is still disabled by default when the snapshot is sent to Percy for screenshot stability.

Because it is enabled during asset discovery, we are capturing JavaScript assets and XHR requests that will never be used in Percy's infrastructure due to JS being disabled there. This PR adds tracking the paused request's resource type and a condition around that type during processing that will only allow specific resources to be captured when JS is disabled. When JS is enabled, all resources will be captured as before (provided none of the other non-capturing conditions are met).

While setting up the changes, the discovery handlers were altered slightly to be provided with the network instance rather than just the network meta info. This allows the discovery handlers to access network and page settings. However, because the page's default javascript setting is determined by the presence of an existing DOM snapshot, this value cannot be used in determining when to capture JS based assets. Instead, the `enableJavaScript` option is passed directly with other network intercept options.

### DIFF
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -4,12 +4,12 @@ import { normalizeURL, hostnameMatches, createResource } from './utils';
 const MAX_RESOURCE_SIZE = 15 * (1024 ** 2); // 15MB
 const ALLOWED_STATUSES = [200, 201, 301, 302, 304, 307, 308];
 
-export function createRequestHandler({ disableCache, getResource }, m) {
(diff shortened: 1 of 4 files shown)


Robdel12 (approved):
🏁 nice

PR #565, opened 2021-09-28
🐛 Fix the eta shown when installing the browser

wwilsman:
## What is this?

Two small logging-related bug fixes when the browser is installed:

1. The calculated seconds shown was for the entire eta estimate, including minutes. This meant that if the eta was 4 minutes, the eta shown would be `04:240`. 

    This was fixed by adding the modulus operator (`%`) to get the remaining seconds.

2. The calculated minutes shown was accidentally rounded when using `.toFixed()`. This is the expected behavior of this method, but not what we want when formatting minutes. 

    This was fixed by using `.toString().split('.')[0]` in its place, which does not round.

I modified an existing test to verify the formatting, which helped to remove a previously ignored line of coverage.

### DIFF
--- a/packages/core/src/install.js
+++ b/packages/core/src/install.js
@@ -17,8 +17,8 @@ function formatBytes(int) {
 
 // Formats milleseconds as "MM:SS"
 function formatTime(ms) {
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
🏁

PR #571, opened 2021-10-04
📝 Update issue template to include new debugging doc

Robdel12:
## What is this?

We now have a nice debugging doc for the SDKs, it would be good to link to this in the issue template to hopefully cut down on the common setup issues we see questions about.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -5,7 +5,14 @@ title: ''
 labels: ''
 assignees: ''
 ---
+
 <!--
+## Have you read the "Debugging SDKs" doc yet?
+
+There are common setup gotchas that happen with Percy's SDKs, it would be worth reading
+the debugging document, which might already answer your question:
+https://docs.percy.io/docs/debugging-sdk
+
 ## Reach out to Percy support instead?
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -5,6 +5,13 @@ title: ''
...
 assignees: ''
 ---
+
+## Have you read the "Debugging SDKs" doc yet?
+
+There are common setup gotchas that happen with Percy's SDKs, it would be worth reading
+the debugging document, which might already answer your question:
+https://docs.percy.io/docs/debugging-sdk
+
 <!--
```
wwilsman, line 8: Should the new paragraph be included in this comment (just move this line to be above it)? That way it won't end up in the issue description when somebody inevitabley doesn't remove excess parts of the template when creating a new issue.

wwilsman (changes requested):

.github/ISSUE_TEMPLATE/bug_report.md
Robdel12: I battled with wrapping it in a comment or not. I guess it doesn't matter, if it's already answered in the doc, we're going to reply with that anyways.

wwilsman (approved):
🎮

PR #644, opened 2021-11-30
🐛 Ensure `networkIdleTimeout` is correctly passed and set

### LINKED ISSUE #643: Network idle timeout config is not being applied
### Discussed in https://github.com/percy/cli/discussions/634

Looks like network idle timeout is not being respected no matter how its passed (cli flag & in the config file). 

<div type='discussions-op-text'>

<sup>Originally posted by **mweitzel** November 24, 2021</sup>
## The problem

Newly introduced `maximum: 750` for `discovery.networkIdleTimeout` breaks existing builds.

SVGs when inlined as html, instead of referenced via url will render intermittently. Setting a higher timeout was a valid workaround until recently.

There are two issues:
1 - inlined SVG elements will intermittently render at low networkIdleTimeouts
2 - a newly introduced timeout maximum breaks existing builds

In this bug report I'd like to **only** address the timeout, as it is a recent regression. I can file another bug for the svg issue.

## Environment

- Node version:
- `@percy/cli` version: >= `@percy/cli/1.0.0-beta.69`
- Version of Percy SDK you’re using: ruby/rails sdk
- OS version: linux-x64
- Type of shell command-line [interface]:   `sh %(#{PERCY_AGENT_BIN} exec -t 1500 -- rspec --tag visual)`
- build IDs illustrating intermittent svg rendering with lower timeout: 14108483, 14137118

## Details

Historically our inlined SVG elements will not render consistently with lower timeouts. After upgrading from beta.68 to beta.69 or beta.70 the problem was re-introduced.

Commit introducing the issue 3c0c0bd5aa871e1a8d2486decf151a78b38520bb

## Debug logs

```
$ bundle exec rake percy:snapshots                                                                                                                                                                                                                                                                                
[redacted]/node_modules/.bin/percy exec -t 1500 --debug -- rspec --tag visual                                                                                                                                                                                                  
[percy:config] Config file not found (0ms)                                                                                                                                                                                                                                                                               
[percy:config] Invalid config: (15ms)                                                                                                                                                                                                                                                                                    
[percy:config] - discovery.networkIdleTimeout: must be <= 750 (0ms)                                                                                                                                                                                                                                                      
[percy:config] Using config:                                                                                                                                                                                                                                                                                             
{                                                                                                                                                                                                                                                                                                                        
  discovery: {                                                                                                                                                                                                                                                                                                           
    networkIdleTimeout: 750                                                                                                                                                                                                                                                                                              
  }                                                                                                                                                                                                                                                                                                                      
} (1ms)              
...     
...                     
```

## Code to reproduce issue

Run a snapshot with an inlined svg on version beta.69 or above.</div>

Robdel12:
## What is this?

It appears `networkIdleTimeout` is currently not being set from config or the CLI. This PR ensures the network idle timeout gets correctly set. 

Will fix #643.

### DIFF
--- a/packages/core/src/snapshot.js
+++ b/packages/core/src/snapshot.js
@@ -47,6 +47,7 @@ export function getSnapshotConfig(percy, options) {
     // only specific discovery options are used per-snapshot
     discovery: {
       allowedHostnames: [uri.hostname, ...config.discovery.allowedHostnames],
+      networkIdleTimeout: config.discovery.networkIdleTimeout,
       requestHeaders: config.discovery.requestHeaders,
       authorization: config.discovery.authorization,
       disableCache: config.discovery.disableCache,
--- a/packages/core/test/discovery.test.js
+++ b/packages/core/test/discovery.test.js
@@ -99,6 +99,38 @@ describe('Discovery', () => {
     ]);
   });
 
+  it('waits for discovery network idle timeout', async () => {
+    percy.setConfig({ discovery: { networkIdleTimeout: 400 } });
+
+    server.reply('/', () => [200, 'text/html', dedent`
+      <html><body><script>
+        let img = document.createElement('img');
+        img.src = '/img.gif';
+        document.body.appendChild(img);
+      </script></body></html>
+    `]);
+
+    await percy.snapshot({
+      widths: [500],
+      name: 'test snapshot',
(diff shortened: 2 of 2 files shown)


wwilsman (approved):
🎮 gg

PR #654, opened 2021-12-08
✨Cancelable @percy/core methods

wwilsman:
## What is this?

Currently, tasks within the internal queue, including snapshots, are cancelable. This is done by utilizing generators as tasks. This PR takes that concept and uses it to make other core methods cancelable as well. Having these methods be cancelable will allow better handling of process termination in the CLI rather than relying on promises to completely settle.

While implementing this, I discovered that I had many hanging Chromium processes after several tests. This prompted me to tighten up some browser internals to alleviate it. Also while implementing this, the upload queue's concurrency was fixed to mirror the snapshot queue, and the `setConfig` method was updated to allow configuring the discovery concurrency after percy has already started.

To make certain core methods cancelable, the queue's cancelable generators were refactored into a common util and used to reimplement the `waitFor` helper. Since this helper is provided to the browser during `execute`, the new `generatePromise` helper must also be provided. This helper wraps an async generator function to become thennable and cancelable. This helper was then used in core methods such as `start`, `stop`, and others to allow them to be cancelable.

While making the `stop` method cancelable, part of it was refactored into a new `flush` method. In the future, this new method could also be used to flush snapshots from the queue when retries are enabled.

One caveat of the cancelable promise-like generator object is some implicit callback promise handling does not work as expected. In these situations, the returned object is awaited on directly to trigger running the underlying generator.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,6 +6,7 @@ rules:
   prefer-const: off
   no-unused-expressions: off
   babel/no-unused-expressions: warn
+  node/no-callback-literal: off
   promise/param-names: off
   semi: [error, always]
   multiline-ternary: off
--- a/packages/cli-exec/src/commands/exec/start.js
+++ b/packages/cli-exec/src/commands/exec/start.js
@@ -32,7 +32,10 @@ export class Start extends Command {
     });
 
     // only stop when terminated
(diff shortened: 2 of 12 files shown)


Robdel12 (approved):
🏁 Verrrrrry nice level up here

PR #655, opened 2021-12-08
🐛 Fix additional snapshots executing string functions

### LINKED ISSUE #600: execute in YAML doesn't work in additionalSnapshots
## The problem

An `execute` function specified in YAML for `additionalSnapshots` does not get run.

## Environment

- Node version: `v15.4.0`
- `@percy/cli` version: `@percy/cli@^1.0.0-beta.70`
- OS version: `macOS`

## Details

There seems to be a regression of: https://github.com/percy/cli/issues/392 where an execute function passed to `additionalSnapshots` is not being run. I am using a yaml file for specifying the snapshots + the execute function definitions. I believe this is because the the value passed execute is not being mapped to a sub key of `beforeSnapshot` as it should be. I can manually do this in the yaml file but then I get a warning: 
```
Invalid snapshot options:
```
as that type is not allowed for in the types defintion:
https://github.com/percy/cli/blob/ceadd44a520750782f6c72197457b4514370a785/packages/core/src/config.js#L183

## Debug logs
When I put a 3s delay in that execute function, I expect to see something like this:
```
[percy:core:network] Wait for 100ms idle (3100ms)
```
however when the function does run, I see:
```
[percy:core:network] Wait for 100ms idle (100ms)
```

wwilsman:
## What is this?

Fixes #600 (`additionalSnapshots.execute` not working with a string value). The config schema for additional snapshots specifically allows strings, functions, or arrays of strings and/or functions. This PR inverts the previous check to only use `execute.beforeSnapshot` if the `execute` option is an object but not an array.

This also fixes an issue where `execute` could be incorrectly inherited by additional snapshots and run multiple times.

### DIFF
--- a/packages/core/src/page.js
+++ b/packages/core/src/page.js
@@ -182,8 +182,9 @@ export default class Page {
     }
 
     // execute any javascript
-    await this.evaluate(typeof execute === 'function'
-      ? execute : execute?.beforeSnapshot);
+    await this.evaluate(
+      typeof execute === 'object' && !Array.isArray(execute)
+        ? execute.beforeSnapshot : execute);
 
     // wait for any final network activity before capturing the dom snapshot
     await this.network.idle();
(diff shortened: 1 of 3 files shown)


Robdel12 (approved):
🏁

PR #665, opened 2021-12-13
🔊 Small logging improvements

wwilsman:
## What is this?

This PR consists of two small changes to `@percy/logger`:

### Give access to additional logger group properties

When a logger group is created with the default `logger(namespace)` function, it only has access to methods that will produce logs (and an additional `shouldLog` util). This PR adds other useful properties that are currently used in various places directly from `logger[property]` equivalents. The top-level properties will still exist, with the difference being that the first argument is bound to the group namespace for new group equivalent methods. This is useful when passing a logging group around without needing to import `logger` directly to have access to these methods. With this change, the `format()` method arguments were ordered to match other method arguments order which makes binding more consistent and straightforward during the logger group creation.

### Use brighter colors when logging

While adding a green color for future packages to use, I noticed a usability issue with our warning color. Being a colorblind person myself, it was hard to distinguish between green and yellow using default terminal colors. On the wiki for ansi color codes, I found [a table that shows how the color codes are represented in various environments](https://en.wikipedia.org/wiki/ANSI_escape_code#Colors). Using this as a guide, I updated our red, yellow, and magenta colors to be brighter versions of themselves.

### DIFF
--- a/packages/logger/src/logger.js
+++ b/packages/logger/src/logger.js
@@ -41,8 +41,8 @@ export default class PercyLogger {
 
   // Change log level at any time or return the current log level
   loglevel(level) {
-    if (!level) return this.level;
-    this.level = level;
+    if (level) this.level = level;
+    return this.level;
   }
 
   // Change namespaces by generating an array of namespace regular expressions from a
--- a/packages/logger/src/logger.js
+++ b/packages/logger/src/logger.js
@@ -78,9 +78,13 @@ export default class PercyLogger {
       .reduce((group, level) => Object.assign(group, {
         [level]: this.log.bind(this, name, level)
       }), {
-        progress: this.progress.bind(this, name),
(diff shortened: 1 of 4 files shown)


packages/logger/src/logger.js
```diff
@@ -90,10 +94,18 @@ export default class PercyLogger {
   }
 
   // Formats messages before they are logged to stdio
-  format(message, debug, level, elapsed) {
+  format(debug, level, message, elapsed) {
```
Robdel12, line 97: Is this publicly used / breaking?

wwilsman: Was previously used by the install script (not since having a dedicated progress method). Otherwise it is only used internally (including for our log dumping test helper)

wwilsman: Probably don't even need to include it in the group properties, but it felt left out of the others

PR #666, opened 2021-12-13
✨ Improve bail ability of config loading

wwilsman:
## What is this?

This wraps and improves the config file searching with a try-catch to handle unexpected not found errors and also returns early during loading when the `bail` option is provided.


Robdel12 (approved):
🏁

PR #668, opened 2021-12-13
🐛 Fix @percy/client retries

wwilsman:
## What is this?

Given zero retries, the retry util would always retry at leasty once. Given any number of retries, the retry util would count the first try against remaining retries. For example, a failed request without retries would always retry at least once, and a failed request with 5 retries would fail once, then retry 4 times.

When properly counting retries, the first try is not counted against retries, and similarly will not retry when given zero retries.


Robdel12 (approved):
🏁 🤓

PR #669, opened 2021-12-13
✅ Fix sdk-utils test server shutdown

wwilsman:
## What is this?

While running tests for the entire repo, I noticed that tests for `@percy/sdk-utils` would occasionally log an error but not fail. After some brief digging, I found it was because the data being asserted against is a buffer, not a string like the comparison is expecting. This PR adds `.toString()` to this check to ensure that it's actually comparing against a string.


Robdel12 (approved):
🏁

Issue #635, opened 2021-11-28
packages/cli uses `@oclif/command` but does not declare in it's dependencies


csi-lk:
## The problem

looks like it's being used here:
https://github.com/percy/cli/blob/master/packages/cli/index.js#L69

seems that it's relying on package hoisting which causes issues with yarn v3

## Environment

- yarn v3.1.1

## Debug logs

```
Error: @percy/cli tried to access @oclif/command, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

Required package: @oclif/command (via "@oclif/command/flush")
Required by: @percy/cli@npm:1.0.0-beta.70 (via ~/dev/csi.lk/.yarn/cache/@percy-cli-npm-1.0.0-beta.70-47d784f1db-64b983337f.zip/node_modules/@percy/cli/bin/)
```

## Code to reproduce issue

Create new repo using yarn v3
Install percy cli
Run percy cli
See issue

Edit: looks like `@oclif/errors` is missing too

wwilsman:
Fixed with #670

PR #670, opened 2021-12-14
🚨 Lint against extraneous dependencies

wwilsman:
## What is this?

In all circumstances where it matters in this repo, the extraneous dependencies are guaranteed to be present and resolve to their expected versions. However, to better support Yarn 2+ and any other potential node loaders that might block on this, this eslint rule will help ensure that packages only import from their specific dependencies.

The rule is disabled in tests where it usually complains about other packages in the workspace which are always available during development.

Where utils were being transitively imported through `@percy/core` dependencies, are now imported through the `@percy/core` package itself (with help of a recent change re-exporting client's request util).


packages/sdk-utils/src/percy-enabled.js
```diff
@@ -4,16 +4,17 @@ import request from './request';
...
-  await logger.remote(() => {
+  await logger.remote(async () => {
     let url = percy.address.replace('http', 'ws');
 
     if (process.env.__PERCY_BROWSERIFIED__) {
       return new window.WebSocket(url);
     } else {
-      let socket = new (require('ws'))(url);
+      /* eslint-disable-next-line import/no-extraneous-dependencies */
+      let { default: WebSocket } = await import('ws');
```
wwilsman, line 14: While here, I also switched this to a dynamic import rather than a commonjs require in preparation for a full ES module adaptation in the future (dynamic imports will work in the current commonjs environment)

Robdel12 (approved):
🏁

csi-lk:
Thanks for the work on this @wwilsman

hellodave76:
Thanks from me as well @wwilsman :+1:

Any idea when this will be released? Or know a workaround until it is?
I'm having the same issue as #635

wwilsman:
@hellodave76 If you're on Yarn 2+, you can use a `.yarnrc` file with a `packageExtensions` field: https://yarnpkg.com/configuration/yarnrc#packageExtensions

hellodave76:
thanks @wwilsman that worked 😃

PR #673, opened 2021-12-14
♻️ Repurpose @percy/cli-command as new CLI framework

wwilsman:
## What is this?

Previously, the `@percy/cli-command` package was used to extend the base `@oclif/command` class with additional Percy specific features. Other Percy CLI plugins then extended from this package to access those shared features such as our own error, process termination, and flag deprecation handling.

Currently, we are experiencing growing pains with some aspects of oclif. Namely, plugin handling, plug-n-play compatibility, autoupdate restrictions, multiple installations, additional flag and argument features, etc.

This PR aims to repurpose the `@percy/cli-command` package to be the foundation framework for Percy CLI, freeing us from the above restrictions and resulting workarounds.

### New features

With complete control over the the CLI parser and runner, we can implement several new features!

- Better, more customizable help output.
- Integrated logger namespace group creation.
- Automatic `@percy/core` instance initialization.
- Built-in global flags such as logging and percy options.
- Immediate `@percy/config` schema and migration registration.
- Support for alternate subcommand syntaxes (e.g. `percy build:finalize`/`percy build finalize`).
- Command cancelability and process termination handling with generator callbacks.
- Decentralized and decoupled plugin handling (this package does not handle plugins).
- And of course, the ability to freely add any other features!

### New syntax

Rather than extending classes, the new package is entirely functional. A command is created using a new command creator function: `command(name[, definition][, callback])`.

[See the readme for complete usage details.](https://github.com/percy/cli/blob/a8751d6560e2aaaa026de0e884ecb7f2dcd60ae5/packages/cli-command/README.md)

### Backwards compatibility

Once `@percy/cli` is updated to use the new `@percy/cli-command`, it can utilize the included `legacyCommand` function to transform class-based commands from CLI plugins into new functional commands that will use our own parser and runner. 

Legacy support includes handling command classes that inadvertently extend from the new command creator function. These classes will include a static `run()` method that wraps command execution with `legacyCommand` to prevent test failures during the period between upgrading to the new syntax.


Robdel12 (approved):
🏁 Big change here, enables a lot of really nice things for us in the future. Also nice to side step the refactor/changes if we were going to stick with oclif (to `oclif/core`). 

A few small comments. We're going to want to test the hell out of this and also we should get all the doc updates drafted prior to release.

PR #676, opened 2021-12-15
♻️ Refactor @percy/cli to use @percy/cli-command

wwilsman:
## What is this?

This PR refactors `@percy/cli` to use the new `@percy/cli-command` foundation.

Historically, this package was used as the entry point to Percy CLI and its various commands through oclif plugins, which are defined by the `package.json` file's `oclif` entry. To support autoloading of plugins found within a project's dependencies, this package was adapted to modify its own package.json at load time before oclif reads it. 

This alleviated requiring oclif's plugins plugin to manually install plugins and allowed us to rely on Node's normal module resolution. However, in part due to oclif's plugin loading, and in part due to our own plugin discovery, the CLI _only_ worked with Node's native module loading (and not plug-n-play environments).

With the new `@percy/cli-command` as a foundation, oclif can be peeled away to give us complete control over plugin loading. This is done by using a function as the `commands` definition argument which when called will search for plugins within the projects dependencies as well as this packages own and sibling dependencies. 

For environments with a native module loading, it will walk through top-level directories within `node_modules` (and recurse into the `@percy` scope). For Yarn 2+, it will use the PnP API to find declared package dependencies. Matching CLI plugins will be named either `@percy/cli-` or `percy-cli-` and contain a `package.json` entry for `@percy/cli`. And for legacy Percy oclif plugins, command classes are found and transformed into functional commands using the `legacyCommand` function from `@percy/cli-command`.


Robdel12 (approved):
🏁 

> With the new @percy/cli-command as a foundation, oclif can be peeled away to give us complete control over plugin loading

Ahh, this is pretty nice. So long, friend.

PR #678, opened 2021-12-16
✨ Add @percy/cli update notice

wwilsman:
## What is this?

This PR adds an update notice when it detects a newer version of the CLI is available.

Before parsing command-line arguments, releases are requested from GitHub's API and compared against the currently installed CLI's version. 

Releases are cached for 3 days before being re-requested, with fallbacks in place so the check still happens if there are errors when reading or writing to the cache file.

The `mockfs` test helper was updated to support a better `spyOn` syntax for `memfs.fs` methods and also automatically mock the current working directory.

### DIFF
--- a/packages/cli/bin/run
+++ b/packages/cli/bin/run
@@ -6,6 +6,8 @@ if (parseInt(process.version.split('.')[0].substring(1), 10) < 12) {
   process.exit(1);
 }
 
-import('@percy/cli').then(async ({ percy }) => {
-  await percy(process.argv.slice(2));
-});
+import('@percy/cli').then(
+  async ({ checkForUpdate, percy }) => {
+    await checkForUpdate();
+    await percy(process.argv.slice(2));
+  });
--- a/packages/cli/src/update.js
+++ b/packages/cli/src/update.js
@@ -1,0 +1,97 @@
+import fs from 'fs';
+import path from 'path';
+import logger from '@percy/logger';
+import { colors } from '@percy/logger/dist/util';
+import pkg from '../package.json';
+
+// filepath where the cache will be read and written to
+const CACHE_FILE = path.join(__dirname, '..', '.releases');
+// max age the cache should be used for (3 days)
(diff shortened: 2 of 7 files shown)


packages/cli/bin/run
```diff
@@ -6,6 +6,8 @@ if (parseInt(process.version.split('.')[0].substring(1), 10) < 12) {
   process.exit(1);
 }
 
-import('@percy/cli').then(async ({ percy }) => {
-  await percy(process.argv.slice(2));
-});
+import('@percy/cli').then(
+  async ({ checkForUpdate, percy }) => {
+    await checkForUpdate();
```
Robdel12, line 11: Such a nice implementation

packages/cli/src/update.js
```diff
@@ -0,0 +1,97 @@
+import fs from 'fs';
+import path from 'path';
+import logger from '@percy/logger';
+import { colors } from '@percy/logger/dist/util';
+import pkg from '../package.json';
+
+// filepath where the cache will be read and written to
+const CACHE_FILE = path.join(__dirname, '..', '.releases');
```
Robdel12, line 8: Any reason to not make this a json file? Would avoid having to stringify and parse it. I was annoyed doing that in the UI & localStorage

Robdel12 (approved):
🏁 A big key to allow global installs!

packages/cli/src/update.js
wwilsman: That would not avoid having to stringify, since `writeFile` still needs a string or buffer.

As for reading and parsing, If you're implying to lean on `require`, that will not work in native ES modules (once we're there). The Node documentation recommends using `fs` like we're doing here, or we could add extra code to enable traditional `require` functionality.

```js
import { createRequire } from "module";
const require = createRequire(import.meta.url);
// ...
let { releases, createdAt } = require('./releases.json');
```

Since it's recommended to explicitly read and parse JSON files anyway, I don't see a reason to add `.json` to the file.

PR #684, opened 2021-12-17
🐛 Fix tail call recursion in core util

wwilsman:
## What is this?

This fixes the `waitFor` util in `@percy/core` to properly return the tail call. Without tail call optimization, error stacks that bubble from within the util can grow exponentially.


Robdel12 (approved):
🏁

PR #685, opened 2021-12-17
🐛 Fix legacy process termination when missing instance method

wwilsman:
## What is this?

Some legacy commands do not define their own `finally` method. If they extend from a class that also does not have a `finally` method (like if they accidentally extend from the `command` function), process termination throws an error.


Robdel12 (approved):
🏁

PR #686, opened 2021-12-17
🧹 Clean up package.json files

wwilsman:
## What is this?

With recent refactoring of the various `@percy/cli*` packages, their respective `package.json` files were cleaned up slightly. This PR cleans up other package's `package.json` files the same way those refactors did for their packages.

Mainly, repository and publish information were moved to the top of the files with other meta info.

For the top-level `package.json` file, the `@oclif/dev-cli` dependency was removed. The top-level `test` script was also adjusted to allow failures since it is only used locally, whereas in CI dedicated parallel jobs are run for each package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,10 @@
   "workspaces": [
     "packages/*"
   ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/percy/cli"
+  },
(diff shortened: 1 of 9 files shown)


Robdel12 (approved):
🏁 o-no-more-clif

PR #718, opened 2022-01-13
🐛 Import local `@percy/cli` to avoid caching/multiple CLI issues

Robdel12:
## What is this?

It looks like `npx` is doing weird caching of `@percy/cli`, so when we do `import('@percy/cli')`, it's possible for the import to be an old cached (incorrect) version of the CLI.

To avoid this, we're going to require the root CLI package so it's _always_ getting the same package & ensure the package matches the invoking bin


wwilsman (approved):
🎮

wwilsman (approved):
🎮

Issue #735, opened 2022-01-20
RangeError: Maximum call stack size exceeded


dmitry12:
## The problem

After upgrading from `@percy/storybook": "^3.3.1` to `"@percy/cli": "^1.0.0-beta.73"`, `"@percy/storybook": "^4.1.0"` and running it on github hosted runner I encountered this error:

```
RangeError: Maximum call stack size exceeded
Exception in PromiseRejectCallback:
/home/runner/work/app/app/storybook/node_modules/@percy/core/dist/utils.js:162
      return yield* check(start);
             ^
```

## Environment

- Node version: node 16.13.2
- `@percy/cli` version: 1.0.0-beta.73
- Version of Percy SDK you're using: 
- If needed, a build or snapshot ID: Build #10120 (Last successful build that was run locally)
- OS version: Ubuntu 20.04.3 LTS (ubuntu-latest)
- Type of shell command-line [interface]: /usr/bin/bash

## Details

- We have ~160 stories in storybook and it always fails on random one. (Usually finishes about 70% of the stories)
- When running locally on more powerful machine it manages to finish, but still fails sometimes with the same error.

## Debug logs

Debug output: https://gist.github.com/dmitry12/fc9d07665b3171dd372857fc9d6bed8e

## Code to reproduce issue

I created repo with 300 sample stories and was able to reproduce the error: https://github.com/dmitry12/percy-bug-repro

```
npm/yarn install
percy storybook --debug ./storybook-static
```

cancan101:
I am also having this error. I was previously on `1.0.0-beta.69` which did not raise the error but does seem to hand indefinitely on:
```
[percy:client] Uploading resource: /percy.1629232245XXX.log... (2ms)
[percy:client] Finalizing snapshot XXX784804... (323ms)
[percy:client] Finalizing snapshot XXX784803... (2ms)
```

filipw01:
@wwilsman I tried pinning `@percy/cli@1.0.0-beta.71` and it was producing the same error

cancan101:
`@percy/cli@1.0.0-beta.71` works for me but then when I made other changes I am still seeing it hanging on the `Finalizing` step.

wwilsman:
I just released [v1.0.0-beta.74](https://github.com/percy/cli/releases/tag/v1.0.0-beta.74) which should hopefully fix the call stack error. 🤞 

@cancan101 if updating doesn't help your hanging issue, maybe it's worth opening another discussion for that

[...]

PR #739, opened 2022-01-24
🐛 Avoid recursion in generators due to no tail-call optimizations

wwilsman:
## What is this?

The check within the `waitFor` util was originally created as a recursive function. With recent changes, the util was converted to make use of the cancelability of generators. However, generators cannot be tail-call optimized like the previous implementation was. This is due to technical reasons within V8 around the nature of how generator's `next()` function handles script evaluation.

Because the `waitFor` util polls and may run for a long time, the lack of tail-call optimization caused the call stack to continuously grow the longer the util took to resolve. Eventually this could/will result in call stack errors (#735).

This PR refactors the generator within the util to use a `while (true)` loop rather than a recursive `yield*`. The infinite loop is broken if the `predicate` passes at least once and at least one more time if `idle` is provided.

### DIFF
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -100,14 +100,16 @@ export function waitFor(predicate, options) {
     ? { timeout: options } : (options || {});
 
   return generatePromise(async function* check(start, done) {
-    if (timeout && Date.now() - start >= timeout) {
-      throw new Error(`Timeout of ${timeout}ms exceeded.`);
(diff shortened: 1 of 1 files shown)


Robdel12 (approved):
> This is due to technical reasons within V8 around the nature of how generator's next() function handles script evaluation.

Ohhhhhhhhhh man, yeah here we are lol

🏁

PR #769, opened 2022-02-07
🐛 Update `bin/run` to not use an esm dynamic import

Robdel12:
## What is this?

Currently the `bin/run` file uses an ESM dynamic import. This wasn't something intended to ship _yet_ (especially since that file isn't touched by babel to transpile). Very soon we'll be going all in on ESM (which will require node 12.22+), but it shouldn't have slipped into the past two releases. This only impacts folks running a very old version of Node.


wwilsman (approved):
🎮

PR #780, opened 2022-02-15
✨ Automatically handle json bodies in client request util

wwilsman:
## What is this?

The client request util is used in various places internally. This PR adds a small feature to automatically stringify json bodies and add the appropriate `application/json` content-type header. 

One planned future use of this feature is within our own tests. As such, response headers were also added to request errors (along with renaming the error response status to match the response property name).


Robdel12 (approved):
🏁

PR #781, opened 2022-02-15
♻️ Core server improvements

wwilsman:
## What is this?

This PR improves upon core's internal API server for future plans to incorporate `cli-snapshot` features into the core snapshot method. The most important feature is the static server, which is a concept used within several SDKs. 

Currently, these SDKs use a third-party library to serve static directories while core's API server is a native Node server with basic routing built in. Initially, core's API server was adjusted to be able to stream files, including basic range support. Util functions from `cli-snapshot` were copied over to handle URL rewrites, and an additional helper function was created mirroring core's `createPercyServer` called `createStaticServer`. The additional helper created a server that rewrote request URLs, checked if a file existed, and served that static file.

While referencing Node docs, I noticed that the native `http.createServer` function accepts custom `IncomingMessage` and `ServerResponse` classes while the function itself was just a wrapper for `new http.Server()`. This inspired me to extend these classes for our desired features, and it actually ended up resulting in much more succinct server code!

### `http.IncomingMessage`

This class is extended to automatically parse its own request URL and body. If the request content-type includes "json" the request body is parsed with `JSON.parse`.

### `http.ServerResponse`

This class is extended to add express-like response methods. 

`#send([status][, headers], body)` allows responding to requests with a status, headers, and a body. 

`#text([status], text)` wraps `send()` with the `text/plain` content-type header.

`#json([status], json)` wraps `send()` with the `application/json` content-type header and stringifies provided json.

`#file([status], filepath)` streams file responses, including any requested ranges and appropriate file headers (content-type, disposition, etc).

### `http.Server`

This class is extended to handle basic routing and middleware. Default middleware is pre-installed to handle CORS requests and CORS headers. 

`#route([method][, pathname], handler)` installs a route handler/middleware. The handler function is called with `IncomingMessage`, `ServerResponse`, and a third argument `next` that can optionally be called to pass execution onto the next route handler (like middleware). 

`#serve([pathname], directory[, options])` installs middleware to serve the provided static directory's contents at the pathname. The `options.cleanUrls` option will strip `index.html` and `.html` from file URLs, and `options.rewrites` allows much finer control over rewrite rules.

`#websocket([pathname], handler)` installs a websocket handler for the specified route.

### `createPercyServer()`

This function was moved into its own dedicated `api.js` file. The function utilizes the new server code from above to create the `localhost:5338/percy` API server, which works just as it did before. In the future, this file will also house a `createStaticServer` function which will handle creating static snapshot servers for SDKs in the same manner.

## Tests

The core test server helper was updated to accept the old initialization and reply syntax. This way most tests can still remain as-is without having to undergo wider changes.

Existing API server tests were moved to an API specific test module. Server-specific tests were moved to and created in a new unit test module. Within this unit test a new package, `quibble`, is utilized for dependency mocking since our currently used dependency mocking library, `mock-require`, is now deprecated.


Robdel12 (approved):
🏁 I dig quibble too!

PR #794, opened 2022-02-24
✨ Serialize current video frame into a `poster` on the `<video>`

### LINKED ISSUE #774: Capture first frame of videos
## The problem

Asset discovery doesn't capture and save videos since it's a non-trivial task and Percy's infrastructure will never play these videos. [The proper fix would be for customers to provide a `poster` for the video.](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#attr-poster) But as a fall back, the SDKs should capture the first frame of the video (via canvas) and set the poster automatically.

Robdel12:
## What is this?

Capturing videos via request interception can be tricky and highly depends on the customers server that's serving the video. Percy's infrastructure will also not play videos for screenshots so it doesn't make too much sense to capture & store large resources like videos. 

This feature will capture the current frame of the videos on page and turn them into a poster image for the video. This is so there's something to display on the `<video>` element when capturing a screenshot in Percy's browsers.

It's possible for this to introduce flakes since this will capture the current frame of the video. If the video is playing when `percySnapshot` is called, we're going to capture different states of the video. In these cases the customer should provide their own `poster`, which will result in the most stable screenshots anyways.

Will close #774

## TODO

- [x] Before release this needs to be documented in the debugging SDKs doc as I suspect this might causing surprising results for some customers. (https://docs.percy.io/docs/debugging-sdks#videos--posters)


packages/dom/test/serialize-videos.test.js
```diff
@@ -0,0 +1,45 @@
...
+  it('does not serialize videos with an existing poster', async () => {
+    withExample(`
+       <video src="base/test/assets/example.webm" id="video" poster="//:0" />
+    `);
+
+    let $video = window.video;
+    await new Promise(r => {
+      if ($video.readyState > 2) r();
+      else $video.addEventListener('canplay', r);
+    });
```
wwilsman, line 31: With util:

```suggestion
    await canPlay(window.video);
```

wwilsman (commented):
Two suggestions

wwilsman (approved):
🎮

PR #804, opened 2022-03-01
✨ Minor @percy/config improvements

wwilsman:
## What is this?

This PR implements a few `@percy/config` improvements which are needed for some soon-to-ship features to work properly.

- Use newer JSON Schema 2019 specification for validations.
- Refactor `getSchema` to respect new spec for schema inheritance.
- Hide some confusing errors as a result of deep schema references.
- Add validation message for new `unevaluatedProperties` spec.
- Handle parsing paths that having missing property names.
- When registering multiple migrations, accept an object instead of a reverse argument tuple.
- Normalize migration deprecation util mapping to allow easier migration composition.

### DIFF
--- a/packages/cli-command/src/command.js
+++ b/packages/cli-command/src/command.js
@@ -39,8 +39,8 @@ function withBuiltIns(definition) {
 
   // add percy specific built-in config options
   if (def.percy) {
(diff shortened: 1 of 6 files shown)


Robdel12 (approved):
🏁

PR #805, opened 2022-03-01
♻️ Improve build failure handling

wwilsman:
## What is this?

When a build fails, the error is logged and the internal queues closed. When attempting to push or upload new snapshots, the queue throws a `Closed` error. The `sdk-utils` package checks if any API error have the message `Closed` to disable the local SDK. These `Closed` errors can occasionally still show up in the logged output which is confusing without any other context (context is usually further up in the logs anyway).

This PR updates this handling to better recommunicate the original error. When builds fail, the reason is cached so other methods can rethrow it rather than let the closed error bubble. Our tests no longer exercised the closed error, so it was made to silently fail for race condition protection. The CLI API was updated to include build details with errors, and the `sdk-utils` package updated to check for the build errors rather than a `Closed` error message.

### DIFF
--- a/packages/core/src/api.js
+++ b/packages/core/src/api.js
@@ -18,6 +18,7 @@ export function createPercyServer(percy, port) {
 
       // return json errors
       return next().catch(e => res.json(e.status ?? 500, {
+        build: percy.build,
         error: e.message,
         success: false
       }));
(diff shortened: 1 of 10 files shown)


Robdel12 (approved):
🏁

PR #806, opened 2022-03-01
♻️ Move browser timeout default to other launch option defaults

wwilsman:
## What is this?

Small housekeeping task I noticed and changed while working on other things. The default browser launch timeout is in an obscure place and as such requires passing around the custom option. Moving it to be next to other defaults allows for some cleaner code in this file.


Robdel12 (approved):
🏁

PR #807, opened 2022-03-01
✅ Update install unit tests to use mock filesystem

wwilsman:
## What is this?

While writing a new test suite for an unrelated feature, I utilized `memfs` and `quibble` to mock the filesystem. For some reason, this caused the install tests to fail (seems one of those libraries doesn't clean up as good as it should). Moving these tests to also use the mock filesystem makes them happy again.


Robdel12 (approved):
🏁

PR #808, opened 2022-03-02
✨ Use default widths instead of throwing an error when missing

wwilsman:
## What is this?

Currently, there is logic to produce a hard error when widths are missing from a snapshot. In reality, when widths are missing from a snapshot (even after invalid widths are scrubbed), the empty array is removed and the default Percy config widths are used. If the user provides empty config widths, they are also removed and the default ends up being used.

Having a hard error on missing widths also makes it difficult to reason about snapshot option overrides with upcoming features where options may or may not apply to a provided set of snapshots. This also raises the question as to why don't we produce hard errors for any other invalid options? Historically there has only ever been a hard error for missing snapshot URLs and widths.

This PR removes the hard error on missing widths while also making it clear that default widths will be used when missing other widths. In the future, we may want to explore throwing a hard error on _any_ invalid option so to not waste screenshot usage when accidentally making configuration mistakes.

This PR also adds logic to deduplicate widths just in case the same width is provided multiple times.

### DIFF
--- a/packages/core/src/snapshot.js
+++ b/packages/core/src/snapshot.js
@@ -3,60 +3,58 @@ import PercyConfig from '@percy/config';
 import { merge } from '@percy/config/dist/utils';
 
 import {
+  configSchema
+} from './config';
+import {
   hostnameMatches,
   createRootResource,
   createPercyCSSResource,
   createLogResource
 } from './utils';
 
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
🏁

PR #809, opened 2022-03-02
✨ Accept alternate core snapshot syntaxes

wwilsman:
## What is this?

This PR makes use of the new static server capabilities within `@percy/core` (#781) and the updated config schema from `@percy/config` (#804) to allow the `snapshot()` method to accept several alternate syntaxes for taking snapshots.

The updated config work is necessary to allow validating options against several related schemas which all share commonly used properties. Before, these properties would need to be repeated for each schema. With the updated schema spec, these schemas can inherit from one another and better reference each other within nested properties.

The new internal static server is started and stopped around a group of snapshots when providing specific snapshot server options. This better couples static servers to core's own internals which will help prevent issues where an SDKs own server might be shut down prematurely before asset discovery is finished.

This new snapshot server feature can become a drop in replacement for the `@percy/cli-snapshot` static server, which also accepts static server options as well as variable snapshot options, including filtering. These additional options are also shared by `@percy/cli-snapshot`'s other two features: snapshot lists and sitemaps. Because of that, the remaining unshared option, specifically `sitemap`, was also made to be a supported core snapshot syntax.

### New syntaxes

The following syntaxes are now all supported by core's `percy.snapshot()` method:

`percy.snapshot(url|{url}|[...url|{url}])` — The current syntax mixed with a new shorthand syntax where a URL string can be used in place of an entire snapshot options object.

`percy.snapshot({snapshots:[...url|{url}]})` — An object containing a top-level `snapshots` array may be provided instead of providing an array of snapshots directly. This allows other useful top-level options, such as a `baseUrl` prepended to snapshot URLs missing their own, and an `options` object or array of objects that apply common snapshot options to one or more snapshots.

`percy.snapshot(sitemap|{sitemap})` — The new sitemap syntax accepts a sitemap URL ending in `.xml` or an object containing a top-level `sitemap` property. This syntax also accepts an `options` property just like the list syntax and additional `include`/`exclude` options to filter the sitemap.

`percy.snapshot({serve})` — The new server syntax accepts an object containing a top-level `serve` property, which should be a path to the static directory to serve. This syntax shares all of the list syntax options from above, with the only exception being the `baseUrl` option. The `baseUrl` option when used with the server syntax should **not** include a protocol or hostname since those are derived from the servers address. This new server syntax also accepts options from `@percy/cli-snapshot` such as `cleanUrls`, `rewrites`, `include`, and `exclude`.

[More details about the accepted options can be found in the updated README](https://github.com/percy/cli/blob/f5525221f4a3cf213e30910a451ce7ed75ba967b/packages/core/README.md#snapshotoptions).

### Next steps

Once this is merged, it could be safely release since the existing syntax did not change. However there is duplicated code in `@percy/cli-snapshot` to delete in a follow up PR which will also move it to fully leaning on these new core syntaxes.

The Gatsby and Storybook SDKs can also benefit from these changes since both of those SDKs start a static server for taking snapshots. While Storybook is a CLI plugin, and safely handles closing the server at the right time, Gatsby will majorly benefit from this as it currently has an issue closing it's server too early.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,6 +4,7 @@ plugins:
   - babel
 rules:
   prefer-const: off
+  no-extra-parens: off
   no-unused-expressions: off
   babel/no-unused-expressions: warn
   import/no-extraneous-dependencies: error
--- a/packages/cli-snapshot/src/config.js
+++ b/packages/cli-snapshot/src/config.js
@@ -1,77 +1,22 @@
-import { snapshotSchema } from '@percy/core/dist/config';
-
-// Common schemas referenced by other schemas
-export const commonSchema = {
-  $id: '/snapshot/cli',
-  $refs: {
-    predicate: {
-      oneOf: [
-        { type: 'string' },
-        { instanceof: 'RegExp' },
-        { instanceof: 'Function' },
-        { type: 'array', items: { $ref: '#/$refs/predicate' } }
-      ]
-    }
-  }
-};
-
 // Config schema for static directories
 export const configSchema = {
   static: {
     type: 'object',
-    additionalProperties: false,
+    $ref: '/snapshot#/$defs/filter',
+    unevaluatedProperties: false,
     properties: {
-      baseUrl: {
-        type: 'string',
-        pattern: '^/',
-        errors: {
-          pattern: 'must start with a forward slash (/)'
-        }
-      },
-      include: {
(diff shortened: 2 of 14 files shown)


Robdel12 (approved):
🏁 I can't wait to get around to updating the gatsby SDK where the server issue trolls the most. Great freggin' work, as usual! 👏🏼 🔥

PR #818, opened 2022-03-08
🐛 Fix core snapshot method compatibility with cli-snapshot command

wwilsman:
## What is this?

Removing components of `@percy/cli-snapshot` and replacing them with the new `@percy/core` snapshot syntaxes revealed a couple of small gaps in implementation details.

- Static server snapshots, when missing a specific list of snapshot, will fallback to the sitemap of the static server, however it is currently missing any provided `baseUrl`. This is fixed by joining the `baseUrl` with `path.posix.join`.

- If a glob pattern for the new `include`/`exclude` options does not start with a leading slash, it has erratic matching behavior. This is made more predictable by only matching on the path basename when missing a leading slash.

- With default options, a build will be created on start and finalized on stop, even if a sitemap, static server, or invalid list of snapshot results in no snapshots being taken. When `deferUploads` is set, a build is not created until the queue is flushed, usually when core is stopping. However, an extra check is required to ensure empty builds are not created if no uploads are queued behind it.

- On the note of a sitemap, static server, or invalid list of snapshot resulting in no snapshots being taken: a new error was added to the snapshot method to alert of missing snapshots.

### DIFF
--- a/packages/core/src/api.js
+++ b/packages/core/src/api.js
@@ -1,4 +1,5 @@
 import fs from 'fs';
+import path from 'path';
 import logger from '@percy/logger';
 import Server from './server';
 import pkg from '../package.json';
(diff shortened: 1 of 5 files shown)


Robdel12 (approved):
🏁

PR #819, opened 2022-03-09
♻️ Update cli-snapshot command to use new core snapshot syntaxes

wwilsman:
## What is this?

This PR refactors `@percy/cli-snapshot` to lean on new core snapshot syntaxes. All extraneous code and dependencies were removed, with any necessary utils being moved to live side by side with the main command function.

Since the new core snapshot syntaxes use an `options` property instead of `overrides`, a migration was added for static and sitemap config file entries.

Blocked by #818


Robdel12 (approved):
🏁

PR #823, opened 2022-03-11
✨ Introduce the ability to abort requests from specificed hostnames (`disallowedHostnames`)

### LINKED ISSUE #773: Introduce disallowed hostnames
## The problem

While we aspire to make request interception as resilient as it possibly can, sometimes this isn't possible (it's an arms race vs the web, in some ways). Especially with live sites that have ad tracking pixels or data collection requests that sometimes hang asset discovery. 

To help solve this, we should introduce the inverse of `allowedHostnames`, which will immediately abort requests from matching hostnames. This will help speed up asset discovery and also make it more reliable while also giving users control over what requests should be made.

Robdel12:
## What is this?

There are many cases where asset discovery can be hung up by remote assets that will never be captured. Most common issues are ad tracking requests or analytics requests. We can't wholesale exclude requests from asset discovery though, since it's possible for remote assets to request local assets (or assets that _are_ included in `allowedHostnames`). 

In order to make asset discovery more stable for customers, we are introducing the ability for them to specify hostnames which will be fast failed and aborted in asset discovery. This makes it so ad trackers or analytic trackers won't hang up or timeout asset discovery. 

Will close #773

### DIFF
--- a/packages/cli-command/src/flags.js
+++ b/packages/cli-command/src/flags.js
@@ -56,6 +56,15 @@ export const allowedHostnames = {
   short: 'h'
 };
 
+export const disallowedHostnames = {
+  name: 'disallowed-hostname',
+  description: 'Disalllowed hostnames to abort in asset discovery',
+  percyrc: 'discovery.disallowedHostnames',
+  type: 'hostname',
+  multiple: true,
+  group: 'Percy'
+};
+
 export const networkIdleTimeout = {
   name: 'network-idle-timeout',
   description: 'Asset discovery network idle timeout',
@@ -95,6 +104,7 @@ export const PERCY = [
 
 export const DISCOVERY = [
   allowedHostnames,
+  disallowedHostnames,
   networkIdleTimeout,
   disableCache,
   debug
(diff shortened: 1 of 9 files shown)


packages/cli-command/src/flags.js
```diff
@@ -56,6 +56,15 @@ export const allowedHostnames = {
   short: 'h'
 };
 
+export const disallowedHostnames = {
+  name: 'disallowed-hostname',
+  description: 'Disalllowed hostnames to abort in asset discovery',
```
wwilsman, line 61:
```suggestion
  description: 'Disallowed hostnames to abort in asset discovery',
```

packages/core/src/config.js
```diff
@@ -46,6 +46,20 @@ export const configSchema = {
           }]
         }
       },
+      disallowedHostnames: {
```
wwilsman, line 49: You'll also need to add a reference around L140 in this same file if we want to support per-snapshot disallowed hostnames.

wwilsman (changes requested):
Small typo and a forgotten config schema

wwilsman (approved):
🎮

PR #830, opened 2022-03-17
♻️ Update package exports and refactor test helpers

wwilsman:
## What is this?

This PR updates packages' `exports` field so consuming modules no longer have to know of a package's structure to import things other than the main export. Mostly, this involves borrowing utils from dependent packages (like core borrowing client's request util, or cli packages borrowing config utils). Simultaneously, some packages can have occasional dependency issues due to them needing to explicitly depend on multiple `@percy/*` packages. While cleaning up exports, common utils have been made to bubble up through `@percy/cli-command` so all SDKs can have a single dependency (`cli-command` or `sdk-utils`).

A lot of utils are also referenced from tests, and some test helpers have implicit global hooks which can cause unexpected issues when importing / not importing the right file. Because of this, all test helpers have also been updated to remove global hooks and instead have explicit setup functions.

One such helper also aims to solve a pain point occasionally experienced when testing: filesystem fixtures. Some tests mock parts of fs (or all of) while other tests use a temporary directory in the real filesystem (which can sometimes fail to clean up). The old config helper, upon import, would automatically install mocks for specific fs methods. However, if it was imported at the wrong time, the original fs module would be cached in the config package before being mocked causing it to fail. Additional other packages use fs and these methods were not mocked in a way that made them friendly with other packages.

A new helper replaces the existing config helpers that mocks all fs methods to use an underlying in-memory filesystem. The helper also allows certain files to bypass mocking which was another pain point experienced when using the in-memory filesystem in certain places. All tests that previously mocked fs or set up temporary directories were migrated to use this new helper.

### Breaking

The updated `exports` field will cause breaking changes for any package that import from `dist` or another file not explicitly meant to be exported. This should only be an issue for Gatsby and Storybook SDKs, which both currently have issues with the next release anyway due to other recent changes.


packages/cli-exec/test/ping.test.js
```diff
@@ -1,10 +1,19 @@
...
+  beforeEach(() => {
+    process.env.PERCY_TOKEN = '<<PERCY_TOKEN>>';
+    setupTest();
+  });
+
   afterEach(async () => {
+    delete process.env.PERCY_TOKEN;
+    delete process.env.PERCY_ENABLE;
+    delete process.env.PERCY_PARALLEL_TOTAL;
+    delete process.env.PERCY_PARTIAL_BUILD;
```
Robdel12, line 16: 👏🏼

Robdel12 (approved):
🏁 Big one here!

PR #831, opened 2022-03-17
✨ Add sdk-utils waitForPercyIdle helper

wwilsman:
## What is this?

When requesting the `/percy/idle` endpoint, if queued snapshots take longer to idle than the request timeout, the request will throw an error. This new `@percy/sdk-utils` helper will recursively call itself when the request times out, resulting in the ability to await on long running idle requests.

There is no logic for aborting the idle wait since the server should always eventually respond with a success or error. Even if the server hangs, killing it will result in a connection refused/disconnect error rather than a timeout. This new helper will always return true when idle, or false when a non-timeout error occurs (which will be logged by the CLI).


Robdel12 (approved):
🏁

PR #832, opened 2022-03-17
✨ Throw an error when a static server directory does not exist

wwilsman:
## What is this?

Performing this check when starting the static server will prevent us and other SDKs from having to do this check before creating a server at all. Without this error, all URLs will simply 404 since the files would not exist at the configured location.


Robdel12 (approved):
🏁

PR #833, opened 2022-03-17
✨ Allow alternate snapshot syntaxes to be cancelable

wwilsman:
## What is this?

Amongst the new alternate snapshot syntaxes, the sitemap and serve options should be cancelable. Both should be cancelable either during URL gathering, or during snapshotting itself. Since all snapshots are queued immediately, relevant cancel code was moved into a dedicated method that is called for each snapshot when canceled.

The snapshot method is meant to be useable asynchronously. When providing missing or invalid snapshot URLs, it will throw immediately. Also calling the method without awaiting on it will queue snapshots regardless. With the current approach to cancelable methods, the `generatePromise` helper returns a promise-like generator that must be await on or yielded to in order to run. This would prevent synchronous snapshots from working as they would not automatically run.

The existing cancelable methods were refactored to be async generators with try-catches to handle cancelations. In the constructor, these methods are overridden to return promises and the original generator methods are moved to a new `.yield` property. This allows using `yield*` with `.yield.<method>` while preserving normal promise behavior of running without having to explicitly await.

### DIFF
--- a/packages/cli-exec/src/exec.js
+++ b/packages/cli-exec/src/exec.js
@@ -56,7 +56,7 @@ export const exec = command('exec', {
     log.warn('Percy is disabled');
   } else {
     try {
-      yield* percy.start();
+      yield* percy.yield.start();
     } catch (error) {
       if (error.canceled) throw error;
       log.warn('Skipping visual tests');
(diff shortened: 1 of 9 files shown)


Robdel12 (approved):
🏁

PR #843, opened 2022-03-24
🐛 Insert serialized CSSOM `style` tag after CSSOM `ownerNode`

Robdel12:
## What is this? 

Currently when we serialize in memory CSS (CSSOM) we only append the serialized styles to the bottom the documents `<head>`. This can cause issues with the ordering of the CSS. In order to fix this we should append the serialized styles _right after_ the CSSOM's `<style>` tag.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5346,10 +5346,10 @@ mime-db@1.45.0:
   resolved "https://registry.yarnpkg.com/mime-db/-/mime-db-1.45.0.tgz#cceeda21ccd7c3a745eba2decd55d4b73e7879ea"
   integrity sha512-CkqLUxUk15hofLoLyljJSrukZi8mAtgd+yE5uO4tqRZsdsAJKv0O+rFMhVDRJgozy+yG6md5KwuXhD4ocIoP+w==
 
-mime-db@1.52.0:
-  version "1.52.0"
-  resolved "https://registry.yarnpkg.com/mime-db/-/mime-db-1.52.0.tgz#bbabcdc02859f4987301c856e3387ce5ec43bf70"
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -5346,10 +5346,10 @@ mime-db@1.45.0:
...
   integrity sha512-CkqLUxUk15hofLoLyljJSrukZi8mAtgd+yE5uO4tqRZsdsAJKv0O+rFMhVDRJgozy+yG6md5KwuXhD4ocIoP+w==
 
-mime-db@1.52.0:
-  version "1.52.0"
-  resolved "https://registry.yarnpkg.com/mime-db/-/mime-db-1.52.0.tgz#bbabcdc02859f4987301c856e3387ce5ec43bf70"
-  integrity sha512-sPU4uV7dYlvtWJxwwxHD0PuihVNiE7TyAbQ5SWxDCB9mUYvOgroQOwYQQOKPJ8CIbE+1ETVlOoK1UC2nU3gYvg==
+mime-db@1.51.0:
+  version "1.51.0"
+  resolved "https://registry.yarnpkg.com/mime-db/-/mime-db-1.51.0.tgz#d9ff62451859b18342d960850dc3cfb77e63fb0c"
+  integrity sha512-5y8A56jg7XVQx2mbv1lu49NR4dokRnhZYTtL+KGfaa27uq4pSTXkwQkFJl4pkRMyNFz/EtYDSkiiEHx3F7UN6g==
```
wwilsman, line 5352: This keeps happening to me locally and I just assumed that depbot updated it at some point. Do we risk depbot opening a PR to change this back to `1.52.0`? Or should we just keep pruning this rollback to `1.51.0` until it goes away? 😆

wwilsman (approved):
🎮

PR #848, opened 2022-03-28
🐛 Wrap `serializeVideos` in try/catch to catch security errors

### LINKED ISSUE #845: Build fails with SecurityError on toDataURL method.
## The problem

Build fails with the following security error. We don't have any occurrence of this method in our code. A quick search seems to suggest it's on Percy side. 

https://github.com/search?q=org%3Apercy+toDataURL&type=code

```
 SecurityError: Failed to execute 'toDataURL' on 'HTMLCanvasElement': Tainted canvases may not be exported.
  Error: Failed to execute 'toDataURL' on 'HTMLCanvasElement': Tainted canvases may not be exported.
      at serializeVideos (eval at <anonymous> (http://localhost:3010/__cypress/tests?p=cypress/support/index.js), <anonymous>:171:30)
      at Object.serializeDOM (eval at <anonymous> (http://localhost:3010/__cypress/tests?p=cypress/support/index.js), <anonymous>:213:7)
      at Context.eval (http://localhost:3010/__cypress/tests?p=cypress/support/index.js:145
```

## Environment

- Node version: v14.17.0
- `@percy/cli` version: `1.0.0-beta.76`
- Version of Percy SDK you’re using: @percy/cypress 3.1.1
- OS version: Ubuntu 20.04.4 LTS

Robdel12:
## What is this?

Sometimes it's not possible to serialize videos into poster due to security (`Tainted canvases may not be exported.`). Wrapping the `serializeVideos` function call in a try/catch will allow the DOM serialization to continue.

Will close #845

### DIFF
--- a/packages/dom/src/serialize-video.js
+++ b/packages/dom/src/serialize-video.js
@@ -9,10 +9,11 @@ export function serializeVideos(dom, clone) {
     let canvas = document.createElement('canvas');
     let width = canvas.width = video.videoWidth;
     let height = canvas.height = video.videoHeight;
+    let dataUrl;
 
     canvas.getContext('2d').drawImage(video, 0, 0, width, height);
+    try { dataUrl = canvas.toDataURL(); } catch {}
 
-    let dataUrl = canvas.toDataURL();
     // If the canvas produces a blank image, skip
     if (!dataUrl || dataUrl === 'data:,') continue;
 
--- a/packages/dom/test/serialize-videos.test.js
+++ b/packages/dom/test/serialize-videos.test.js
@@ -37,4 +37,15 @@ describe('serializeVideos', () => {
     $ = parseDOM(serializeDOM());
     expect($('#video')[0].getAttribute('poster')).toBe(null);
   });
+
+  it('does not hang serialization when there is an error thrown', () => {
+    withExample(`
+       <video src="//:0" id="video" />
+    `);
+
+    spyOn(window.HTMLCanvasElement.prototype, 'toDataURL').and.throwError(new Error('An error'));
+
+    $ = parseDOM(serializeDOM());
+    expect($('#video')[0].getAttribute('poster')).toBe(null);
+  });
 });


packages/dom/src/serialize-dom.js
```diff
@@ -35,7 +35,12 @@ export function serializeDOM(options) {
   let clone = dom.cloneNode(true);
   serializeInputs(dom, clone);
   serializeFrames(dom, clone, { enableJavaScript });
-  serializeVideos(dom, clone);
+
+  try {
+    serializeVideos(dom, clone);
+  } catch (error) {
+    console.log(`Couldn't serialize videos: ${error}`);
```
wwilsman, line 42: Are we sure we want to log this? Could potentially cause some runners like karma or cypress to show these where the user won't know what it's referring to.

We should at least add `[percy]` to this if we want to keep it. But I vote for having a silent failure for now until we implement other serialization warnings and logs like planned.

wwilsman: Also, since this try-catch is wrapped around the function invocation, this would result in any single failed video serialization to cause any remaining videos to not serialize since the loop would be broken. Instead, we should move this try-catch to within the `serializeVideos` loop to give each individual video a chance to serialize.

wwilsman (changes requested):

packages/dom/src/serialize-video.js
```diff
@@ -12,7 +12,11 @@ export function serializeVideos(dom, clone) {
 
     canvas.getContext('2d').drawImage(video, 0, 0, width, height);
 
-    let dataUrl = canvas.toDataURL();
+    let dataUrl;
+    try {
+      dataUrl = canvas.toDataURL();
+    } catch (e) {}
```
wwilsman, line 18: We can drop the `(e)`

```suggestion
    } catch {}
```

wwilsman (approved):
🎮

PR #850, opened 2022-03-28
✨ Automatically include package information with CLI commands

wwilsman:
## What is this?

A very small feature that saves a few lines of code now, but will save more lines of code in the future.

Currently, client and environment information is expected to be provided directly to core from the SDK. The vast majority of SDKs simply use information gathered from `package.json` files.

When automatically loading CLI based SDKs, we already have access to these package.json files. By setting a package information property on the autoloaded SDK command, we can then utilize that to automatically include client information when creating a core percy instance.

With this change, CLI based SDKs will no longer need to provided their own package.json based client information as it will already be available. This only leaves environment information — and since our CLI SDKs are loaded in the same node process, we can automate that too!

Currently, this only saves us from `require('../package.json')`. But once we move to ESM, we will no longer be able to `import '../package.json'` and instead be forced to use `JSON.parse(fs.readFileSync(url.fileURLToPath(...)))`. With ESM, this change saves us from having to import 3 different native modules just to read the package name and version.

### DIFF
--- a/packages/cli-command/src/command.js
+++ b/packages/cli-command/src/command.js
@@ -104,6 +104,7 @@ async function runCommandWithContext(parsed) {
   // include flags, args, argv, logger, exit helper, and env info
   let context = { flags, args, argv, log, exit };
   let env = context.env = process.env;
+  let pkg = command.packageInformation;
   let def = command.definition;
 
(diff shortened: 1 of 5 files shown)


Robdel12 (approved):
🏁

PR #852, opened 2022-03-28
♻️ Move packages to ESM

wwilsman:
## What is this?

As per the title, this PR moves packages and relevant files to ESM. This transition is mostly straightforward, with a few caveats specific to our development environment.

### ESM syntax changes

- File import specifiers need to be valid URLs, including the file extension.
 
- Some dynamic imports were transformed by babel into `require` calls. These needed to be adjusted now that babel is no longer transforming import statements with a default interop.

- All remaining `require` statements were converted to async dynamic imports. This sometimes required a slight refactoring in order to make relevant contexts async as well. Most notable were several test helpers, including `logger.mock()`, `api.mock()`, and core's shared `setupTest()` helper.

- CommonJS globals like `__filename`, `__dirname`, and `require.resolve` no longer exist and need to be replaced with ESM equivalents derived from `import.meta.url`. In order to be used to reference the file or directory name, `url.fileURLToPath` is required to transform the file URL into a filepath. 

  While `require.resolve` is replaced with `path.resolve` in most places, it is still necessary to resolve the path to the `@percy/dom` bundle. For this purpose, the `module.createRequire` function is used in one place, however in the future we may be able to use `import.meta.resolve`.

### Development caveats

- Runtime babel registration and transforms do not work in ESM. Until babel has more official support for ESM and its loader API, we must transform modules using the experimental loader API ourself. This turns out to be fairly simple using a `transformSource` hook and babel's core API.

- Modules were mocked via spying on internal module methods, however these methods can only be stubbed for CommonJS modules. In order to facilitate ESM mocks, we _must_ use Node's experimental loader hooks. Which actually works out since we already need one for babel.

  ES modules are mocked in two primary ways:

  - First, by utilizing a separate global cache for mocking module exports which is checked when resolving an import specifier. If it contains a mock, the `getSource` hook is used to generate a source string that references the mock within the global cache. 

  - Second, is via the existing mock filesystem. To avoid node internals using C bindings to check the real filesystem, if an import specifier exists in the mock filesystem the `getFormat` hook is used to determine if it's a CommonJS module, and if not the `getSource` hook will load it directly from the mock filesystem.

- Because of the new loader hooks, we also no longer need to rely on a babel plugin to transform development paths for testing. However, without the babel plugin, Rollup needs to utilize the same specifier replacement when creating bundles.

- Nock, used to mock external requests, does not work in ESM environments. In its place, jasmine can spy on and stub methods exported by the http modules. Initially, this was only done in the `@percy/client` test helper; however, the technique is useful enough in other places to warrant it be made generic enough to support mocking any request.  

- Some test helpers added named exports to make working with dynamic imports easier, however Rollup doesn't like named exports when bundling as a default export. A new custom Rollup plugin was added to remedy this by stripping named exports from bundled test helpers.

### Miscellaneous

- Of note, `@percy/dom` was not changed to ESM. This is because we need to keep this bundle as an immediately invokable function to be injected with various browser drivers that may or may not like the ESM syntax.

- On windows, import specifiers cannot start with `C:` since it needs to be a valid file URL and the Windows drive letter gets confused for a protocol. Node has a nice warning for this, and since we're using loader hooks we can easily handle it by checking if a specifier starts with the repository root path.

- I also attempted to qualm some other Windows flakes, but there are still a couple annoying ones that can be addressed in some follow up work.


packages/cli/bin/run
```diff
@@ -1,14 +1,13 @@
 #!/usr/bin/env node
 
-if (parseInt(process.version.split('.')[0].substring(1), 10) < 12) {
+if (parseInt(process.version.split('.')[0].substring(1), 10) < 14) {
   console.error(`Node ${process.version} is not supported. Percy only ` + (
-    'supports the current LTS version of Node. Please upgrade to Node 12+'));
+    'supports current LTS versions of Node. Please upgrade to Node 14+'));
```
Robdel12, line 5: 👏🏼

Robdel12 (approved):
🏁 This is awesome. We're on our way to 1.0 now! I can't wait to give this a whirl locally 💯

PR #853, opened 2022-04-01
🐛 Fix supressed validation error causing options to not be scrubbed

wwilsman:
## What is this?

When validating options against schemas, `allOf`, `anyOf`, and `oneOf`, errors are surpressed since they are confusing and the real error will typically bubble from one of the nested schemas. However, if those nested schemas are simple primative types, there is no additional error. And without an error the validation logic does not scrub the option.

This change fixes the issue by adding explicit errors for two of these schemas. In addition, the execute option was relocated so additional snapshots are not allowed to use an object (only a single execute function is allowed for additional snapshots).

Ideally, an error should only be suppressed if a nested error did (or will) bubble. If a would-be-suppressed error would be the only error, it should never be suppressed. Even if the resulting message is confusing, it still hints that a property is invalid (and we can add a better message when that happens). However, this PR implements a short-term fix so we can ship the next version of CLI sooner than later.

### DIFF
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -148,7 +148,7 @@ export const snapshotSchema = {
       }
     },
     exec: {
-      error: 'must be a function, function body, or array',
+      error: 'must be a function, function body, or array of functions',
       oneOf: [
         { oneOf: [{ type: 'string' }, { instanceof: 'Function' }] },
         { type: 'array', items: { $ref: '/snapshot#/$defs/exec/oneOf/0' } }
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -148,6 +148,7 @@ export const snapshotSchema = {
       }
     },
     exec: {
+      error: 'must be a function, function body, or array',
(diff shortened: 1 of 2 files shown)


packages/core/src/config.js
```diff
@@ -148,6 +148,7 @@ export const snapshotSchema = {
       }
     },
     exec: {
+      error: 'must be a function, function body, or array',
```
Robdel12, line 151: > must be a function, function body, or array of functions

? 🤷🏼‍♂️ Take it or leave it

Robdel12 (approved):
🏁 Good stuff

packages/core/src/config.js
wwilsman: I had much more of a mouthful at first

> must be a function, function body, or array of functions or function bodies

I like your compromise

Robdel12: Yo dawg, I heard you like functions so we put a function body inside of a function body so you can evaluate your functions

PR #858, opened 2022-04-01
🐛 Remove relative path from package files

wwilsman:
## What is this?

The 1.0.0 release is botched because paths with leading dots cause yarn and npm to not properly package files. 🙃


Robdel12 (approved):
🏁 Oh boy

PR #863, opened 2022-04-05
🐛 Audit package file exports

wwilsman:
## What is this?

Like #862, except I audited all package.json files. If a file is available in the `exports` field, it should also be included in the packaged files.


Robdel12 (approved):
HELPERS, HELP ME! 🏁

PR #865, opened 2022-04-05
✅ More test adjustments

wwilsman:
## What is this?

Besides misspelling a file in the last PR, we've been running into issue with package manager's `pack` output. This PR fixes the misspelling and adds a command to aid in packing up unreleased packages for testing.

This PR also changes config test helpers to lazily import their required deps. This makes it so packages can import these helpers without installing those deps. They will only need to be installed if one of the helpers that requires it is used.

Along with this import change, core exports were updated to allow import `test/helpers/server` directly rather than through the main `test/helpers` entry. This makes it so the sdk-utils test helper does not end up incidentally importing config and client helpers when it doesn't need them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ oclif.manifest.json
 .local-chromium
 packages/logger/test/client.js
 packages/sdk-utils/test/client.js
+packs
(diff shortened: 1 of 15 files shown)


Robdel12 (approved):
🏁 this is the one

PR #866, opened 2022-04-06
🐛 Core snapshot validation bug fixes

wwilsman:
## What is this?

This PR fixes two small bugs related to snapshot options validation.

1. **Do not pre-scrub empty snapshot lists.** When pre-scrubbed, empty lists are removed and cause warnings due to the missing property. By adding back an empty array when needed, the incorrect validate warning no longer occurs.

2. **Remove method of validation error suppression.** Before schemas could be properly inherited, a check would ensure that errors bubbled from the correct schema. Now with inheritance, errors can bubble from related schemas as well.


Robdel12 (approved):
🏁

PR #868, opened 2022-04-06
🐛 Fix readme generation script

wwilsman:
## What is this?

It seems this script was never properly tested before release. This PR fixes a couple of reference errors and adds a file extension which is required of scripts within ESM packages.


Robdel12 (approved):
🏁

PR #877, opened 2022-04-11
✨@percy/config util improvements

wwilsman:
## What is this?

This PR allows the normalization config util to accept a custom `skip` option and also ensures normalization is always skipped for class instance properties.

Also, to avoid infinite recursion when merging objects with circular references, such objects are ignored completely during a merge. If we absolutely need to handle creating circular references in the future, we might need to refactor the current merge strategy.


Robdel12 (approved):
🏁

PR #878, opened 2022-04-11
✨ Autoload CLI plugin from the current working directory

wwilsman:
## What is this?

When working on CLI plugins, like the Storybook SDK, it can be useful to be able to manually test the plugin without releasing it or linking it in another project. This PR adds package.json detection in the current directory so the `percy` command can now autoload a CLI plugin the plugin's own directory.


Robdel12 (approved):
🏁  this will make both of our lives easier 😆

PR #883, opened 2022-04-12
🐛 Check for alternate casing when determining dom snapshot syntax

wwilsman:
## What is this?

When alternate snapshot syntaxes were introduced, the validation schema is determined by checking for specific snapshot options. For DOM snapshots, we check for the `domSnapshot` option. However, property normalization is based on the determined schema (such as not normalizing headers for example). Therefore, when determining which schema to use, the options have yet to be normalized. For all other syntaxes, this is no problem. For DOM snapshots however, the accepted property can be either `domSnapshot`, `dom-snapshot`, or `dom_snapshot` (as seen [here in the capybara SDK](https://github.com/percy/percy-capybara/blob/master/lib/percy/capybara.rb#L31)).

This PR updates the schema check to account for these possible combinations of dom snapshot properties. A relevant test was also updated to make sure this does not regress again in the future.

### DIFF
--- a/packages/core/src/snapshot.js
+++ b/packages/core/src/snapshot.js
@@ -128,25 +128,18 @@ export async function gatherSnapshots(percy, options) {
 // properties. Eagerly throws an error when missing a URL for any snapshot, and warns about all
 // other invalid options which are also scrubbed from the returned migrated options.
 export function validateSnapshotOptions(options) {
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
AHHH this makes sense 🏁

PR #885, opened 2022-04-14
🐛 Make config normalize skip option behave as expected

wwilsman:
## What is this?

The `@percy/config` util, `normalize`, recent received a `skip` option. However the option was implemented to skip all or nothing; i.e. completely skipping normalizing higher level keys if a lower level key was intended to be skipped. 

By moving the skip check to within the key map loop during normalization, it allows the opportunity to skip any one specific key within the value's key-path array.


Robdel12 (approved):
🏁

PR #895, opened 2022-04-19
🐛 Fix Chrome Revision script

Robdel12:
## What is this?

This fixes the chrome revision script we use to find the right revisions for the asset discovery browser


Robdel12:
Looks like since the script needs `.js` in the file name it's getting linted now. We can ignore it (probably by changing to an eslint ignore file) or we can fix the lint errors in the script

wwilsman (approved):
🎮

PR #896, opened 2022-04-19
✨ Update asset discovery browser to Chromium v96.0.4664.0

Robdel12:
## What is this?

This PR updates the asset discovery browser to v96.0.4664.0 to match the current Chrome browser in Percy's rendering infrastructure.


wwilsman (approved):
🎮

Issue #869, opened 2022-04-07
ESM/TypeScript: Cant get new 1.0.x version to work with Jest/Puppeteer/TypeScript or ESM


tmcconechy:
## The problem

Not sure if this is a bug or support issue if it is is just let me know and i'll redirect my question to discussions or support.
The problem is i cant get your new 1.0.5 (any 1.0.0) version to work anymore in any configuration using jest/puppetteer.
Seems like all the tooling is not in place on those ends so not sure what to do.

## Environment

- Node version: 16.5.0
- `@percy/cli` version: 1.0.5
- Version of Percy SDK you’re using: 1.0.5
- OS version: Mac OS Monterey
- Type of shell command-line [interface]: zsh

## Details

Given below details i cant seem to find a way to use the new version. So sort of at a road block and need to revert to `1.0.0-beta.76` which may not be sustainable if you change things. Also wondering if this should be a major version as it requires major project changes in order to use it

## Comment on Versioning

Also (this took me a while to figure out...)
Reverting to  "@percy/cli": "^1.0.0-beta.76" "@percy/puppeteer": "^2.0.0" works. 

- upgrading @percy/puppeteer from 2.0.0 to 2.0.1 -> gives you the new @percy/cli
- upgrading @percy/cli from 1.0.0-beta.76 to 1.0.5 -> requires a lot of changes and the ESM and TS support with Jest is still a bit experimental and hard to get working (havent yet).

So given that shouldn't these be major versions with breaking changes and more details and examples? Cant seem to find an example of using puppetteer in TS or ESM with jest and percy.

## Debug logs

```
  ● Ids Accordion Percy Tests › should not have visual regressions in new contrast theme (percy)
    Cannot find module '#logger' from 'node_modules/@percy/logger/dist/index.js'
      at Resolver.resolveModule (node_modules/jest-resolve/build/resolver.js:324:11)
```

## Code to reproduce issue

**Without ESM Conversion**
- clone this repo https://github.com/infor-design/enterprise-wc and go to this branch https://github.com/infor-design/enterprise-wc/pull/new/no-esm-conversion
- uses node 16
- `npm i`
- Run command `npx jest -- accordion-percy`
- Run Command `JEST_PUPPETEER_CONFIG=jest-puppeteer.config.cjs node --experimental-vm-modules node_modules/jest/bin/jest.js -- accordion-percy`

**With ESM Conversion**
[...]

PR #916, opened 2022-05-04
🐛 Rollback sdk-utils to commonjs for broader compatibility

wwilsman:
## What is this?

With Node 14 being the oldest currently supported LTS release, ES modules and accompanying features _should_ be supported by the JavaScript ecosystem, as long as you are on a maintained version of Node. However, it seems that this notion is a fallacy, as several JavaScript tools come with their own build systems and bundlers, which themselves may not have parity with Node's LTS releases.

For this reason, the `@percy/sdk-utils` package, which is used by most other JavaScript based SDKs, needs to be rolled back to being a commonjs module for the time being. Some babel packages and files were revived from the graveyard to recreate the same behavior that our custom ESM loader achieves. Since we allow packages to share files without building, babel needs to be configured in way so that individual files have their ES modules preserved when necessary. This is done by utilizing babel overrides and a filter which checks cached package info for files.

Unfortunately, a couple of other changes also needed to be made to make `@percy/sdk-utils` play nicely with third-party build tools and bundlers:

First, the Rollup bundles will now have their dynamic imports scrubbed. While the codepath to these imports isn't ever actually taken, build tools and bundlers will sometimes attempt to build and bundle those imports anyway. The Rollup bundles now replace `import(.*)` with `({})` to act as a no-op.

Second, `@percy/sdk-utils` will no longer include `@percy/logger` directly. Instead, it contains it's own cross-platform mini-logger still capable of remote logging connections. This change allows the logger to continue to remain an ES module without it needing to worry about where sdk-utils gets bundled.

Because of the last change, `@percy/logger` no longer needs to support browsers or remote connections itself since these features only existed for sdk-utils. The package has been refactored to remove these features, and the remaining remote code for accepting connections was moved into core's API WebSocket server. While doing so, I also added a logger specific endpoint for remote connections while maintaining compatibility with root endpoint connections.


Robdel12:
Will resolve #869

Robdel12 (approved):
🏁 Love it, this will solve a little of little annoying issues that proliferated from ESM.

PR #917, opened 2022-05-05
🐛 Fix logger TTY logic for logging progress

wwilsman:
## What is this?

In #916 the logger was refactored slightly. While doing so, I misread Node docs that `cursorTo` returned `this`. 

Turns out, it returns a boolean.


Robdel12 (approved):
🏁 Nice quick turnaround

PR #920, opened 2022-05-09
🐛 Fix sdk-utils test helper log capture

wwilsman:
## What is this?

Since our sdk-utils logger got a little less sophisticated, so did how we mock it during tests. However, the simple mock was a little too simple and would sometimes capture logs that it didn't need to.

This PR updates the sdk-utils test helper to only capture logs during a `logger.log()` call. Usage of `log()` or `log.bind()` within the file was updated to include the `logger` namespace so stubs work no matter when a logging group was created.


Robdel12 (approved):
🏁

PR #921, opened 2022-05-11
🐛 Request fonts as binary via Node

Robdel12:
## What is this?

_todo, add more of a description before undrafting PR_

Sometimes, through the SDKs, fonts are captured as corrupted. This has been a
long standing problem in Percy's SDKs. It turns out we cannot trust the decoding
the browser does in request interception with fonts -- a lot of the time those
fonts network responses are already corrupted by the time our SDK wants to save
the response body.

In order to work around this, we use Node to request the font files directly as
binary data. This solves all issues we've seen with fonts being captured as
corrupt.

### DIFF
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -1,4 +1,6 @@
+import mime from 'mime-types';
 import logger from '@percy/logger';
+import { request as makeRequest } from '@percy/client/utils';
 import { normalizeURL, hostnameMatches, createResource } from './utils.js';
 
 const MAX_RESOURCE_SIZE = 15 * (1024 ** 2); // 15MB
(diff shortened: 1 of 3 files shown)


packages/core/src/discovery.js
```diff
@@ -79,7 +81,18 @@ export function createRequestFinishedHandler(network, {
           return log.debug(`- Skipping disallowed resource type [${request.type}]`, meta);
         }
 
-        resource = createResource(url, body, response.mimeType, {
+        // Ensure we're getting a proper mimetype if the server or
+        // asset discovery browser is sending `text/plain`
+        let resourceMimeType = response.mimeType === 'text/plain' ? mime.lookup(response.url) : response.mimeType;
```
wwilsman, line 86: This could be a boolean `false` if a URL without an extension has a `text/plain` mimetype.

```suggestion
        // a text/plain mimetype could be a fallback, try to infer one from the response url
        let mimeType = (url.response.mimeType === 'text/plain' && mime.lookup(response.url)) || response.mimeType;
```

Changing it from a ternary will ensure that `response.mimeType` is used if `mime.lookup()` returns `false`.

packages/core/src/discovery.js
```diff
@@ -79,7 +81,18 @@ export function createRequestFinishedHandler(network, {
...
+
+        if (resourceMimeType?.includes('font')) {
+          // Request & save font files directly as binaries rather than relying on
+          // the asset discovery browser to decode font responses properly.
+          // Without this, some fonts may be captured as corrupted
```
wwilsman, line 91:
```suggestion
          // font responses from the browser may not be properly encoded, so request them directly
```

packages/core/src/discovery.js
```diff
@@ -90,7 +103,7 @@ export function createRequestFinishedHandler(network, {
         });
 
         log.debug(`- sha: ${resource.sha}`, meta);
-        log.debug(`- mimetype: ${resource.mimetype}`, meta);
+        log.debug(`- mimetype: ${resourceMimeType}`, meta);
```
wwilsman, line 106: Not necessary by this point because we pass `resourceMimeType` into `ceateResource` above, so `resource.mimetype === resourceMimeType`.

wwilsman (changes requested):
Potential bug to fix. Your call if you wanna add another test for it. Other suggestions are just a personal opinion on the comments.

wwilsman (approved):
🎮

PR #922, opened 2022-05-12
✨ Add additional snapshot command exec helpers

### LINKED ISSUE #857: Add additional execute helpers
We currently support a `waitFor` helper within `execute` functions. While this is a great building block used to support the `waitForSelector` option (as well as being used elsewhere throughout core), it would be nice to have some other execute helpers. 

In particular, a common problem is custom lazy loaded images that require the page to be scrolled. An excellent candidate for a new helper or two.

wwilsman:
## What is this?

This PR adds additional helpers provided with scripts evaluated on pages using the core browser. 

In practice, these helpers are available to any `execute` option when using the `percy snapshot` command. For our own internal use, we can utilize these helpers when given their associated snapshot option (like `waitForSelector`). We could also potential use them while scraping for snapshots in other SDKs (like Storybook).

While addressing these helpers and in making them publicly accessible, I also decided to address how the current `generatePromise` function works. Currently, it returns a promise-like generator that can be "canceled" and have registered handlers called. However, only one place (besides tests) actually uses this feature: the queue. After updating the util, the queue was updated to utilize replacement utils instead.

Will close #857 

### Helpers:

- `generatePromise(generator[, signal][, callback])` - Updated to accept an AbortSignal, available in browser environments and later Node versions. However for some current versions of Node, a polyfill is required to support the minimum features needed to cancel a generator loop. The `generatePromise` util can now also receive an optional Node-style callback argument, which is utilized by the queue.

- `waitFor(predicate[, options])` - Split into two separate functions. A new util, `yieldFor`, which is an async generator function that behaves exactly the same as `waitFor`. The main difference is instead of running automatically, it runs the predicate check only when iterated. The existing `waitFor` util is now just a wrapper for `generatePromise(yieldFor(...))`.

- `waitForTimeout(timeout[, value])` - A promisified version of `setTimeout` (without a callback). The first argument is the timeout and the second optional argument is the value to resolve.

- `waitForSelector(selector[, timeout])` - Alias for `waitFor(() => document.querySelector(selector), timeout)`.

- `waitForXPath(selector[, timeout])` - Like `waitForSelector(...)`, but uses `document.evaulate(...)` instead.

- `scrollToBottom([options][, onScroll])` - A browser util to scroll to the bottom of the page, one window at a time. Any provided `options` are passed along to `window.scrollTo(...)`, and any provided `onScroll` function will be awaited on after each window segment is scrolled.

- `config` - An object containing global Percy config options. Also available as `window.__PERCY__.config`.

- `snapshot` - An object containing the current snapshot options. Only available when `page.snapshot` is called internally (for snapshot `execute` options). Also available as `window.__PERCY__.snapshot`.

### Also changed in this PR:

The change from "cancelable" generators to AbortController highlighted that our intent was always to _abort_ (cancel implies it has not yet started). With this, I also updated `@percy/cli-command` to use an AbortController for dealing with process signals. In combination with the new `generatePromise` util, the code for process signal handling was actually cut back significantly. However, any code that previously relied on `error.canceled` needed to be updated to check the error name, which is the recommended way for checking for an `AbortError` (emitted by an `AbortSignal`).

In order to get a global config reference in the right place, the browser class constructor was refactored slightly to accept a core percy instance. Other constructor code was moved into the launch method, and some process code was moved into a spawn method.

### To Do:

These helpers should be publicly documented. The snapshot command itself also lacks good documentation, so this could be an opportunity to write better documentation around that in general.


Robdel12 (approved):
🏁 Slick stuff! Documenting this will be really nice and give folks using `snapshot` methods (the CLI, storybook, etc) will love these helpers 🔥

PR #923, opened 2022-05-13
🐛 Fix regex matching accepting non-regex patterns

wwilsman:
### What is this?

When given a path such as `/index.html`, the fallback regular expression matching would match anything containing the string, including paths such as `/other-path/index.html`.

This PR fixes the issue by not using the predicate as a fallback regular expression. To use a regular expression pattern, the pattern should look like: `/regex-here/` (like a normal JS regular expression).

I also adjusted two merge snapshot option paths that I noticed were being logged as empty arrays.


Robdel12 (approved):
🏁

PR #936, opened 2022-05-24
✨ Introduce `scope` snapshot option

Robdel12:
## What is this?

Add a snapshot option to scope a screenshot down to a specific selector. This currently won't work since it relies on infrastructure & API changes to be made, but now the SDKs will pass this option to the API.

### DIFF
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -25,6 +25,9 @@ export const configSchema = {
       },
       enableJavaScript: {
         type: 'boolean'
+      },
+      scope: {
+        type: 'string'
       }
     }
   },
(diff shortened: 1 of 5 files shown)


packages/core/src/config.js
```diff
@@ -25,6 +25,10 @@ export const configSchema = {
       },
       enableJavaScript: {
         type: 'boolean'
+      },
+      scope: {
+        type: 'string',
+        default: ''
```
wwilsman, line 31: Any reason for this default? Seems like we could remove this and the merge with defaults and it would behave exactly the same.

Robdel12: No specific reason -- I thought it needed the default but I now see `enableJavaScript` doesn't have one either

packages/core/src/snapshot.js
```diff
@@ -194,6 +194,7 @@ export async function getSitemapSnapshots(options) {
 export function getSnapshotConfig(percy, options) {
   return PercyConfig.merge([{
     widths: configSchema.snapshot.properties.widths.default,
+    scope: configSchema.snapshot.properties.scope.default,
```
wwilsman, line 197: No longer needed

wwilsman (approved):
Everything else looks good

PR #938, opened 2022-05-25
🐛 Add back browser executable environment variable

wwilsman:
## What is this?

This area of the browser class was refactored slightly with #922 and the environment variable was accidentally forgotten about and never added back. An added test should ensure that doesn't happen again.


Robdel12 (approved):
🏁

PR #962, opened 2022-06-16
🐛 Fix `@percy/sdk-utils` exports field for browser based SDKs

Robdel12:
## What is this?

This PR fixes browser based SDKs from grabbing the wrong bundle for the wrong environment. Currently they may try to use the node bundle (`index.js`), when it should be using the browser complied version (`bundle.js`). https://webpack.js.org/guides/package-exports/#providing-different-versions-depending-on-target-environment


wwilsman (approved):
🎮

PR #970, opened 2022-06-22
✨ Capture different pixel densities in asset discovery

Robdel12:
## What is this?

With the addition of real native devices in our rendering infrastructure, we now need to capture 2x++ pixel images for pages to render properly (since the real devices have high pixel density screens).

With the new `devicePixelRatio` config option, we now will resize the browser to each specified width at 1x and also whatever the passed value is for `devicePixelRatio`. We also run both `beforeResize` and `afterResize` events with the additional scale resizes. These assets are discovered in their own tab, separate from 1x devicePixelRatio assets to ensure that the right scaled assets are loaded (rather than trying to resize and reset the devicePixelRatio)

### DIFF
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -28,6 +28,10 @@ export const configSchema = {
       },
       scope: {
         type: 'string'
+      },
+      deviceScaleFactor: {
+        type: 'integer',
+        default: 1
       }
     }
   },
@@ -137,6 +141,7 @@ export const snapshotSchema = {
         minHeight: { $ref: '/config/snapshot#/properties/minHeight' },
         percyCSS: { $ref: '/config/snapshot#/properties/percyCSS' },
         enableJavaScript: { $ref: '/config/snapshot#/properties/enableJavaScript' },
+        deviceScaleFactor: { $ref: '/config/snapshot#/properties/deviceScaleFactor' },
         discovery: {
           type: 'object',
           additionalProperties: false,
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -29,7 +29,7 @@ export const configSchema = {
       scope: {
         type: 'string'
       },
-      deviceScaleFactor: {
+      devicePixelRatio: {
         type: 'integer'
       }
     }
@@ -140,7 +140,7 @@ export const snapshotSchema = {
         minHeight: { $ref: '/config/snapshot#/properties/minHeight' },
         percyCSS: { $ref: '/config/snapshot#/properties/percyCSS' },
         enableJavaScript: { $ref: '/config/snapshot#/properties/enableJavaScript' },
-        deviceScaleFactor: { $ref: '/config/snapshot#/properties/deviceScaleFactor' },
+        devicePixelRatio: { $ref: '/config/snapshot#/properties/devicePixelRatio' },
         discovery: {
           type: 'object',
           additionalProperties: false,
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -30,8 +30,7 @@ export const configSchema = {
         type: 'string'
       },
       deviceScaleFactor: {
-        type: 'integer',
-        default: 1
+        type: 'integer'
       }
     }
   },
--- a/packages/core/src/page.js
+++ b/packages/core/src/page.js
@@ -35,14 +35,9 @@ export class Page {
     this.log.debug('Page closed', this.meta);
   }
 
-  async reload() {
-    await this.session.send('Page.reload');
-    this.log.debug('Page reloaded', this.meta);
-  }
-
   // Resize the page to the specified width and height
   async resize({ width, height, deviceScaleFactor = 1, mobile = false }) {
-    this.log.debug(`Resize page to ${width}x${height} at ${deviceScaleFactor}x scale`);
+    this.log.debug(`Resize page to ${width}x${height} @${deviceScaleFactor}x`);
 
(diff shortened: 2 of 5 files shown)


packages/core/src/config.js
```diff
@@ -28,6 +28,10 @@ export const configSchema = {
       },
       scope: {
         type: 'string'
+      },
+      deviceScaleFactor: {
+        type: 'integer',
+        default: 1
```
wwilsman, line 34: When we add defaults to the schema like this, the option and its default will be added to freshly generated percy config files. Without this, we'd have to change the few logic gates slightly, but it won't end up in default config files.

Your call whether you think it should be included in a fresh config file, but I thought I'd at least mention how schema defaults can affect them.

packages/core/src/snapshot.js
```diff
@@ -323,6 +324,70 @@ function waitForDiscoveryNetworkIdle(page, options) {
 // Used to cache resources across core instances
 const RESOURCE_CACHE_KEY = Symbol('resource-cache');
 
+function* triggerResourceRequests(page, snapshot) {
+  // copy widths to prevent mutation later
+  let widths = snapshot.widths.slice();
+  let initialWidth = widths.shift();
```
wwilsman, line 330: The mutation this comment is referring to immediately follows (so it's not really later). Also can use rest spread here now which won't mutate the existing array.

```suggestion
  let [initialWidth, ...widths] = snapshot.widths;
```

packages/core/src/snapshot.js
```diff
@@ -323,6 +324,70 @@ function waitForDiscoveryNetworkIdle(page, options) {
...
+    // when any execute options are provided, inject snapshot options
+    /* istanbul ignore next: cannot detect coverage of injected code */
+    yield page.eval((_, s) => (window.__PERCY__.snapshot = s), snapshot);
+    yield page.evaluate(snapshot.execute.afterNavigation);
+  }
+
+  // resize and set pixel ratio for higher density images
+  if (snapshot.deviceScaleFactor !== 1) {
+    yield waitForDiscoveryNetworkIdle(page, snapshot.discovery);
+    yield page.evaluate(snapshot.execute?.beforeResize);
```
wwilsman, line 351: Yay or nay: should we maybe add `execute.beforeMobile` to replace this part? Technically the width is staying the same so can it count as resizing?

packages/core/src/snapshot.js
```diff
@@ -323,6 +324,70 @@ function waitForDiscoveryNetworkIdle(page, options) {
...
+    yield page.resize({
+      mobile: true,
+      width: initialWidth,
+      height: snapshot.minHeight,
+      deviceScaleFactor: snapshot.deviceScaleFactor
+    });
+
+    yield page.reload();
+    yield waitForDiscoveryNetworkIdle(page, snapshot.discovery);
+    yield page.evaluate(snapshot.execute?.afterResize);
```
wwilsman, line 362: Likewise, should this become `execute.afterMobile`?

packages/core/src/snapshot.js
```diff
@@ -323,6 +324,70 @@ function waitForDiscoveryNetworkIdle(page, options) {
...
+      yield page.evaluate(snapshot.execute?.beforeResize);
+
+      yield page.resize({
+        width,
+        mobile: true,
+        height: snapshot.minHeight,
+        deviceScaleFactor: snapshot.deviceScaleFactor
+      });
+
+      yield page.reload();
```
wwilsman, line 384: Regardless if we decide to add mobile execs, we should add an `after` after this reload.

packages/core/src/snapshot.js
```diff
@@ -323,6 +324,70 @@ function waitForDiscoveryNetworkIdle(page, options) {
...
+    } else {
+      yield waitForDiscoveryNetworkIdle(page, snapshot.discovery);
```
wwilsman, line 386: Is this actually necessary? We already wait for idle at the top of the next loop, or after this whole function is called. So this line will essentially double idle time after resizing without mobile.

packages/core/src/page.js
```diff
@@ -35,13 +35,18 @@ export class Page {
...
+  async reload() {
+    await this.session.send('Page.reload');
+    this.log.debug('Page reloaded', this.meta);
+  }
+
   // Resize the page to the specified width and height
-  async resize({ width, height }) {
-    this.log.debug(`Resize page to ${width}x${height}`);
+  async resize({ width, height, deviceScaleFactor = 1, mobile = false }) {
+    this.log.debug(`Resize page to ${width}x${height} at ${deviceScaleFactor}x scale`);
```
wwilsman, line 45: What do you think about

```suggestion
    this.log.debug(`Resize page to ${width}x${height} @${deviceScaleFactor}x`);
```

So it would read `Resize page to 600x1080 @2x` for example

wwilsman (changes requested):

packages/core/src/config.js
Robdel12: Given that, I'll remove the default. I also never used it further down like I was planning on so it's not bad to remove

packages/core/src/page.js
Robdel12: I like it better than what I have

packages/core/src/snapshot.js
Robdel12: Turns out it's not. I was a bit overzealous with this waits because at one point I was chasing flakes 😅

wwilsman (approved):
Looks good to me as long as you approve of my minimal changes

PR #976, opened 2022-06-27
✨ Export core server util

wwilsman:
## What is this?

This add a small util to encapsulate creating a new server and exports it with other core utils. This is needed by https://github.com/percy/percy-storybook/pull/592, but could also be useful for future projects.

Some code was adjusted to make use of the new util which makes it covered by existing tests.


Robdel12 (approved):
🏁

PR #977, opened 2022-06-30
🐛 Pass auth headers when requesting assets directly

### LINKED ISSUE #969: 🐛 Pass auth config to node requests for fonts
## What is this?

In #921 we realized font responses were being corrupted through the browsers request interception API. To get around that, we now request and save fonts straight from node as a binary (no encoding). In that PR, I forgot to pass the auth config to the request that's being made, which will prevent sites behind auth from capturing fonts properly. 

To fix this we need to pass the config we already have with the node request so it's properly authenticated. We should also add a debug log for when we're requesting fonts via node.

Robdel12:
## What is this? 

In https://github.com/percy/cli/pull/921 we realized font responses were being corrupted through the browsers request interception API. To get around that, we now request and save fonts straight from node as a binary (no encoding). In that PR, I forgot to pass the auth config to the request that's being made, which will prevent sites behind auth from capturing fonts properly.

This PR adds the request headers the browser has to the node request we made to capture the fonts directly. Also adds a debug log when we are requesting these fonts directly. 

Closes #969

### DIFF
--- a/packages/core/test/discovery.test.js
+++ b/packages/core/test/discovery.test.js
@@ -871,6 +871,58 @@ describe('Discovery', () => {
       ]));
     });
 
+    it('captures fonts with valid auth credentials', async () => {
+      percy.loglevel('debug');
+
+      const fontAuthDOM = dedent`
+        <html>
+        <head>
+          <style>
+           @font-face { font-family: "test"; src: url("font-auth/font.woff") format("woff"); }
+           body { font-family: "test", "sans-serif"; }
+          </style>
+        </head>
+        <body>
+          <p>Hello Percy!<p>
+          ${' '.repeat(1000)}
+        </body>
+        </html>
+      `;
+
+      server.reply('/font-auth/font.woff', ({ headers: { authorization } }) => {
+        if (authorization === 'Basic dGVzdDo=') {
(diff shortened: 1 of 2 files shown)


packages/core/test/discovery.test.js
```diff
@@ -871,6 +871,58 @@ describe('Discovery', () => {
...
+         <style>
+          @font-face { font-family: "test"; src: url("font-auth/font.woff") format("woff"); }
+          body { font-family: "test", "sans-serif"; }
+         </style>
+       </head>
+       <body>
+         <p>Hello Percy!<p>
+         ${' '.repeat(1000)}
+       </body>
+       </html>
```
wwilsman, line 889: Very small nitpick, these all look like they could be indented one more space.

Right now it's indented 7 spaces, when you probably intended it to be an even 8.

Robdel12: Nothing gets past this guy

andersrehn:
@Robdel12 thank you for finding and solving this bug, my project was suffering from this exact problem. Tested @percy/cli 1.6.1 and my woff2 fonts behind basic auth works like a charm again! Great work :)

PR #984, opened 2022-07-05
✨ Add core testing mode to help test SDKs

wwilsman:
## What is this?

Currently, the majority of our SDKs (JavaScript SDKs) have access to sdk-utils' test server that mocks the core API. This is the primary method of testing these SDKs since their implementations use sdk-utils internally anyway. For other SDKs, we implement the required utils on a per-language basis, including how they're tested. Testing typically involves mocking HTTP modules or creating stub APIs to simulate the CLI exec command running.

With minimal changes to our existing core API, we can add a testing mode. This new testing mode accomplishes two main tasks. First is to silence all logs and enable dry-run mode to disable uploads and asset discovery. Second is to enable a few extra API endpoints to allow all SDKs to be more easily tested. New API endpoints include access to raw logs, a simple HTML document to snapshot, and a set of commands to manipulate how the API responds to test various circumstances such as errors, disconnects, or missing/invalid core version information.

For now, this new testing mode will be used with the upcoming .NET SDK. In the future however, we should update all of our other tests across all languages and eventually remove the old test server from sdk-utils.

Other changes made include a small adjustment to client's request util to allow `false` as a JSON body, and updated percy-css tests which were failing for me locally from accidentally unstashed, unrelated, changes.


Robdel12 (approved):
🏁 This will make non-js SDKs muchh easier to build too. Great idea 👏🏼

PR #985, opened 2022-07-05
✨ Allow PascalCase options (and other casing)

wwilsman:
## What is this?

While creating the .NET SDK, I noticed that the convention in that land seems to be PascalCase object properties, and as everything is an object, some people might create PascalCase snapshot options objects. However our options normalization does not handle the first word of keys. So even a key of `ALLOWED_HOSTNAMES` would end up becoming `ALLOWEDHostnames`.

With the changes in this PR, the config normalization helpers will now split all words regardless of casing, then convert to the desired casing. Not only does this enable PascalCase options, but will now handle all-caps underscore or even option names with spaces (like `network-idle timeout` but who would do that?).

### DIFF
--- a/packages/config/src/utils/normalize.js
+++ b/packages/config/src/utils/normalize.js
@@ -7,30 +7,26 @@ const CAMELCASE_MAP = new Map([
   ['javascript', 'JavaScript']
 ]);
 
-// Converts kebab-cased and snake_cased strings to camelCase.
-const KEBAB_SNAKE_REG = /[-_]([^-_]+)/g;
+// Regular expression that matches words from boundaries or consecutive casing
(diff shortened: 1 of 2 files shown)


packages/config/src/utils/normalize.js
```diff
@@ -7,30 +7,26 @@ const CAMELCASE_MAP = new Map([
   ['javascript', 'JavaScript']
 ]);
 
-// Converts kebab-cased and snake_cased strings to camelCase.
-const KEBAB_SNAKE_REG = /[-_]([^-_]+)/g;
+// Regular expression that matches words from boundaries or consecutive casing
+const WORD_REG = /[a-z]{2,}|[A-Z]{2,}|[0-9]{2,}|[^-_\s]+?(?=[A-Z0-9-_\s]|$)/g;
```
Robdel12, line 11: And the regex grows 

oh my

Robdel12 (approved):
🏁

PR #986, opened 2022-07-07
🐛 Pass authorization config to asset discovery direct requests

Robdel12:
## What is this?

Turns out when using the `authorization` config in `.percy.yml` (where you pass a username/password rather than encoding it as a header), those aren't passed as request headers. So simply grabbing the same headers and passing them to the node request won't work here. We need to check if the `authorization` config exists and if it does, apply that header to the Node request.

### DIFF
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -41,7 +41,7 @@ export function createRequestHandler(network, { disableCache, disallowedHostname
   };
 }
 
-export function createRequestFinishedHandler(network, authorization, {
+export function createRequestFinishedHandler(network, {
   enableJavaScript,
   allowedHostnames,
   disableCache,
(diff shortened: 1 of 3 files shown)


packages/core/src/network.js
```diff
@@ -31,7 +31,7 @@ export class Network {
 
     if (this.interceptEnabled) {
       this.onRequest = createRequestHandler(this, options.intercept);
-      this.onRequestFinished = createRequestFinishedHandler(this, options.intercept);
+      this.onRequestFinished = createRequestFinishedHandler(this, options.authorization, options.intercept);
```
wwilsman, line 34: No need to pass in extra arguments here. The instance is already passed as the first argument and also already has an `.authorization` property (see 10 lines above).

packages/core/src/discovery.js
```diff
@@ -84,10 +85,18 @@ export function createRequestFinishedHandler(network, {
...
 
+        // font responses from the browser may not be properly encoded, so request them directly
         if (mimeType?.includes('font')) {
-          // font responses from the browser may not be properly encoded, so request them directly
           log.debug('- Requesting asset directly');
-          body = await makeRequest(response.url, { buffer: true, headers: request.headers });
+          if (!headers.Authorization && authorization?.username) {
+            let token = Buffer.from(authorization.password
+              ? `${authorization.username}:${authorization.password}`
+              : `${authorization.username}:`).toString('base64');
```
wwilsman, line 94: Without the extra argument, these would all be `network.authorization`

wwilsman (changes requested):

packages/core/src/discovery.js
wwilsman: Or alternatively use an array join

```js
let token = Buffer.from([
  network.authorization.username, 
  network.authorization.password || ''
].join(':'))..toString('base64');
```

wwilsman (approved):
🎮

PR #989, opened 2022-07-12
✨ In testing mode, clear logs on reset

wwilsman:
## What is this?

During testing mode, full logs are rarely useful, but individual snapshot logs are more important to test against. Clearing logs on reset will allow SDK tests to read only recent and relevant logs without having to filter them.


Robdel12 (approved):
🏁 Niceee, looks good!

PR #990, opened 2022-07-12
✨ Add option to delay uploads without deferring until the end

wwilsman:
## What is this?

This new `delayUploads` option is an alternative to the existing `deferUploads` option. With the existing `deferUploads` option, all uploads (including build creation) are deferred until the instance is stopped. With the new `delayUploads` option, an upload is not processed until the next upload is queued. So the build is not created until the first snapshot is taken, and that snapshot is not uploaded until the next snapshot is taken, and so on until the last snapshot is uploaded when the instance is stopped and the build is finalized.

Currently, this is only useful to speed up the snapshot and storybook commands since those are the only SDKs that utilize `deferUploads` and will error when no snapshots are found. In the future, this could become an opt-in feature for the exec command for the rest of our SDKs, which could error when missing snapshots or allow retrying queued snapshots.

While it wasn't necessarily needed, I felt compelled to add a mirroring test for `deferUploads` which is just as rigorous as the new test for `delayUploads`.

### DIFF
--- a/packages/cli-snapshot/src/snapshot.js
+++ b/packages/cli-snapshot/src/snapshot.js
@@ -58,7 +58,7 @@ export const snapshot = command('snapshot', {
   ],
 
   percy: {
-    deferUploads: true
+    delayUploads: true
   },
 
   config: {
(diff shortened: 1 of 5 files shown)


Robdel12 (commented):
🏁 Nice, I think this will take care of https://github.com/percy/cli/discussions/906 too

PR #1000, opened 2022-07-26
🐛 Improve compatibility with AMD module systems

wwilsman:
## What is this?

Some older implementations of AMD module loading systems might throw an error when the first argument to define a module is not the name of the module itself. This PR adds the package name to the AMD section of the generated bundle output for rollup builds.


Robdel12 (approved):
🏁

Kraindrie:
Daghehh

dfreeman:
@wwilsman This seems to have broken `@percy/ember`. It uses some [arcane build shenanigans](https://github.com/percy/percy-ember/blob/master/index.js#L15-L18) to include `@percy/sdk-utils`, and the code that produces falls over when a module name is provided to define.

PR #1008, opened 2022-07-27
✨ Use env sha for build:wait when there's a project with no commit

wwilsman:
## What is this?

With the `build:wait` command, a project and commit can be provided to search for a build to wait on. However, sometimes the `HEAD` commit, or even local CI environment variables, can refer to a merge commit instead of the branch commit. When a build is created, we use the correct commit in most environments. However when a user provides a commit to the `build:wait` command, it might not match the commit that was used when creating the build. This causes the API to return an empty list since there are no builds associated with that commit for a project.

This PR addresses this issue in two primary ways. First, the command will log `Waiting for build...` when the API returns an empty list. This is because this command can be started before a build is actually created, and will update with the correct info once a matching build is found within the command timeout. An error will be thrown if a build is not found within the timeout. Second, rather than relying on the user to provide the correct commit, we can default to the commit found in the environment, which is the same commit that would be used when creating a new build. This default commit is only used when providing the `--project` flag without a `--commit` flag.

### DIFF
--- a/packages/cli-build/src/wait.js
+++ b/packages/cli-build/src/wait.js
@@ -12,7 +12,6 @@ export const wait = command('wait', {
   }, {
     name: 'project',
     description: 'Build project slug, requires \'--commit\'',
-    requires: ['commit'],
     type: 'slug',
     short: 'p'
   }, {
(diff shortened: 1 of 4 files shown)


Robdel12 (approved):
🏁

PR #1012, opened 2022-07-28
✨ Improve logger functionality

wwilsman:
## What is this?

This PR addresses two small issues with logger and adds another small feature.

- Some URLs don't get fully colored properly in log output. This is fixed with a more verbose URL regular expression.

- In non TTY environments colors aren't always supported and produce raw ansi output. This is fixed by only using colors for TTY stdout streams.

- Depending on the SDK/integration, it can sometimes be beneficial to hook into or override some logging behavior. This is made possible for our logger by moving message formatting into the write method so we can pass along the full log payloads. With full payloads, one is able to set a custom `logger.instance.write` method to handle log output.

Thanks to @grachov for the ideas with #953 & #961.

### DIFF
--- a/packages/logger/src/logger.js
+++ b/packages/logger/src/logger.js
@@ -1,6 +1,6 @@
 import { colors } from './utils.js';
 
-const URL_REGEXP = /\bhttps?:\/\/[^\s/$.?#].[^\s]*\b/i;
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
🏁

PR #1014, opened 2022-07-28
🐛 Fix cli-snapshot warning for top-level references option

### LINKED ISSUE #898: Reporting "references: unknown property" when references is present in snapshots YAML
## The problem

The documentation at https://docs.percy.io/docs/cli-snapshot clearly provides examples of the `references` array at the root of the snapshots file for use when defining YAML references/anchors. However, the CLI now rejects this property with the error (when previously it worked in :

```
[percy] - references: unknown property
```

## Environment

- Node version: v16.14.2
- `@percy/cli` version: 1.1.0
- Version of Percy SDK you’re using: Not sure this is applicable (???)
- OS version: Windows 10
- Type of shell command-line [interface]: `npx percy snapshot scripts/percy/snapshots-test.yml`

## Details

The console generates the following error output:

```
[percy] Percy has started!
[percy] Invalid snapshot options:
[percy] - references: unknown property
[percy] Snapshot taken: home
[percy] Finalized build #BUILD_NUMBER: https://percy.io/RANDOM_HASH/OUR_SLUG/builds/LONG_NUMBER
```

## Code to reproduce issue

```yaml
base-url: http://example.com

# Reusable references
references:
  waitForArticles: &waitForArticles |
    .js-main-listing .article-wrapper:last-of-type.animate

snapshots:
  - name: "home"
    url: "http://example.com/"
    waitForSelector: *waitForArticles
```

wwilsman:
## What is this?

When using the `percy snapshot` command with a snapshots file, we previously were relaxed about top-level options so YAML files could make use of references and anchors. Since moving some `cli-snapshot` features into `core`, top-level options became more strict as other types of files can now be provided (such as sitemap/static server options).

While references still work, they now inadvertently produce a validation warning (#898). This PR removes this warning for the `references` property specifically by removing that property from the config object before passing it along to the snapshot method. While we previously allowed any name for references, going forward they should be kept to the `references` property, otherwise the validation warning will be logged.

Fixes #898

### DIFF
--- a/packages/cli-snapshot/src/snapshot.js
+++ b/packages/cli-snapshot/src/snapshot.js
@@ -81,8 +81,8 @@ export const snapshot = command('snapshot', {
     if (file) {
       // load snapshots file
       let snapshots = yield loadSnapshotFile(file);
-      // accept a config object instead of an array of snapshots
-      let config = Array.isArray(snapshots) ? { snapshots } : snapshots;
+      // remove any references and accept an array of snapshots instead of an config object
+      let { references, ...config } = Array.isArray(snapshots) ? { snapshots } : snapshots;
       options = merge(config, { baseUrl, include, exclude });
     } else if (serve) {
       // serve and snapshot a static directory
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
🏁

PR #1015, opened 2022-07-28
🐛 Improve aborting during bulk snapshots

wwilsman:
## What is this?

Currently, when taking bulk snapshots by either providing an array of snapshots or a snapshots function, a few aspects of the process cannot be aborted as expected, and instead get aborted after running asynchronously.

This PR adds a few core generator utils to create generators from promises (`yieldTo`) and handle running generators concurrently (`yieldAll`). The `yieldTo` util was also made to handle other values besides promises so `yieldAll` can behave similarly to `Promise.all` while still yielding to each generator.

With the help of these additional utils, bulk snapshots processing can now better handle aborting (by using generators instead of promises). A couple of other places were also adapted with these utils, as well as unit tests were added.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,6 +3,7 @@ extends: standard
 plugins:
   - babel
 rules:
+  one-var: off
   prefer-const: off
   no-extra-parens: off
   no-unused-expressions: off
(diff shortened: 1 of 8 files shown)


Robdel12 (approved):
🏁

PR #1016, opened 2022-07-29
🐛 Use then instead of finally for promises with the yieldTo util

wwilsman:
## What is this?

Some versions of V8 complain about promise rejections if there is no detected catch handler. Even though we await on these promises and allow them to reject asynchronously, the error can still surface as "PromiseRejectionHandledWarning: Promise rejection was handled asynchronously."

This is fixed by using then with a catch handler instead of a single finally. The promise reference no longer needs to be updated since it would cause issues to be swallowed.


Robdel12 (approved):
🏁

PR #1020, opened 2022-08-03
✨ Config schema and validation improvements

wwilsman:
## What is this?

While adjusting the config schema for Storybook, there's a requirement to conditionally validate some properties. Specifically, when JavaScript is enabled, DOM capture options cannot be used. While this can be partially achieved with the current `@percy/config` package, this PR makes some adjustments to not only make it easier, but to also add the ability for more broad config validations.

First, we currently suppress confusing validation warnings from `allOf`, `oneOf`, `anyOf`, and `not`. This is because the default validation warnings are _always_ generated even if that schema is not the best match. For example, if `anyOf` is used to accept both a number or an array of numbers, the default warnings would show at least one failure since both cannot be true, but the whole schema is valid since only one needs to be true. For conditional properties like `if`, `then`, and `else`, similar default warnings are generated unless they are suppressed.

Second, Storybook capture options could not be disallowed in a `.percy.yml` config file when `snapshot.enableJavaScript` is true. This is because the current way to manipulate the config file schema is by adding individual object properties with no method available for the properties to introspect each other. To help with this, this PR adds a new `$config` keyword used when registering a config file. This keyword can be used to manipulate the entire config file schema without restriction. This should be used sparingly since it has the potential to remove or replace the config file schema completely.

With these two changes to config schemas, the Storybook SDK will be able to specify a schema in which capture options are only accepted as long as JavaScript is not enabled.

### DIFF
--- a/packages/config/src/validate.js
+++ b/packages/config/src/validate.js
@@ -152,7 +152,7 @@ function a(word) {
 }
 
 // Default errors anywhere within these keywords can be confusing
-const HIDE_NESTED_KEYWORDS = ['oneOf', 'anyOf', 'allOf', 'not'];
+const HIDE_NESTED_KEYWORDS = ['oneOf', 'anyOf', 'allOf', 'if', 'then', 'else', 'not'];
 
 function shouldHideError(key, path, error) {
   let { parentSchema, keyword, schemaPath } = error;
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
🏁

PR #1028, opened 2022-08-09
🐛 Update snapshot typedefs

Robdel12:
## What is this?

We were missing `devicePixelRatio` & `scope` in the TS typedefs. This will close https://github.com/percy/percy-playwright/issues/115


wwilsman (approved):
🎮

PR #1031, opened 2022-08-12
✨ Improve testing mode and support in sdk-utils based tests

wwilsman:
## What is this?

Currently, all SDKs that use `@percy/sdk-utils` utilize its testing utils, including a small testing server that emulates some CLI API features to make the SDKs think they are interacting with the real CLI. When the .NET SDK was implemented, it used a new CLI feature made specifically to help test SDKs: testing mode (#984).

This PR adds a few more features to testing mode: 

- Support for sabotaging remote logging connections
- Tracking API requests and allowing access to tracked requests
- Ability to mimic a failed build for specific error handling

These additional features then allow `@percy/sdk-utils` test helpers to work with testing mode so we can remove the CLI API emulation script from our SDK tests and test against the CLI directly. The rest of this PR replaces all `@percy/sdk-utils` test helpers with new helpers that allow working with testing mode from consuming SDKs:

- `await helpers.setupTest()` - Replaces `helpers.setup()`, and now automatically resets testing mode.
- `await helpers.test(cmd[, arg])` - Will make a POST request to the `/test/api/:cmd` endpoint with any provided second argument as the request body.
- `await helpers.get(what[, map])` - Will make a GET request from the `/test/{logs|requests}` endpoint and optionally map the resulting response data. By default, `logs` are mapped to their `message` property.
- `helpers.testSnapshotURL` - A getter that returns the URL of a small HTML page at `/test/snapshot`.

### Changes for SDK tests

It's important to note that none of these changes have an effect on SDKs' functionality, but will require changes to be made to SDK tests to use the new `@percy/sdk-utils/test/helpers`. This list serves as a reference while we're making appropriate changes to those SDK test suites.

<details>
<summary><code>helpers.setup()</code> & <code>helpers.teardown()</code></summary><br>

The `helpers.setup()` function is now async and named `helpers.setupTest()`. The `helpers.teardown()` function is no longer required and was removed completely.
<hr></details>

<details>
<summary><code>helpers.mockSite()</code> & <code>helpers.closeSite()</code></summary><br>

These helpers were used to stand up and start a mock site for taking snapshots of in respective SDKs. The new CLI testing mode has a `/test/snapshot` endpoint that serves a basic HTML page for this purpose. If an SDK needs a more advance snapshot to capture for a specific reason we can circle back to this new endpoint and allow it to be customized using another `/test/api/:cmd` command.
<hr></details>

<details>
<summary><code>helpers.logger.*</code></summary><br>

[...]

PR #1034, opened 2022-08-15
🐛 Do not implicitly reset testing mode remote connections

wwilsman:
## What is this?

With #1031, testing mode was updated to support disconnecting remote logging connections. This was done mostly for `@percy/sdk-utils` tests themselves because in practice, SDK tests should only need to connect remotely once.

Currently having remote connections disconnect and reconnect between each test causes unnecessary load and can sometimes even cause SDK tests to hang. By avoiding this implicit disconnect we can avoid SDK test flakes/hangs. 

Connections can still be explicitly disconnected with the `remote-logging` command, which enables `@percy/sdk-utils` tests to continue to exist as-is without changes.


Robdel12 (approved):
🏁

PR #1038, opened 2022-08-17
✨Add core option to skip asset discovery

wwilsman:
## What is this?

While asset discovery is usually necessary for taking DOM snapshots, sometimes we don't actually need to take DOM snapshots at all, such as with the `upload` command. For these circumstances, downloading a browser for asset discovery that won't ever be used is unnecessary.

For the `upload` command, we currently have a 2-part approach to skipping asset discovery. First, we utilize a `@percy/cli-command` specific option, `discoveryFlags`, to suppress asset discovery flags from being accepted. Second, we provide the `percy.start()` method with an option to skip launching a browser, which in-turn prevents the browser download completely.

For future projects, this 2-part solution isn't possible, and even now seems like a workaround for a missing feature. This PR adds that missing feature: an option to skip asset discovery altogether with a new `skipDiscovery` core option.

The new `skipDiscovery` option actually works nicely with the existing `dryRun` option. The `dryRun` option already skips asset discovery along with disabling uploads. With the new `skipDiscovery` option, some logic branches were changed from checking for `dryRun`. Now, besides a few specific places for logging, `dryRun` is equivalent to `skipDiscovery && skipUploads`.

Extra logic was added to prevent snapshot captures when asset discovery is disabled since capturing the DOM requires a browser which `skipDiscovery` specifically prevents from being launched. If any normal DOM snapshots are received, they simply bypass asset discovery completely.

This option is not currently exposed publicly in any way. Only the `upload` command and specific future SDKs will utilize this feature internally.

### DIFF
--- a/packages/cli-command/README.md
+++ b/packages/cli-command/README.md
@@ -171,9 +171,6 @@ option, if the environment variable `PERCY_ENABLE` is `0`, the callback _will no
 instance (and can act accordingly).
 
 - `percy` — Enables creation of a `@percy/core` instance initialized with provided options.
-  - `discoveryFlags` — When `false` prevents percy discovery flags from being accepted, but will
-    still allow other percy flags such as `--config` and `--dry-run`.
-  - `...` — All other options are passed along to the new percy instance.
 
 #### Config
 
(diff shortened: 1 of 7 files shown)


Robdel12 (approved):
🏁

PR #1039, opened 2022-08-17
✨ Format logger messages padded by newlines

wwilsman:
## What is this?

When adding a label to a log message, the label should be added after any leading newlines. Likewise, debug timestamps should be added before any trailing newlines. This is currently useful for the version check log, however will also be utilized for a future planned feature.


Robdel12 (approved):
🏁

PR #1040, opened 2022-08-17
✨ Add CLI command framework features

wwilsman:
## What is this?

This PR adds several additional features to the `@percy/cli-command` package:

- Built-in `--port` flag when `percy.server: true` is set for a command. This allows commands like `exec`, `exec:start`, and others to omit the addition of a port option that is used internally by core.

- Export built-in command flags for reuse in other commands. Commands such as `exec:ping` and `exec:start` that utilize the flags like `--port`, but do not actually start their own core server.

- Hide deeply nested subcommands from help output. While not currently used, future commands might have deeply nested commands that contain more than a single command separator (`:`). These long commands can push the help descriptions even further right. Help output for nested commands themselves will still be visible as usual.

- Allow hiding commands and optionally logging warnings for them. Again, not currently useful, but future commands might need to be hidden from help output while in development. Also, if found anyway, these hidden commands might want to log a warning about why they are hidden to begin with.


Robdel12 (approved):
🏁

Issue #1021, opened 2022-08-04
Pass PERCY_BUILD env var to subprocess


fotinakis:
It would be helpful to know the percy build_id in subprocesses.

There are only a few ways of passing this information, either the subprocess has to fetch it (harder), or it can be passed in (easier). Since build creation is blocking before the subprocess is started, we can go with the easier option and add a `PERCY_BUILD` env var to the current environment before creating the subprocess, so that it inherits it and has access to it.

PR #1041, opened 2022-08-17
✨ Expose current build ID

wwilsman:
## What is this?

This PR introduces two methods of exposing the build ID of the currently running Percy build. It currently relies on #1040 for sharing the server port between all command that need it.

For the `exec` command, an additional environment variable was added to the current provided environment variables. The variable, `PERCY_BUILD_ID`, will reflect the `percy.build.id` value if present. This solves #1021, although using a different environment variable name.

For other methods, the healthcheck endpoint already responds with a build ID if available. However for some languages, it is easier to run another process rather than request specific data from an endpoint. For such scenarios, a new `build:id` command was added to the build namespace. This command mirrors the `exec:ping` command, however upon a successful connection will write the build ID to stdout without any additional logs.

While adding tests for `build:id`, which is heavily influenced by `exec:ping`, I noticed that all of `cli-exec`, `cli-snapshot`, and `cli-upload` tests imported their test subjects from the package directly. In contrast, all of `cli-build` and `cli-config` tests imported subjects from source files. Changing these to point to the package itself allows us to test that exports are correct, and by extension I found that I wasn't exporting the new `build:id` command properly.


Robdel12 (approved):
🏁

PR #1042, opened 2022-08-18
🐛 Fix static server base-url issue

wwilsman:
## What is this?

The base-url option for static servers wasn't working correctly for two reasons. First, is the assumed sitemap URL was being constructed in a way that would leave off trailing paths from URLs, like how relative links work. Second, when creating a static server, the generated sitemap was only being served at the server root, not at the base-url root.

The first part of the issue is fixed by ensuring that any provided base-url has a trailing forward slash so when the URL is constructed, it treats the base-url as the path root. The second part is fixed by adjusting static server creation to host the generated sitemap at the correct path under the base-url.

### DIFF
--- a/packages/core/src/api.js
+++ b/packages/core/src/api.js
@@ -160,22 +160,25 @@ export function createPercyServer(percy, port) {
 
 // Create a static server instance with an automatic sitemap
 export function createStaticServer(options) {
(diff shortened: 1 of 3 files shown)


Robdel12 (approved):
🏁

PR #1043, opened 2022-08-19
✨ Add hidden CLI app command for future SDKs

wwilsman:
## What is this?

We are currently working on support for native app screenshots, in which the browser download normally done by the `exec` command is wholly unnecessary. With the new `skipDiscovery` option introduced with #1038, we could map a flag or environment variable to enable this option, however there might still be future difference we wish to separate by means of an entirely new CLI command namespace.

This PR introduces that new namespace, `percy app`. Currently all subcommands are very thin wrappers for the `exec` commands which enables `skipDiscovery` while changing the commands help output description slightly. The plan is for native apps to use `percy app:exec` instead of `percy exec` (and similarly with `app:exec:start`).

Currently, this PR uses a `hidden` command warning since it isn't ready to be used by any SDKs (there currently isn't any native app SDKs to support). Once we ship an SDK that utilizes these new commands, we can release a separate CLI update to unhide these commands and remove the warning.

### DIFF
--- a/packages/cli-app/LICENSE
+++ b/packages/cli-app/LICENSE
@@ -1,0 +1,18 @@
+Copyright © Perceptual Inc.
+
+The MIT License (MIT)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
+associated documentation files (the "Software"), to deal in the Software without restriction,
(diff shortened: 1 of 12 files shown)


Robdel12 (approved):
🏁

PR #1051, opened 2022-09-01
🐛 Serialize all CSSOM rules

wwilsman:
## What is this?

Previously, we excluded serializing CSSOM rules that were already seemingly included in the document. However, JavaScript on the page can still manipulate this CSSOM without updating the document, which prevents some CSSOM changes from being properly serialized. By removing this exclusion, existing CSSOM style content will also be serialized, which will include any manipulated CSSOM.

I removed two tests as a result of this change. One test regarding a whitespace check since the check is no longer performed. The other being a test to ensure an existing stylesheet wasn't serialized. It is now serialized on purpose, resulting in this test being counter to what we want. We still don't want to serialize external stylesheets, which is now caught by missing coverage, so a `link` element was added to the test document to test that it isn't serialized.

Finally, I changed how we compose our CSSOM style content from being a single long string to now being broken up by newlines. This makes debugging serialized CSS in DOM snapshot a bit easier.

### DIFF
--- a/packages/dom/src/serialize-cssom.js
+++ b/packages/dom/src/serialize-cssom.js
@@ -1,8 +1,7 @@
 // Returns true if a stylesheet is a CSSOM-based stylesheet.
 function isCSSOM(styleSheet) {
-  // no href, has a rulesheet, and isn't already in the DOM
-  return !styleSheet.href && styleSheet.cssRules &&
-    !styleSheet.ownerNode?.innerText?.trim().length;
+  // no href, has a rulesheet, and has an owner node
+  return !styleSheet.href && styleSheet.cssRules && styleSheet.ownerNode;
 }
 
 // Outputs in-memory CSSOM into their respective DOM nodes.
@@ -14,11 +13,13 @@ export function serializeCSSOM(dom, clone) {
       let cloneOwnerNode = clone.querySelector(`[data-percy-element-id="${styleId}"]`);
 
       style.type = 'text/css';
+      style.setAttribute('data-percy-element-id', styleId);
       style.setAttribute('data-percy-cssom-serialized', 'true');
       style.innerHTML = Array.from(styleSheet.cssRules)
-        .reduce((prev, cssRule) => prev + cssRule.cssText, '');
+        .map(cssRule => cssRule.cssText).join('\n');
 
       cloneOwnerNode.parentNode.insertBefore(style, cloneOwnerNode.nextSibling);
+      cloneOwnerNode.remove();
     }
   }
 }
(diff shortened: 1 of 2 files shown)


Robdel12 (approved):
🏁 niceee

esetnik:
@wwilsman I am noticing a change in the rendering of scrollbars after the v1.10.1 update that I believe might be related to this PR. Scrollbars appear fat and loose their rounded appearance. This does not happen on v1.10.0 or any previous version. Scrollbars appear as before in a real browser.

v1.10.0 (before)
Screen Shot 2022-09-06 at 3 06 35 PM

v1.10.1 (after)
Screen Shot 2022-09-06 at 3 06 38 PM

wwilsman:
Hey @esetnik! Could you provide the CSS you're using to style those scrollbars? Is this standard CSS accepted by all browsers or is this browser specific CSS that is accidentally being scrubbed?

esetnik:
@wwilsman I am using styled-components. The incorrect scrollbar presentation only appears on screenshots generated for chrome, firefox and safari are correct.

```ts
export const Scrollbar = css`
  div#root ::-webkit-scrollbar {
    all: revert;
    width: 14px;
  }

  /* 4px scroll thumb. 10px margin from end margin */
  div#root ::-webkit-scrollbar-track {
    all: revert;
    box-shadow: inset 0 0 14px 14px transparent;
    border: solid 4px transparent;
  }

  div#root ::-webkit-scrollbar-thumb {
    all: revert;
    box-shadow: inset 0 0 14px 14px #bbbbbe;
    border: solid 4px transparent;
    border-radius: 14px;
  }

  div#root ::-webkit-scrollbar-button {
    all: revert;
    display: none;
  }

  * {
    scrollbar-color: ${({ theme }) => theme.grey60} transparent;
  }
`;
```

amysahli:
@wwilsman We also may be seeing some issues that could be related.  We are seeing issues with a drop down rendering incorrectly (Percy ci only).  Everything looks good in a real browser.  This was alright in v1.10.0.
Before:
Screen Shot 2022-09-07 at 10 29 53 AM
Afte
Screen Shot 2022-09-07 at 10 29 46 AM
r:

esetnik:
@wwilsman would it be helpful if we open a separate bug report?

wwilsman:
@esetnik no need for another issue, I'll be shipping something that might help with this today

Robdel12:
👋🏼 @esetnik @amysahli can you give v1.10.2 a shot to see if that helps? We shipped a change in #1056 that _should_ fix this 🤞🏼

wwilsman:
@esetnik & @amysahli give the latest version `1.10.2` a try. It adds a bit more restrictions around when we serialize CSSOM, so hopefully it will no longer serialize those unnecessarily for you 🤞

amysahli:
@wwilsman Fixed!  Thanks so much! ❤️

PR #1055, opened 2022-09-06
🐛 Use a file URL when importing snapshot files

wwilsman:
## What is this?

When using the `percy snapshot` command and importing a snapshot file on Windows, the drive letter can sometimes get interpreted as a URL protocol, which causes the import to fail. By using an explicit file URL with a `file://` protocol, the import will work in all environments. This PR updates the snapshot file import to use `URL.pathToFileURL()` which will also automatically resolve the absolute path like we were doing before.

This was likely missed by our Windows CI due to mocked ES modules and the mocked filesystem. Unfortunately there isn't much to do to test for this without undoing the mocks and writing a real file when running tests in Windows.

### DIFF
--- a/packages/cli-snapshot/src/snapshot.js
+++ b/packages/cli-snapshot/src/snapshot.js
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import path from 'path';
+import { pathToFileURL } from 'url';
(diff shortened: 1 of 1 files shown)


Robdel12 (approved):
🏁 Siccc

PR #1056, opened 2022-09-07
🐛 Restrict CSSOM serialization to modified stylesheets

wwilsman:
## What is this?

With #1051, CSSOM serialization was made to be much more relaxed, serializing all available document stylesheets. This seems to be problematic in some instances with non-standard vendor prefixed CSS rules.

This PR attempts to avoid serializing these stylesheets while preserving the original goal of #1051, to serialized modified stylesheets. To do this, a check is added to compare the cloned stylesheet with the original stylesheet, and only serialize those stylesheets that do not already match.

Tests were updated to include a modified stylesheet and expectations changed to expect the unmodified stylesheet to not be serialized.


Robdel12 (approved):
🏁

PR #1057, opened 2022-09-07
🐛 Fix sdk-utils websocket constructor hanging

wwilsman:
## What is this?

It seems that the WebSocket constructor can sometimes cause browsers to hang during the initial handshake. This PR adjusts `@percy/sdk-utils` remote logger connection to start the timeout immediately, before a WebSocket is even created. This should cause the resulting promise to be rejected if the WebSocket constructor hangs for more than the default 1s timeout. This timeout is not visible outside of debug logs since remote logging connections aren't required for SDKs to work with the CLI. Unfortunately, a test for this cannot be added without blocking the event loop.


Robdel12 (approved):
🏁 Gotta love those whacky browser bugs

PR #1075, opened 2022-09-19
🗑 Remove deprecated pre-release options

wwilsman:
## What is this?

This PR removes options were meant to be removed with 1.0.0, but snuck past and stuck around a bit long.

The core snapshot list migration was left empty but will be reused for a future deprecation.


Robdel12 (approved):
🏁 ✂️

PR #1076, opened 2022-09-19
♻️ Move devicePixelRatio snapshot discovery option

wwilsman:
## What is this?

The `devicePixelRatio` option is only used in asset discovery and was meant to be nested with related snapshot asset discovery options. This PR corrects the location of this option and adds a new deprecation notice for the old location.


Robdel12 (approved):
🏁

PR #1077, opened 2022-09-19
♻️ Refactor core queues

wwilsman:
## What is this?

This PR refactors the queues in `@percy/core` to allow greater encapsulation and functionality. Queue specific tasks were grouped together and as a result caused other parts of core to be slightly refactored.

### new Queue

The queue class previously allowed arbitrary task functions to be queued together and run concurrently. All discovery related queued tasks did the same thing, but with different snapshots. Likewise with the upload queue, with the exception of deferred build creation.

The new queue class allows queuing arbitrary _items_ and processes them concurrently using configured handlers. The discovery queue is configured to first launch a browser before processing items (discovering resources). The snapshots queue is configured to create and finalize a build around processing items (uploading snapshots). These new queues also allow configuring push handlers which will (in the future) let use merge queued items before being processed.

### Core changes

Almost everything that dealt with the queue had to be updated to accommodate new queue features, including a few logs. Build handling was moved from the core class to the snapshot queue, as was browser launching to the discovery queue. Since the logical place for the new snapshot queue code to live would be in `snapshot.js`, it also ended up next to the new discovery queue, which meant that this file grew pretty exponentially. To break this file up, the discovery queue and related functions were moved to `discovery.js`. The existing network intercept code from that file was then moved and integrated directly into `network.js`.

### Discovery changes

We previously used a combination of two helper functions to create a new browser page, discover snapshot resources for widths and pixel ratios, and then optionally capture DOM snapshots. Page creation was removed from these helpers to be handled directly by the queue, and the remaining code was merged into a single helper. After triggering asset requests from resizing the page, the DOM is optionally captured before additional resources are discovery for other pixel ratios. In the future, we can easily add additional DOM capture points during page resizing.

### Upload command changes

The upload command used a private method, `_scheduleUpload()` to push a function into the internal upload queue which would upload snapshot based on images from the filesystem. Without this private method, the upload method would no longer function properly or be able to dry-run quickly. To solve this, a new `upload()` method was added to core specifically for the upload command. It utilizes additional queued item arguments to provide a processing function which is yielded to during task processing, before the snapshot is uploaded. In the future, this method may adapt to other APIs which can utilize the snapshot queue.

### Other changes

A couple internal methods were also changed slightly in response to some of the refactoring. First, the browser page method `snapshot()` was updated to return the provided payload minus any capture specific options. The returned payload previously only included `dom`, but now includes `domSnapshot` to match what is expected by other helper functions. The core class `setConfig()` method is now simply `set()` (inspired by queue's matching method) which better reflects it's ability to set more than only configuration options (currently also client and env info).

### DIFF
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -1,136 +1,273 @@
-import mime from 'mime-types';
 import logger from '@percy/logger';
-import { request as makeRequest } from '@percy/client/utils';
-import { normalizeURL, hostnameMatches, createResource } from './utils.js';
+import Queue from './queue.js';
+import {
+  normalizeURL,
+  hostnameMatches,
+  createRootResource,
+  createPercyCSSResource,
+  createLogResource,
+  yieldAll
+} from './utils.js';
 
-const MAX_RESOURCE_SIZE = 15 * (1024 ** 2); // 15MB
-const ALLOWED_STATUSES = [200, 201, 301, 302, 304, 307, 308];
-const ALLOWED_RESOURCES = ['Document', 'Stylesheet', 'Image', 'Media', 'Font', 'Other'];
+// Logs verbose debug logs detailing various snapshot options. When `showInfo` is true, specific
+// messages will be logged as info logs rather than debug logs.
+function debugSnapshotOptions(snapshot, showInfo) {
+  let log = logger('core:snapshot');
 
-export function createRequestHandler(network, { disableCache, disallowedHostnames, getResource }) {
-  let log = logger('core:discovery');
+  // log snapshot info
+  log.debug('---------', snapshot.meta);
+  if (showInfo) log.info(`Snapshot found: ${snapshot.name}`, snapshot.meta);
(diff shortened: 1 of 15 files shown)


packages/core/src/discovery.js
```diff
@@ -1,136 +1,273 @@
...
+      capture(processSnapshotResources(captured));
     }
-  };
+  }
+
+  // recursively trigger resource requests for any alternate device pixel ratio
+  if (deviceScaleFactor !== discovery.devicePixelRatio) {
+    yield waitForDiscoveryNetworkIdle(page, discovery);
+
+    yield* captureSnapshotResources(page, snapshot, {
```
Robdel12, line 161: Love this

Robdel12 (approved):
🏁 🔥 Thanks for queuing this up

esetnik:
@wwilsman can you release this change? We are seeing missing snapshots in ~5% of all CI builds using identical source code and cypress test configuration when using https://github.com/percy/percy-cypress and I'm wondering if the changes introduced in this PR will solve the issue.

PR #1078, opened 2022-09-20
🐛 Add sdk-utils logger mocking test helpers

wwilsman:
## What is this?

With #254 and the removal of remote logging, SDK tests will need to be updated to mock local logs. Rather than needing to replicate mocking code for each SDK, we can simply replace the internal log function with our own mock. This mock cannot be tested like a normal mock or spy (no calls or argument tracking). Instead, this mock very minimally mimics the original logging code, without colors, and pushes logs into appropriate arrays for testing.

I moved logger test setup code back into the logger suite so the other tests could use the new mock helper. This revealed a few lines of missing coverage that required another log be added to the two existing tests. Log history tracking was also removed since it was only used for sending to remote connections, which then prompted the linter to start failing because the `const log` variable was not being used or exported. Rather than add an ignore statement, I moved the two `logger` static methods to `Object.assign` them.

### DIFF
--- a/packages/sdk-utils/src/logger.js
+++ b/packages/sdk-utils/src/logger.js
@@ -17,11 +17,6 @@ const loglevel = logger.loglevel = lvl => {
 const log = logger.log = (ns, lvl, msg, meta) => {
   let err = typeof msg !== 'string' && (lvl === 'error' || lvl === 'debug');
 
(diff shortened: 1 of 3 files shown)


Robdel12 (approved):
🏁

PR #1085, opened 2022-09-26
➖ Remove `ws` dependency

Robdel12:
## What is this?

With the removal of the remote logger, we no longer need the `ws` dependency. This was missed by me in the original PR removing the logger.


Robdel12:
Monday brain fart 🤦🏼 WS is still used by CDP

PR #1086, opened 2022-09-26
♻️ Adjust how direct snapshot uploads are handled

wwilsman:
## What is this?

This PR continues some of the refactor work started with #1077. In that PR, the upload command needed a new upload method, so a quick one was implemented. This PR polishes the upload command and moves some related logs around within the queues. This further isolates queue features from each other and also sets the stage for the core upload method and snapshot queue to accept alternate API requests (such as a future comparisons API).


Robdel12 (approved):
🏁

PR #1087, opened 2022-09-27
✨ Add comparisons API for future SDK usage

wwilsman:
## What is this?

Soon, the Percy API will be able to accept specific snapshot comparisons to upload and associate with a build. This PR adds methods to `@percy/client` and `@percy/core` to enable SDKs to communicate with this upcoming API.

### Client

Comparison methods added to client mirror other snapshot methods already present, with a few changes. For example, comparison creation and finalization is similar to snapshot creation and finalization, but with different endpoints and payloads. Likewise for snapshot resources and comparison tiles. The only major difference between the two APIs, is the comparisons flow includes creating a parent snapshot first, but not finalizing it.

Other small changes were made such as repurposing the build ID validation function to validate any type of ID, and using `fs.promises` to read resources and tiles from the filesystem.

### Core queue

With recent work refactoring and isolating the internal queue responsibilities, adding comparison uploads to the existing queue was actually a very small change. When a snapshot payload includes a `tag` it is assumed to be a comparison and `client.sendComparison()` is used in place of `client.sendSnapshot()`

### Core API

Also with the recent refactor work came the addition of the core `percy.upload()` method used by the upload command. Thanks to this method, a new `/percy/comparisons` endpoint can utilize it to facilitate comparison uploads. Now when the method detects the presence of a `tag` or `tiles` property, it will validate all options against a newly added comparison schema, and throw an error when missing the required snapshot name or comparison tag name. This new endpoint also returns a generated link to the comparison redirect API which will redirect to the comparison's web-url when available. 

### SDK utils

A new comparison posting utility function was added to mirror the snapshot posting utility function. These two functions, including their tests, are nearly identical with the only differences being the endpoint used by the function and payload used for the tests. Any JavaScript SDKs that would use the new comparison API should use this new utility function.

### DIFF
--- a/packages/client/README.md
+++ b/packages/client/README.md
@@ -7,6 +7,7 @@ builds. Can also be used to query for a project's builds using a read access tok
 - [Usage](#usage)
 - [Create a build](#create-a-build)
 - [Create, upload, and finalize snapshots](#create-upload-and-finalize-snapshots)
+- [Create, upload, and finalize comparisons](#create-upload-and-finalize-comparisons)
 - [Finalize a build](#finalize-a-build)
 - [Query for a build*](#query-for-a-build)
 - [Query for a project's builds*](#query-for-a-projects-builds)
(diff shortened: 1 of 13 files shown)


Robdel12 (approved):
🏁

Robdel12 (approved):
🏁

PR #1092, opened 2022-10-03
✨ Add snake_case normalization to config utils

wwilsman:
## What is this?

Our `normalize` config utility allows transforming object keys to a different casing while removing empty values from the provided object. Currently, it supports transforming object keys from any casing into either camelCase or kebab-case. This PR adds the ability to transform keys into snake_case by copying the kebab-case functionality and replacing `-` with `_`.


Robdel12 (approved):
🏁

PR #1095, opened 2022-10-05
✨ Expose deferred uploads

wwilsman:
## What is this?

For some time, our internal queues have been able to be deferred so snapshots are only processed and uploaded after asset discovery has completely finished for all snapshots. This was only ever utilized in a couple of SDKs which specifically did not create empty builds when no snapshots were found (storybook and the upload command). With the recent refactor around our internal queues, deferred items can now be cancelled or merged with existing deferred items to allow better encapsulation and greater control over internal queues.

This PR not only exposes the `deferUploads` option via a new config file entry, but also exposes a new core endpoint to flush deferred snapshots. The core flush method was also updated to accept one or more specific snapshots to flush rather than flush all snapshots. Finally, an sdk-utils helper was added which posts to this new core endpoint and can accept one or more snapshot names rather than a snapshot options object.

### DIFF
--- a/packages/core/src/api.js
+++ b/packages/core/src/api.js
@@ -86,6 +86,10 @@ export function createPercyServer(percy, port) {
       if (!req.url.searchParams.has('async')) await snapshot;
       return res.json(200, { success: true });
     })
+  // flushes one or more snapshots from the internal queue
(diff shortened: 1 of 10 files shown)


Robdel12 (approved):
🏁

PR #1097, opened 2022-10-06
🐛 Fix multi-dom capture when using the snapshot command

wwilsman:
## What is this?

While testing multi-dom support with the snapshot command, I noticed two issues that this PR fixes.

1. When both `delayUploads` and `deferUploads` are `true`, the snapshot queue acts delayed but not deferred. This is fixed by checking for `!deferUploads` before automatically flushing the queue.

2. When the first snapshot is taken at the initial width, it is missing it's singular `width`. This happened because `width` was undefined for the first loop, and fixed by setting it to the first width.

One test was updated to test for both of these issues. First `delayUploads` was added to ensure it does not affect `deferUploads`. Second, debug logging was added to test for a specific snapshot log to verify each singular width.


Robdel12 (approved):
🏁

PR #1105, opened 2022-10-19
✨ Show network logs on navigation errors

Robdel12:
## What is this?

Sometimes asset discovery navigation to a snapshot fails due to hanging requests or the load event never firing on the page that's being loaded. In order to help make debugging easier, we're now logging network requests that are hanging/still active when trying to navigate.

We also now suppress not needed error messages from page clean up from in-flight requests. This is most commonly logged when asset discovery is closing and external requests haven't completed.


wwilsman (approved):
🎮

PR #1115, opened 2022-11-01
✨ Expose build URL via env var (`PERCY_BUILD_URL`)

Robdel12:
## What is this?

This exposes the current build URL as an env var to the current process (`PERCY_BUILD_URL`).


wwilsman (approved):
🎮

PR #1122, opened 2022-11-09
✨ Return comparison response in sdk-utils

ninadbstack:
- We need this response in SDKs so that we can parse `link` from it


wwilsman (approved):
The failing test is one of the known windows flakes.

🎮

PR #1127, opened 2022-11-10
:sparkles: Skip github update check if `PERCY_SKIP_UPDATE_CHECK` is truthy

### LINKED ISSUE #1058: Allow option to skip update check
## The problem

`checkForUpdate()` is stalling indefinitely (15min+ before build timeout kicks in) on our newer CI infrastructure for an unknown reason. Our current workaround is to patch the `node_modules/.bin/percy` and remove the `checkForUpdate()` call.

It would be nice to have an option to skip the update check.

## Environment

- Node version: v16.17.0
- `@percy/cli` version: v0.10.2
- Version of Percy SDK you’re using: 
- OS version: Docker `ruby:3.1.1` 

## Details

The `checkForUpdate()` call in `percy` hangs forever.

## Code to reproduce issue

Unsure how to reproduce this at this point.

itsjwala:
**Problem**

When running in CI environments, the cache file would normally be cached along with the project's dependencies. However, for some projects that do not cache dependencies, the GitHub request will always occur when the CLI runs after installation. This can sometimes cause the CLI to hang depending on rate limits or the presence of any locally defined GitHub token.

fixes #1058

**solution**

* we're exposing an environment variable that will skip this update check.

---

**Testing**

* post code changes are done

```sh
yarn lint
yarn build
```

* mock edited `node_modules/@percy/cli/.releases` file with new release `v1.14.0` in cache file(doesn't exist in github releases).

* run cli 
```sh
➜  cli git:(jigar/skip_github_update_check) ✗ export PERCY_LOGLEVEL=debug   
➜  cli git:(jigar/skip_github_update_check) ✗ export PERCY_SKIP_UPDATE_CHECK=true
➜  cli git:(jigar/skip_github_update_check) ✗ npx percy                          
[percy:cli:update] Skipping update check (0ms)
```

### DIFF
--- a/packages/cli/src/update.js
+++ b/packages/cli/src/update.js
@@ -69,7 +69,7 @@ export async function checkForUpdate() {
   let pkg = getPackageJSON(import.meta.url);
   let log = logger('cli:update');
 
-  if (process.env.PERCY_SKIP_UPDATE_CHECK === 'true') {
+  if (process.env.PERCY_SKIP_UPDATE_CHECK) {
     log.debug('Skipping update check');
     return;
   }
--- a/packages/cli/src/update.js
+++ b/packages/cli/src/update.js
@@ -69,6 +69,11 @@ export async function checkForUpdate() {
   let pkg = getPackageJSON(import.meta.url);
   let log = logger('cli:update');
 
+  if (process.env.PERCY_SKIP_UPDATE_CHECK === 'true') {
+    log.debug('Skipping update check');
+    return;
+  }
+
   try {
     // request new release information if needed
     if (!releases) {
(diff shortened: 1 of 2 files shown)


packages/cli/src/update.js
```diff
@@ -69,6 +69,11 @@ export async function checkForUpdate() {
   let pkg = getPackageJSON(import.meta.url);
   let log = logger('cli:update');
 
+  if (process.env.PERCY_SKIP_UPDATE_CHECK === 'true') {
+    log.debug('Skipping update check');
+    return;
+  }
```
samarsault, line 75: Shouldn't we allow any truthy value for this?

itsjwala: hmm, makes sense. so just `if(process.env.PERCY_SKIP_UPDATE_CHECK){ ... }` will work

wwilsman (approved):
Looks good! The failing test is only a windows flake

🎮

PR #1134, opened 2022-11-17
✨  Updated comparisonSchema to set max width to 10000

ninadbstack:
- For comparison tags we have backend validation for <= 10000. Making this consistent

### DIFF
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -373,12 +373,12 @@ export const comparisonSchema = {
         width: {
           type: 'integer',
           minimum: 1,
-          maximum: 10000,
+          maximum: 10000
         },
         height: {
           type: 'integer',
           minimum: 1,
-          maximum: 10000,
+          maximum: 10000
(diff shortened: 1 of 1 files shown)


pankaj443 (approved):

packages/core/src/config.js
```diff
@@ -372,12 +372,13 @@ export const comparisonSchema = {
         osVersion: { type: 'string' },
         width: {
           type: 'integer',
-          maximum: 2000,
-          minimum: 120
+          minimum: 1,
```
itsjwala, line 375: min width 1px is allowed?

ninadbstack: Yes, we have `> 0` validation on API. This is because the App percy does not depend on what is the screensize of the device, we just collect this metadata to show in UI so that users can see it. In case SDK fails to collect this info, a default 1px, 1px width, height is used

itsjwala (approved):

PR #1135, opened 2022-11-18
✨ Accept serialized DOM resources

wwilsman:
## What is this?

This PR expands on how DOM snapshots are accepted to build towards addressing both serialized assets and also any future DOM serialization warnings. Primarily, this PR addresses accepting additional serialized DOM options in core that will enable `@percy/dom` to make serialized assets external rather than embedded in a follow up PR.

## Approach

The config schema for DOM snapshots was updated to allow additional `domSnapshot` options:

- `domSnapshot.html` - The HTML string of the DOM. This is equivalent to providing `domSnapshot` as a string.
- `domSnapshot.warnings` - An array of warning strings to be logged when encountered during snapshot validation.
- `domSnapshot.resources` - An array of resource objects that will be converted to proper snapshot resources:
  - `resource.url` - The complete URL of the resource including the hostname.
  - `resource.mimetype` - The specific mimetype of the resource, typically an image.
  - `resource.content` - The base64 encoded content string of the resource.

When a DOM snapshot is provided to core with serialized resources, these assets will be marked within the resource cache as to always be served and never be overwritten. The resulting snapshot uploaded will include these serialized resources.

As a convenience for any SDKs that may need to stringify JSON responses from the browser, core validation will automatically parse `domSnapshot` options that look like JSON strings.

### DIFF
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -287,8 +287,34 @@ export const snapshotSchema = {
       properties: {
         url: { type: 'string' },
         name: { type: 'string' },
-        domSnapshot: { type: 'string' },
-        width: { $ref: '/config/snapshot#/properties/widths/items' }
+        width: { $ref: '/config/snapshot#/properties/widths/items' },
+        domSnapshot: {
+          oneOf: [{ type: 'string' }, {
+            type: 'object',
+            required: ['html'],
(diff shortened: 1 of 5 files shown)


Robdel12 (approved):
🏁

PR #1136, opened 2022-11-18
✨ Serialize DOM resources

### LINKED ISSUE #425: Create proper assets from serialized canvas elements
## Problem

If there are a large number of canvas elements on a page, the DOM clone can become too large of an asset to upload to the API. This is because the serialized canvas elements are put into the DOM clone as base64 encoded PNGs. 

## Possible solve

In order to avoid this, we should create proper assets from the serialized canvas elements and write them to disk. Then create an asset for Percy and upload it to the API.

wwilsman:
## What is this?

As part 2 of #1135, this PR refactors `@percy/dom` slightly to return the new `domSnapshot` options. To track warnings and resources throughout our serialize functions, a new context object is used to keep references of the DOM, the clone, and new sets for warnings and resources.

To exercise this new serialization return format, the canvas and video serialization functions were adjusted to create new resource objects (tracked by the `resources` context set). The frame serialization function was also updated to recursively merge the resulting frame warnings and resources.

While changing the code to update the src/poster attributes to the fake resource URLs, I noticed that the browser would actively make these requests immediately even though the cloned DOM is not mounted. To workaround this, those attributes are set as `data-percy-serialized-attribute-<name>="<value>"`. After the DOM clone is converted to an HTML string, all occurrences of `data-percy-serialized-attribute-` are removed so the resulting HTML string is left with real attribute names and values.

Finally, as a convenience for any SDKs that may have to serialize the new JSON response, a new serializer-only option was added, `stringifyResponse`. When true, the resulting JSON payload will be stringified before being returned.

This closes #425

### DIFF
--- a/packages/core/test/percy.test.js
+++ b/packages/core/test/percy.test.js
@@ -80,10 +80,11 @@ describe('Percy', () => {
     });
 
     expect(snapshot.url).toEqual('http://localhost:8000/');
-    expect(snapshot.domSnapshot).toEqual(
-      '<!DOCTYPE html><html><head></head><body>' + (
+    expect(snapshot.domSnapshot).toEqual(jasmine.objectContaining({
+      html: '<!DOCTYPE html><html><head></head><body>' + (
         `<p>Hello there, Percy!</p>${img}`
-      ) + '</body></html>');
+      ) + '</body></html>'
+    }));
   });
 
   describe('.start()', () => {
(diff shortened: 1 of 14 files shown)


JithilNCA:
In which version I can find this fix?

samarsault:
1.15.0+

JithilNCA:
I am using the latest one and facing issues while uploading larger images.

JithilNCA:
I am getting 413 error

PR #1153, opened 2022-12-08
✨  Check Type(app/web) Validity while creating builds

hiteshsalavi:
### Enhancement:
Now, that we Percy supports App Percy along with Web Percy, Customers may use tokens of web project with App Percy or vice versa. CLI should be able to provide a understandable error if that's the case.

### Solution
- We get the command context from the `@percy/command` and pass it to Percy.
- Users who use Write token in the CI, for them, an empty failed build will be created in order to get the type from createBuild api.

API PR: percy/percy-api#2941

### DIFF
--- a/packages/cli-command/src/command.js
+++ b/packages/cli-command/src/command.js
@@ -62,8 +62,8 @@ function exit(exitCode, reason = '') {
 // and other common command helpers and properties.
 async function runCommandWithContext(parsed) {
   let { command, flags, args, argv, log } = parsed;
-  // include flags, args, argv, logger, exit helper, and env info
-  let context = { flags, args, argv, log, exit };
+  // include command as parentCommand, flags, args, argv, logger, exit helper, and env info
+  let context = { parentCommand: command, flags, args, argv, log, exit };
   let env = context.env = process.env;
   let pkg = command.packageInformation;
   let def = command.definition;
--- a/packages/cli-command/src/command.js
+++ b/packages/cli-command/src/command.js
@@ -62,8 +62,8 @@ function exit(exitCode, reason = '') {
 // and other common command helpers and properties.
 async function runCommandWithContext(parsed) {
   let { command, flags, args, argv, log } = parsed;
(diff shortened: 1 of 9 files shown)


packages/cli-command/src/command.js
```diff
@@ -62,8 +62,8 @@ function exit(exitCode, reason = '') {
 // and other common command helpers and properties.
 async function runCommandWithContext(parsed) {
   let { command, flags, args, argv, log } = parsed;
-  // include flags, args, argv, logger, exit helper, and env info
-  let context = { flags, args, argv, log, exit };
+  // include command as parentCommand, flags, args, argv, logger, exit helper, and env info
+  let context = { parentCommand: command, flags, args, argv, log, exit };
```
ninadbstack, line 66: you can just call it `command`. no need to rename it to parentCommand. Or if you want to differentiate between user command [argv] and this then call it cliCommand

packages/cli-exec/src/exec.js
```diff
@@ -58,7 +58,8 @@ export const exec = command('exec', {
     log.warn('Percy is disabled');
   } else {
     try {
-      yield* percy.yield.start();
+      const execType = parentCommand.name === 'app:exec' ? 'app' : 'web';
```
ninadbstack, line 61: We use commands as separate plugin modules. Instead of adding this line here. Add it in plugin and just pass `execType` here [ instead of parentCommand ]

packages/core/src/percy.js
```diff
@@ -142,7 +142,12 @@ export class Percy {
   }
 
   // Starts a local API server, a browser process, and internal queues.
-  async *start() {
+  async *start(execType) {
+    this.execType = execType;
```
ninadbstack, line 146: instead of setting this here, you can pass it to `throwIfTypeInvalid` function [ unless you are planning to use it in multiple locations ]

itsjwala:
* @hiteshsalavi please update description, I'll review the PR today.

hiteshsalavi:
With respect to #1157

PR #1164, opened 2023-01-05
✨ skip upload when sha is present

pankaj443:
- update tile config to include sha
- skip uploading step when sha is present for tiles

### DIFF
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -412,20 +412,7 @@ export class PercyClient {
     this.log.debug(`Uploading comparison tile: ${index + 1}/${total} (${comparisonId})...`);
     if (filepath) content = await fs.promises.readFile(filepath);
     if (sha) {
-      let retries = 10;
-      await waitForTimeout(1000);
-      let success = await this.verifyComparisonTile(comparisonId, sha);
-      while (retries > 0 && !success) {
-        retries -= 1;
-        await waitForTimeout(500);
-        success = await this.verifyComparisonTile(comparisonId, sha);
-      }
-
-      if (!success) {
-        this.log.error('Uploading comparison tile failed');
-        return false;
-      }
-      return true;
+      return await this.verify(comparisonId, sha);
     }
 
     return this.post(`comparisons/${comparisonId}/tiles`, {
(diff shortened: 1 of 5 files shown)


prklm10 (approved):

packages/client/src/client.js
```diff
@@ -406,10 +406,18 @@ export class PercyClient {
...
     this.log.debug(`Uploading comparison tile: ${index + 1}/${total} (${comparisonId})...`);
     if (filepath) content = await fs.promises.readFile(filepath);
+    if (sha) {
+      let success = await this.verifyComparisonTile(comparisonId, sha);
+      if (!success) {
+        this.log.error(`Uploading comparison tile for ${sha} failed`);
+        return false;
+      }
+      return true;
+    }
```
ninadbstack, line 420: Ideally would loop in here than on API. API loop will adversely affect API Apdex score

pankaj443: looping here would mean multiple API calls, if we loop in API we will have only 1 api call. Also any changes/fixes can be deployed easily there.

packages/client/src/client.js
```diff
@@ -422,6 +438,20 @@ export class PercyClient {
...
+    validateId('comparison', comparisonId);
+    this.log.debug(`Verifying comparison tile with sha: ${sha}`);
+
+    try {
+      return await this.get(`comparisons/${comparisonId}/tile/verify?sha=${sha}`);
+    } catch (error) {
+      if (error.message.includes('409')) {
+        return false;
+      }
+      throw new Error(error.message);
```
itsjwala, line 451:
```suggestion
      throw error;
```

packages/client/src/client.js
```diff
@@ -406,10 +406,26 @@ export class PercyClient {
...
   }
 
-  async uploadComparisonTile(comparisonId, { index = 0, total = 1, filepath, content } = {}) {
+  async uploadComparisonTile(comparisonId, { index = 0, total = 1, filepath, content, sha } = {}) {
     validateId('comparison', comparisonId);
     this.log.debug(`Uploading comparison tile: ${index + 1}/${total} (${comparisonId})...`);
     if (filepath) content = await fs.promises.readFile(filepath);
+    if (sha) {
+      let retries = 10;
+      setTimeout(() => this.log.debug(`Verifying comparison tile with sha: ${sha}`), 1000);
```
itsjwala, line 415: whats the the reason for `setTimeout` here?

pankaj443: added an initial sleep so that our avg number for retries is low, 1st call would return true n most cases when we provide 1s sleep.

itsjwala: a correct way to sleep would be awaiting on the promise something like [this](https://github.com/percy/cli/blob/54fa2c10eea66c2bdaec3c540d46ecce4b1d5acc/packages/core/src/utils.js#L149-L166)].

packages/client/src/client.js
```diff
@@ -2,7 +2,7 @@ import fs from 'fs';
 import PercyEnv from '@percy/env';
 import { git } from '@percy/env/utils';
 import logger from '@percy/logger';
-import { waitForTimeout } from '../../core/src/utils.js';
+import { waitForTimeout } from '../../core/dist/utils.js';
```
itsjwala, line 5: this is not correct way - you should add `@percy/core` as dependency in `@percy/client` and require it.

pankaj443: got it thanks :)

itsjwala (approved):

packages/client/package.json
```diff
@@ -32,6 +32,6 @@
   },
   "dependencies": {
     "@percy/env": "1.17.0",
-    "@percy/logger": "1.17.0"
+    "@percy/logger": "1.17.0",
```
itsjwala, line 35:
```suggestion
    "@percy/logger": "1.17.0"
```

ninadbstack (approved):

PR #1165, opened 2023-01-05
✨ Shadow DOM serialisation support

itsjwala:
Shadow DOM is a critical part of the newer Web Component standard which is growing in adoption day by day. This shadow DOM is not captured in a Percy snapshot by our existing serialization methods. Hence, this PR adds support for capturing it.

**Constraints**

1) We can’t serialize the DOM if the shadow root is attached in **closed** mode since we don’t have access to the HTML that makes up the tree. **We need the shadow root to be open**.

2) Making it work across all browsers in the infrastructure.

3) Shadow DOM capturing only works if the user’s test browser is `Chromium 90+`

**Implementation details**

Below is summarised lifecycle of `percySnapshot` command (*explained in detail [here](https://docs.percy.io/docs/debugging-sdks#how-percy-works)*)

image

* Currently serialization of Shadow DOM is only possible by [Declarative Shadow DOM](https://web.dev/declarative-shadow-dom/) API provided by the browser.
  * The APIs are currently somewhat stable in chromium, but there is no sign of their adoption in Firefox. (*ref https://caniuse.com/declarative-shadow-dom*)
  * Method we're trying to work with is `getInnerHTML`   (**We've not polyfilled this method for Firefox**)
* Once serialized the `#shadow-root (open)` are converted into `<template shadowRoot="open">` tags
* We can then pass this dom snapshot across the wire in our infrastructure browsers to load the Shadow DOM.
* Though Declarative Shadow DOM and its API are still not supported `Firefox/Safari`, and thus we need to deserialize this Declarative Shadow DOM back to its original form.  
  * We inject a custom polyfill for Declarative Shadow DOM which does this job.

image


gdborton:
@itsjwala any idea when the alpha is gonna hit?  We're starting to look into alternatives to Percy next week due to being blocked on shadow dom.

itsjwala:
@gdborton expect the Alpha release by end of next week - we're in process of deploying initial Infrastructure changes for all of this to work.

packages/dom/test/helpers.js
```diff
@@ -46,10 +38,12 @@ export function withCSSOM(rules = [], prepare = () => {}) {
   for (let rule of [].concat(rules)) {
     $style.sheet.insertRule(rule);
   }
+
+  withShadowCSSOM(rules, prepare);
 }
 
 export function withShadowCSSOM(rules = [], prepare = () => {}) {
-  let $test = getExampleShadowRoot();
+  let $test = document.getElementById('test-shadow').shadowRoot;
```
itsjwala, line 46: lets keep it consistent, have it inside `withCSSOM` and pass `options` which contains `shadow` as one of the option.

packages/dom/test/helpers.js
```diff
@@ -1,5 +1,5 @@
 // create and cleanup testing DOM
-export function withExample(html) {
+export function withExample(html, shadow = true) {
```
itsjwala, line 2: let's pass `options` instead of `shadow` we can deflate these options to bring more options in the future.

packages/dom/test/serialize-dom.test.js
```diff
@@ -68,6 +78,89 @@ describe('serializeDOM', () => {
       expect(html).not.toMatch('<template shadowroot');
       expect(html).not.toMatch('Hey Percy!');
     });
+
+    it('renders single nested ', () => {
+      if (!navigator.userAgent.toLowerCase().includes('chrome')) {
```
itsjwala, line 83:
```suggestion
      if (!navigator.userAgent.toLowerCase().includes('chrome')) {
```

There is a helper `getTestBrowser` that can be used.

packages/dom/test/serialize-dom.test.js
```diff
@@ -68,6 +78,89 @@ describe('serializeDOM', () => {
...
+        '<p>Percy-1</p>',
+        '<div id="Percy-2" .*>',
+        '<template shadowroot="open">',
+        '<p>Percy-2</p>',
+        '</template>'
+      ].join('')));
+    });
+
+    it('renders many nested', () => {
+      if (!navigator.userAgent.toLowerCase().includes('chrome')) {
```
itsjwala, line 108:
```suggestion
      if (!navigator.userAgent.toLowerCase().includes('chrome')) {
```

same comment

packages/dom/test/serialize-dom.test.js
```diff
@@ -68,6 +78,89 @@ describe('serializeDOM', () => {
...
+        return;
+      }
+      withExample('<div id="content"></div>', { withShadow: false });
+      const baseContent = document.querySelector('#content');
+
+      const levels = 5;
```
itsjwala, line 114:
```suggestion
      const levels = 10000;
```

let's use the biggest threshold.

packages/dom/src/serialize-dom.js
```diff
@@ -49,7 +50,7 @@ export function serializeDOM(options) {
   };
 
   ctx.dom = dom;
-  ctx.clone = cloneNodeAndShadow(ctx.dom);
+  ctx.clone = cloneNodeAndShadow(ctx.dom, disableShadowDom);
```
itsjwala, line 53: pass `disableShadowDOM` as part of `ctx`

packages/dom/src/serialize-dom.js
```diff
@@ -38,7 +38,8 @@ export function serializeDOM(options) {
     // allow snake_case or camelCase
     enableJavaScript = options?.enable_javascript,
     domTransformation = options?.dom_transformation,
-    stringifyResponse = options?.stringify_response
+    stringifyResponse = options?.stringify_response,
+    disableShadowDom = options?.disableShadowDom
```
itsjwala, line 42: lets keep it consistent as above.  - `options?.disable_shadow_dom`

packages/dom/test/inject-polyfill.test.js
```diff
@@ -0,0 +1,26 @@
...
+
+describe('injectDeclarativeShadowDOMPolyfill', () => {
+  let dom;
+
+  beforeEach(() => {
+    dom = withExample(`
+        <div data-percy-shadow-host>
+        <template shadowroot="open">
+                <div data-percy-shadow-host>
+                    <template shadowroot="open">
```
itsjwala, line 12: `template` tags won't be loaded as shadow dom, since declarative shadow DOM is interpreted during  DOM parsing.

packages/core/src/page.js
```diff
@@ -140,7 +140,7 @@ export class Page {
     execute,
     ...snapshot
   }) {
-    let { name, width, enableJavaScript } = snapshot;
+    let { name, width, enableJavaScript, disableShadowDom } = snapshot;
```
itsjwala, line 143: * requires docs update https://docs.percy.io/docs/cli-configuration#per-snapshot-configuration which would be common for all the SDKs

c69-addepar:
Hi team, 
we have tried the `1.19.1.alpha.0` and it still does not work:
- about half of our tests fail with `Error: Failed to set the 'adoptedStyleSheets' property on 'ShadowRoot': Sharing constructed stylesheets in multiple documents is not allowed`
- our application is using Ember 3.x, so we are using a forked version of `@percy/ember`
- error analysis implies that [@percy/ember](https://github.com/percy/percy-ember) needs to be updated as well.

Happy to provide you with more information, if needed.

cc: @mmun @aschenoni

itsjwala:
@c69-addepar thanks for raising this, can you please start a GitHub discussion thread for this?

c69-addepar:
Created a discussion in percy-ember https://github.com/percy/percy-ember/discussions/704

packages/dom/test/serialize-dom.test.js
samarsault: Using 1000

packages/dom/test/inject-polyfill.test.js
itsjwala: we're testing if polyfill converts template tags to shadow DOM, which is happening currently.

packages/dom/src/serialize-cssom.js
```diff
@@ -34,6 +42,47 @@ export function serializeCSSOM({ dom, clone }) {
...
+  // https://github.com/WICG/construct-stylesheets/issues/93
+  if (!adoptedStylesMap) {
+    adoptedStylesMap = new Map();
+  }
+
+  for (let sheet of dom.adoptedStyleSheets) {
+    const styleLink = document.createElement('link');
+    styleLink.setAttribute('rel', 'stylesheet');
+
+    if (!adoptedStylesMap.has(sheet)) {
```
itsjwala, line 56: * can sheet be used as a key? (Not sure if a hash function is implemented for sheet)
* add a test with multiple adopted sheets of same & different content to verify.

packages/dom/src/serialize-cssom.js
```diff
@@ -34,6 +42,47 @@ export function serializeCSSOM({ dom, clone }) {
...
+      const styles = Array.from(sheet.cssRules)
+        .map(cssRule => cssRule.cssText).join('\n');
+      let resource = resourceFromText(uid(), 'text/css', styles);
+      resources.add(resource);
+      adoptedStylesMap.set(sheet, resource.url);
+    }
+    styleLink.setAttribute('href', adoptedStylesMap.get(sheet));
```
itsjwala, line 63: Lets use 'data-percy-serialized-attribute-href' to avoid making network request (check video/canvas serialization)

packages/dom/src/serialize-cssom.js
samarsault: Yup, have already added a test here - https://github.com/percy/cli/blob/d4ff7c7256fa342b0d57a3014fcc24b9ff841c25/packages/dom/test/serialize-css.test.js#L97-L136

packages/dom/src/serialize-cssom.js
samarsault: Nice!

samarsault (approved):

c69-addepar:
Confirmed:
- `percy/cli 1.19.2` and `percy/ember 4.0.2` fix our issues.

Thank you, team !

mmun:
Thanks again @itsjwala and @samarsault for your work. I hope it brings additional value to all your customers!

By the way, Safari [announced recently](https://webkit.org/blog/13851/declarative-shadow-dom/) that their technical preview has enabled declarative shadow DOM APIs by default. I'm not sure if this inludes the `getInnerHTML({ includeShadowRoots: true })` API though.

itsjwala:
@mmun yes we're keeping an eye on this development for Safari, though it would be particularly be helpful for browsers in infrastructure *(we only support `chrome/firefox` as a test browser so serialization point of view we want Firefox to pickup this development soon)*

For now, declarative shadow DOM polyfill - [ref](https://github.com/percy/cli/pull/1165#issue-1521070245) is working for `Firefox`, `Safari`, and `Safari on iPhone` platforms in our infrastructure.

PR #1175, opened 2023-01-29
 🐛 fix device pixel ratio config for percy exec

samarsault:
device-pixel-ratio setting in config was not getting passed down to asset discovery function. This fixes that

### DIFF
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -193,7 +193,8 @@ async function* captureSnapshotResources(page, snapshot, options) {
 
     yield* captureSnapshotResources(page, snapshot, {
       deviceScaleFactor: discovery.devicePixelRatio,
-      mobile: true
+      mobile: true,
+      capture: snapshot.domSnapshot ? null : capture
     });
   }
 
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -43,6 +43,7 @@ function debugSnapshotOptions(snapshot) {
   debugProp(snapshot, 'execute.beforeSnapshot');
   debugProp(snapshot, 'discovery.allowedHostnames');
   debugProp(snapshot, 'discovery.disallowedHostnames');
+  debugProp(snapshot, 'discovery.devicePixelRatio');
   debugProp(snapshot, 'discovery.requestHeaders', JSON.stringify);
   debugProp(snapshot, 'discovery.authorization', JSON.stringify);
   debugProp(snapshot, 'discovery.disableCache');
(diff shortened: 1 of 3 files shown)


packages/core/src/discovery.js
```diff
@@ -192,7 +193,8 @@ async function* captureSnapshotResources(page, snapshot, options) {
 
     yield* captureSnapshotResources(page, snapshot, {
       deviceScaleFactor: discovery.devicePixelRatio,
-      mobile: true
+      mobile: true,
+      capture: snapshot.domSnapshot ? null : capture
```
samarsault, line 197: Since we end up calling capture twice, it ends up pushing twice to queue

packages/core/src/discovery.js
```diff
@@ -132,7 +133,6 @@ async function* captureSnapshotResources(page, snapshot, options) {
     if (captureWidths) options = { ...options, width };
     let captured = await page.snapshot(options);
     captured.resources.delete(normalizeURL(captured.url));
-    capture(processSnapshotResources(captured));
```
itsjwala, line 135: is L134 required?

packages/core/src/discovery.js
```diff
@@ -197,9 +197,15 @@ async function* captureSnapshotResources(page, snapshot, options) {
...
 
   // wait for final network idle when not capturing DOM
-  if (capture && snapshot.domSnapshot) {
+  if (capture) {
     yield waitForDiscoveryNetworkIdle(page, discovery);
-    capture(processSnapshotResources(snapshot));
+    if (!snapshot.domSnapshot) {
+      // invoked through cli-snapshot
+      snapshot.resources.delete(snapshot.url);
+      capture(processSnapshotResources({ domSnapshot: baseSnapshot.domSnapshot, ...snapshot }));
```
itsjwala, line 205: Is `baseSnapshot` always present?

packages/core/src/discovery.js
samarsault: Yes, we set it to `baseSnapshot` from which we eventually extract the page HTML

itsjwala: Sorry I meant L135

packages/core/src/snapshot.js
```diff
@@ -110,6 +110,7 @@ function getSnapshotOptions(options, { config, meta }) {
       allowedHostnames: config.discovery.allowedHostnames,
       disallowedHostnames: config.discovery.disallowedHostnames,
       networkIdleTimeout: config.discovery.networkIdleTimeout,
+      devicePixelRatio: config.discovery.devicePixelRatio,
```
itsjwala, line 113: * add unit tests in `packages/core/test/unit/snapshot.test.js`
* update desc.
* Create new Github issue DPR fix `cli-snapshot`, with possible solution if any.

itsjwala (approved):

PR #1184, opened 2023-02-14
:sparkles: Add blob image assets serialization support

itsjwala:
TBA

### DIFF
--- a/packages/dom/src/prepare-dom.js
+++ b/packages/dom/src/prepare-dom.js
@@ -5,7 +5,7 @@ export function uid() {
 
 export function markElement(domElement, disableShadowDOM) {
   // Mark elements that are to be serialized later with a data attribute.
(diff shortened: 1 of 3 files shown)


itsjwala:
we'll revisit this PR later, closing for now.

AaronBuxbaum:
@itsjwala 

I'm working on some functionality which involves blob image src; having this would be very useful!

itsjwala:
hey @AaronBuxbaum, we'd to drop this PR because it required some more work - in the current state serialization performance got hit from ~500ms to ~3-5 seconds in some of our tests. 

workaround for this currently would be using [domTransformation](https://github.com/percy/cli/tree/master/packages/dom#examples) - What needs to be done is replace the blob assets link with the actual link of the image in the `src` attribute, this will trigger network request in the [asset discovery phase](https://docs.percy.io/docs/debugging-sdks#how-percy-works).

PR #1185, opened 2023-02-15
✨  DOM events triggered on clone fixed

samarsault:
### What is this?
In https://github.com/percy/cli/pull/1165, we refactored our DOM cloning logic. We moved to creating a clone and adding it to a cloned node of document. This, however triggers DOM events like connectedCallback which mess up the user DOM. This PR moves from keeping the clone node in a document to a `DocumentFragment`

### DIFF
--- a/packages/dom/src/serialize-cssom.js
+++ b/packages/dom/src/serialize-cssom.js
@@ -22,8 +22,8 @@ function styleSheetFromNode(node) {
   /* istanbul ignore if: sanity check */
   if (node.sheet) return node.sheet;
 
-  // Cloned style nodes inside don't have a sheet instance unless cloned along
-  // with document; we get it by temporarily adding the rules to DOM
+  // Cloned style nodes don't have a sheet instance unless they are within
+  // a document; we get it by temporarily adding the rules to DOM
   const tempStyle = document.createElement('style');
   tempStyle.setAttribute('data-percy-style-helper', '');
   tempStyle.innerHTML = node.innerHTML;
--- a/packages/dom/src/serialize-cssom.js
+++ b/packages/dom/src/serialize-cssom.js
@@ -18,6 +18,25 @@ function styleSheetsMatch(sheetA, sheetB) {
   return true;
 }
 
+function styleSheetFromNode(node) {
+  /* istanbul ignore if: sanity check */
+  if (node.sheet) return node.sheet;
+  /* istanbul ignore if: sanity check */
+  if (node.constructor.name !== 'HTMLStyleElement') return;
+
+  // Cloned style nodes inside don't have a sheet instance unless cloned along
+  // with document; we get it by temporarily adding the rules to DOM
+  const tempStyle = document.createElement('style');
+  tempStyle.id = 'style-percy-helper';
+  tempStyle.innerText = node.innerText;
+  document.head.appendChild(tempStyle);
+  const sheet = tempStyle.sheet;
+  // Cleanup node
+  tempStyle.remove();
+
+  return sheet;
+}
+
 export function serializeCSSOM({ dom, clone, warnings, resources, cache }) {
   // in-memory CSSOM into their respective DOM nodes.
   for (let styleSheet of dom.styleSheets) {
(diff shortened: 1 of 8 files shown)


packages/dom/src/serialize-cssom.js
```diff
@@ -18,6 +18,25 @@ function styleSheetsMatch(sheetA, sheetB) {
...
+
+  // Cloned style nodes inside don't have a sheet instance unless cloned along
+  // with document; we get it by temporarily adding the rules to DOM
+  const tempStyle = document.createElement('style');
+  tempStyle.id = 'style-percy-helper';
+  tempStyle.innerText = node.innerText;
+  document.head.appendChild(tempStyle);
+  const sheet = tempStyle.sheet;
+  // Cleanup node
+  tempStyle.remove();
```
itsjwala, line 35:
```suggestion
  // Cloned style nodes inside don't have a sheet instance unless added to the
  // document; we get it by temporarily adding the rules to the temporary clone document.
  let tempStyle = document.createElement('style');
  tempStyle.innerText = node.innerText;
  let cloneDOMNode = document.cloneNode();
  cloneDOMNode.appendChild(tempStyle);
  let sheet = tempStyle.sheet;
```

packages/dom/src/serialize-cssom.js
```diff
@@ -18,6 +18,25 @@ function styleSheetsMatch(sheetA, sheetB) {
   return true;
 }
 
+function styleSheetFromNode(node) {
+  /* istanbul ignore if: sanity check */
+  if (node.sheet) return node.sheet;
+  /* istanbul ignore if: sanity check */
+  if (node.constructor.name !== 'HTMLStyleElement') return;
```
itsjwala, line 25:
```suggestion
```

we'd ideally want to break here, otherwise stuff ahead of this will break and we won't have a clue.

packages/dom/src/serialize-cssom.js
itsjwala: Nice workaround 🔥

packages/dom/test/serialize-dom.test.js
```diff
@@ -34,6 +34,25 @@ describe('serializeDOM', () => {
...
+    class CallbackTestElement extends window.HTMLElement {
+      connectedCallback() {
+        const wrapper = document.createElement('h2');
+        wrapper.className = 'callback';
+        wrapper.innerText = 'Test';
+        this.appendChild(wrapper);
+      }
+    }
+
+    if (!window.customElements.get('cllback-test')) {
```
itsjwala, line 47:
```suggestion
    if (!window.customElements.get('callback-test')) {
```

itsjwala: does this test fail without the above ^ code changes?

packages/dom/src/serialize-cssom.js
```diff
@@ -75,6 +75,7 @@ export function serializeCSSOM({ dom, clone, warnings, resources, cache }) {
       resources.add(resource);
       cache.set(sheet, resource.url);
     }
+    styleLink.setAttribute('data-percy-adopted-stylesheets-serialized', 'true');
```
samarsault, line 78: Can you update the tests as well

packages/dom/src/serialize-cssom.js
itsjwala: * keep this code and debug with `youtube.com` and understand what value `node.constructor.name` has in that case.
  * we can include that case as well.

packages/dom/test/serialize-dom.test.js
samarsault: no its just a sanity check

packages/dom/src/serialize-cssom.js
samarsault: It was happening for `HTMLSVGStyleElement`, which is just a style tag. Its same to remove the condition

packages/dom/test/serialize-dom.test.js
itsjwala: I meant, the code changes of `v1.19.2-alpha.0`

packages/dom/src/serialize-cssom.js
itsjwala: Nice

itsjwala (approved):

PR #1192, opened 2023-03-01
🐛 fix for parent's baseURI used for iframes

### LINKED ISSUE #1189: iFrame missing css
## The problem

After upgrading to v1.19.2 the snapshots of iframed content does not render any CSS.

See https://percy.io/283b5e1b/ProdigyEMS/builds/25284757/changed/1410160725?browser=chrome&browser_ids=33%2C34&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280

## Environment

- Node version:
- `@percy/cli` version: v1.19.2
- Version of Percy SDK you’re using: @percy/cypress v3.1.2
- If needed, a build or snapshot ID: 25284757
- OS version: linux
- Type of shell command-line [interface]: sh

## Details

I believe it could be related to #1165

## Debug logs

If you are reporting a bug, _always_ include logs! [Give the "Debugging SDKs"
document a quick read for how to gather logs](https://docs.percy.io/docs/debugging-sdks#debugging-sdks)

Please do not trim or edit these logs, often times there are hints in the full
logs that help debug what is going on.

## Code to reproduce issue

Given the nature of testing/environment bugs, it’s best to try and isolate the
issue in a reproducible repo. This will make it much easier for us to diagnose
and fix.

itsjwala:
* while serializing the DOM for shadow DOM support in #1165 we introduced a bug where in we were using `document.createDocumentFragment`  function which creates a fragment of the parent's frame. 
* This will specifically cause issues for DOMs that have iframes and their domain is different since its `baseURI` would be different from the parent frame.
* To fix this, we're passing the iframe's document while serializing and using this object to create the fragment. 
* Fixes #1189


packages/dom/src/clone-dom.js
```diff
@@ -42,20 +42,19 @@ const deepClone = (host, disableShadowDOM) => {
...
  * Deep clone a document while also preserving shadow roots and converting adoptedStylesheets to <style> tags.
  */
 const cloneNodeAndShadow = (ctx) => {
-  let cloneDocumentFragment = deepClone(ctx.dom.documentElement, ctx.disableShadowDOM);
-  cloneDocumentFragment.documentElement = cloneDocumentFragment.firstChild;
-  cloneDocumentFragment.head = cloneDocumentFragment.querySelector('head');
-  cloneDocumentFragment.body = cloneDocumentFragment.querySelector('body');
-  return cloneDocumentFragment;
+  return deepClone(ctx.dom, ctx.disableShadowDOM);
 };
```
samarsault, line 58: This function seems redundant, we can simply do `deepClone(ctx)`

itsjwala: addressed https://github.com/percy/cli/pull/1192/commits/228a6c6cf15c799daa7cd5931bb3b23dc0a0c2d7

packages/dom/test/serialize-frames.test.js
```diff
@@ -60,6 +62,27 @@ describe('serializeFrames', () => {
...
+    if (getTestBrowser() === chromeBrowser) {
+      // we use plain & shadow platform
+      timesCalled = 2;
+    } else if (getTestBrowser() === firefoxBrowser) {
+      // we use only plain platform
+      timesCalled = 1;
+    }
+
+    expect(window.document.createDocumentFragment).toHaveBeenCalledTimes(timesCalled);
+  });
```
samarsault, line 84: Can we have the test check for the base uri's instead? This one doesn't seem very intuitive

samarsault (approved):

PR #1193, opened 2023-03-01
👷 update error code for retry logic

pankaj443:
- we have updated verify endpoint to raise err code 400, we previously were retying on 409 hence need to update it to fix the retry logic.
- currently, if a tile is not uploaded in 500ms which is the initial wait time we throw an error.

### DIFF
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -459,7 +459,7 @@ export class PercyClient {
         }
       });
     } catch (error) {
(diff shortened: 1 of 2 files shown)


itsjwala (commented):
lgtm, verify with team as well

itsjwala (approved):

ninadbstack (approved):

PR #1195, opened 2023-03-03
📝 Remove publish info

samarsault:
* moved to internal confluence


packages/cli/README.md
```diff
@@ -78,30 +78,3 @@ $ yarn workspace @percy/core test
 ### How to update Chromium revision?
 
 check in Core Package's readme [here](./packages/core#readme).
-
-## Publish
```
itsjwala, line 82: lets keep the header, and add a the link to handbook or text

Issue #1194, opened 2023-03-02
TypeError: Cannot read properties of null (reading 'cssRules')


tmcconechy:
## The problem

Upgraded from 1.18.0 to 1.20.1 and now get this error on most tests:

```
[percy] Could not take DOM snapshot "ids-data-grid-new-light"
[percy] Error: Evaluation failed: TypeError: Cannot read properties of null (reading 'cssRules')
    at styleSheetsMatch (pptr://__puppeteer_evaluation_script__:177:50)
    at serializeCSSOM (pptr://__puppeteer_evaluation_script__:210:15)
    at serializeElements (pptr://__puppeteer_evaluation_script__:465:9)
    at Object.serializeDOM (pptr://__puppeteer_evaluation_script__:504:7)
    at pptr://__puppeteer_evaluation_script__:3:23
```    

## Environment

- Node version:
- `@percy/cli` version: 1.20.1
- Version of Percy SDK you’re using:
- If needed, a build or snapshot ID: https://github.com/infor-design/enterprise-wc/actions/runs/4315837554/jobs/7530775761#step:10:267
- OS version: Unix/Mac
- Type of shell command-line [interface]: bash

## Details

Example test https://github.com/infor-design/enterprise-wc/blob/main/test/ids-data-grid/ids-data-grid-percy-test.ts

## Debug logs

https://github.com/infor-design/enterprise-wc/actions/runs/4315837554/jobs/7530775761#step:10:267

## Code to reproduce issue

If needed:
- clone https://github.com/infor-design/enterprise-wc
- change `"@percy/cli": "1.18.0",` to `"@percy/cli": "1.20.1",` in package.json
- `npm i && npm run test`

tmcconechy:
@itsjwala this fixed the cssRules error but i still cant use the latest version it gives some unknown errors now not showing up in the log https://github.com/infor-design/enterprise-wc/actions/runs/4365577412/jobs/7637547882#step:10:29623

Just mentioning it in case you have an idea. If not i can try to get verbose logs turned on (is that possible I forget) when i have more time and make a new issue.

samarsault:
@tmcconechy we see some unusual diffs in the build and will take a look. Can you try using the latest version setting `enableJavascript:true` and `disableShadowDOM:true`?

tmcconechy:
@samarsault should i be using the support at browser stack instead? My concern is we can no longer update and use percy so might need to look at other options before renewing.

samarsault:
[...]

PR #1197, opened 2023-03-03
🐛 copy full style node for deriving CSS rules

samarsault:
Fixes https://github.com/percy/cli/issues/1194
In order to serialize the styles, we mount them temporarily on a document clone. The presence of a content security policy on the page was blocking this from happening, throwing the following error:
```
Refused to apply inline style because it violates the following Content Security Policy directive:
```
To fix this, the node should have a [nonce](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce).

### DIFF
--- a/packages/dom/test/utils.test.js
+++ b/packages/dom/test/utils.test.js
@@ -1,0 +1,13 @@
+import { styleSheetFromNode } from '../src/utils';
+
+describe('utils', () => {
+  it('creates styleSheetFromNode properly', () => {
+    const node = document.createElement('style');
+    node.innerText = 'p { background-color: red }';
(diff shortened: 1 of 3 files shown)


itsjwala (approved):

packages/dom/test/utils.test.js
```diff
@@ -0,0 +1,13 @@
+import { styleSheetFromNode } from '../src/utils';
+
+describe('utils', () => {
```
itsjwala, line 3:
```suggestion
describe('styleSheetFromNode', () => {
```

convention is `describe` block contains the function/class name

itsjwala (approved):

PR #1205, opened 2023-03-10
✨ Check review-state in build:wait

AaronBuxbaum:
* `build:wait` when used with `--fail-on-changes` exits CLI with non-zero code, when the build is approved, ideally we'd want to exit 0 
* This PR adds an additional flag `--pass-if-approved` which will exit 0 when the build is approved even when diffs are found. 
* This flag only works when `--fail-on-changes` is passed


packages/cli-build/src/wait.js
```diff
@@ -124,12 +124,12 @@ function failureMessage(type, {
 
 // Return true or false if a build is considered failing or not
 function isFailing({
-  attributes: { state, 'total-comparisons-diff': diffs } = {}
+  attributes: { state, 'review-state': reviewState, 'total-comparisons-diff': diffs } = {}
```
itsjwala, line 127: * we'd want this behind an additional flag something like `failOnChangesPassIfApproved` 
`--fail-on-changes-pass-if-approved`, this is to avoid regression for existing users.

itsjwala (changes requested):
back to you @AaronBuxbaum

packages/cli-build/src/wait.js
AaronBuxbaum: done! @itsjwala

packages/cli-build/src/wait.js
```diff
@@ -36,6 +36,9 @@ export const wait = command('wait', {
     name: 'fail-on-changes',
     description: 'Exit with an error when diffs are found',
     short: 'f'
+  }, {
+    name: 'pass-if-approved',
+    description: 'Approve the build regardless of if diffs are found, if the build is approved. Overrides --fail-on-changes.',
```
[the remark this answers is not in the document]
itsjwala: How to add `inclusive:` is explained in another comment.

packages/cli-build/test/wait.test.js
```diff
@@ -133,6 +133,22 @@ describe('percy build:wait', () => {
     ]));
   });
 
+  it('does not error on diffs if the review status is approved and pass-if-approved is on', async () => {
```
itsjwala, line 136: Please add additional coverage with 
* `--pass-if-approved` passed `--fail-on-changes` not passed, and vice versa
* when both are passed.

itsjwala (commented):
Nice @AaronBuxbaum! 

I've left some additional comments, post which we're gtg

packages/cli-build/test/wait.test.js
AaronBuxbaum: done!

AaronBuxbaum:
@itsjwala ready to go!

Not sure how, but I'm seeing these tests fail:
```
stops waiting on process termination
logs an error when there are no snapshots
```

hopefully you can help resolve and then we can push this across the finish line. It might just be my local configuration!

AaronBuxbaum:
> @itsjwala ready to go!
> 
> Not sure how, but I was seeing these tests fail:
> 
> ```
> stops waiting on process termination
> logs an error when there are no snapshots
> ```
> 
> I updated them to pass, but give them a look and make any updates desired. then we can push this across the finish line. It might just be my local configuration!

based on CI, it looks like it might be my local configuration. Made a commit to remove those changes.

itsjwala (approved):
Thanks for your contribution @AaronBuxbaum 🔥 

Merging this PR *(windows CI are known to be flaky)* , This will be part of `1.21.0` release ETA end of week.

PR #1206, opened 2023-03-13
Added changes to support base64 content in postComparison API

ninadbstack:
Context:
- So far we passed `filepath` in the api and it was cli which used to read file to upload.
- Now we also want an option to pass file content directly as base64 in the request so that we dont need to pass filename / path

Changes:
- Updated codebase to check if content was passed and was base64 string then parse it into buffer and treat it same as content read from filepath


prklm10 (approved):
lgtm

itsjwala (approved):
lgtm 🔥

Issue #1204, opened 2023-03-09
BUG: Failed to execute 'btoa' on 'Window`  due to some CSS from bootstrap 5 


cgaube:
## The problem

When trying to create the snapshot, percy and percy/storybook  fail because it cannot serialize some text.
`Error: DOMException: Failed to execute 'btoa' on 'Window` 

```
// converts text to base64
let content = window.btoa(data);
```  
https://github.com/percy/cli/blob/520bd156af47080881b651e4819014cbacf76781/packages/dom/src/utils.js#LL23C35-L23C35

This is happening because im injecting some CSS  in the web element shadow dom
More precisely this is coming from BOOTSTRAP CSS  https://github.com/twbs/bootstrap/blob/main/scss/_type.scss#L104

```
 content: "\2014\00A0"; // em dash, nbsp
```

## Environment

- Node version: 18
- `@percy/cli` version: 1.20.3 
- Version of Percy SDK you’re using:  percy/storybook 4.3.5
- OS version: mac
- Type of shell command-line [interface]: zsh
- VITE 4.1.4  / Storybook 7 

## Debug logs

```
[percy:config] Config file not found (0ms)
[percy:storybook] Requesting Storybook: http://localhost:55245 (29ms)
[percy:core:browser] Launching browser (11ms)
[percy:core:browser] Browser connected [2966]: HeadlessChrome/96.0.4664.0 (335ms)
[percy:core] Percy has started! (1ms)
[percy:core:page] Page created (40ms)
[percy:core:page] Page created (2ms)
[percy:core:page] Navigate to: http://localhost:55245/?path=/settings/about (106ms)
[percy:core:page] Navigate to: http://localhost:55245/iframe.html (0ms)
[percy:core:page] Page navigated (20ms)
[percy:core:page] Page navigated (107ms)
[percy:core:page] Page closed (4ms)
[percy:core:page] Page closed (5005ms)
[percy:core:page] Page created (7ms)
[percy:core:page] Navigate to: http://localhost:55245/iframe.html (34ms)
[percy:core:page] Page navigated (97ms)
[percy:core:page] Taking snapshot: Theme/Buttons: Primary (11ms)
[percy:core:discovery] Wait for 100ms idle (1ms)
[percy:core:page] Inject @percy/dom (102ms)
[percy:core:page] Serialize DOM (3ms)
[percy:core:page] Page closed (5ms)
[percy:core] Stopping percy... (1ms)
[percy:core:browser] Closing browser (0ms)
[...]

PR #1209, opened 2023-03-13
Serialize text resources as utf8

samarsault:
Fixes https://github.com/percy/cli/issues/1204

Percy CLI Server expects base64 data for resources by default. There's no straightforward way to convert unicode javascript strings, which have a 2 byte character size to base64 in the browser. Hence, we modify CLI server to accept resources with a text mimetype with UTF8 encoding


itsjwala (commented):
* Add more description
* Add a test in discovery - when file is `text/css` it should use `utf` decode

itsjwala:
RUN_REGRESSION

itsjwala:
RUN_REGRESSION

itsjwala (approved):

PR #1210, opened 2023-03-14
🐛  Don't include shadow root when `disableShadowDOMSerialization` is passed

samarsault:
* When using chromium test browser with enable javascript, JS based shadow roots are being included in the snapshot causing double rendering (an extension of https://github.com/percy/cli/issues/1194), we'll suggest users to disable shadow dom serialization in this case.
* As part of this PR we're also renaming the param from `disableShadowDOM` to `disableShadowDOMSerialization`

### DIFF
--- a/packages/dom/src/clone-dom.js
+++ b/packages/dom/src/clone-dom.js
@@ -9,7 +9,7 @@ import markElement from './prepare-dom';
  * Deep clone a document while also preserving shadow roots
  * returns document fragment
  */
-export function cloneNodeAndShadow({ dom, disableShadowDOM }) {
+export function cloneNodeAndShadow({ dom, disableShadowDOMSerialization }) {
   // clones shadow DOM and light DOM for a given node
   let cloneNode = (node, parent) => {
     let walkTree = (nextn, nextp) => {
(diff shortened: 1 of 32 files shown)


samarsault:
RUN_REGRESSION

packages/dom/src/clone-dom.js
```diff
@@ -56,11 +56,13 @@ export function cloneNodeAndShadow({ dom, disableShadowDOM }) {
...
  */
-export function getOuterHTML(docElement) {
+export function getOuterHTML(ctx) {
+  const docElement = ctx.clone.documentElement;
   // firefox doesn't serialize shadow DOM, we're awaiting API's by firefox to become ready and are not polyfilling it.
   if (!docElement.getInnerHTML) { return docElement.outerHTML; }
   // chromium gives us declarative shadow DOM serialization API
-  let innerHTML = docElement.getInnerHTML({ includeShadowRoots: true });
+
+  let innerHTML = docElement.getInnerHTML({ includeShadowRoots: !ctx.disableShadowDOM && !ctx.enableJavaScript });
```
itsjwala, line 65: we should not serialize shadow DOM ( similar to disable shadow DOM) if JS is enable - this will avoid patches throughout the code

itsjwala:
RUN_REGRESSION

itsjwala:
This is not required currently, we'll revisit this later.

PR #1213, opened 2023-03-16
Update the reversePolyFill function so custom elements get defined

DouglasMeyer:
This helps solve issues where CSS wants to ensure custom elements are defined:

```css
custom-element:defined {
  /* ... */
}
```

### DIFF
--- a/packages/dom/src/inject-polyfill.js
+++ b/packages/dom/src/inject-polyfill.js
@@ -16,7 +16,16 @@ export function injectDeclarativeShadowDOMPolyfill(ctx) {
         template.remove();
       });
 
-      root.querySelectorAll('[data-percy-shadow-host]').forEach(shadowHost => reversePolyFill(shadowHost.shadowRoot));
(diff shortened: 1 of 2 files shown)


itsjwala:
@DouglasMeyer 

* This polyfill only runs for non-chromium browsers, can you enable JS for the snapshot which needs to rely on this CSS to be present?

DouglasMeyer:
@itsjwala I'm afraid I don't follow what change needs to be made. I do see some some tests in another package failing

itsjwala:
hey @DouglasMeyer 

* The polyfill only runs on non-chromium Infrastructure browsers in Percy, I believe Addepar uses chrome browser for their projects.
* You could try enabling javascript for particular snapshot - https://docs.percy.io/docs/cli-configuration#per-snapshot-configuration, in which case your your scripts will run in our environment. 
  *  you may also need to disable shadow dom serialization in this case if JS is enabled - [ref](https://github.com/percy/cli/tree/master/packages/dom#options)

PR #1220, opened 2023-03-30
Regression: WDIO test v2 and add more SDKs

samarsault:
Follow up from https://github.com/percy/cli/pull/1211
- Add testing for WDIO v2
- Add more SDKs - python, capybara and appium-js
- Fix PR branch name not getting updated on `issue_comment` event

Reference Run - https://github.com/percy/cli/actions/runs/4565663845

### DIFF
--- a/.github/workflows/sdk-regression.yml
+++ b/.github/workflows/sdk-regression.yml
@@ -1,11 +1,14 @@
 name: SDK Regression
 on:
-  push
+  pull_request:
+    types: [opened]
+  issue_comment:
+    types: [created, edited] 
 jobs:
   regression:
     name: Trigger SDK regression test
     runs-on: ubuntu-latest
-    # if: ${{ github.event.issue.pull_request && github.event.comment.body == 'RUN_REGRESSION' }}
+    if: ${{ github.event.issue.pull_request && github.event.comment.body == 'RUN_REGRESSION' }}
     strategy:
       matrix:
         repo: 
@@ -45,5 +48,4 @@ jobs:
           github_token: ${{ secrets.WORKFLOW_DISPATCH_ACTIONS_TOKEN }}
           workflow_file_name: test.yml
           ref: ${{ steps.split.outputs._1 || 'master' }}
-          client_payload: '{ "branch": "master" }'
-          wait_interval: 15
+          client_payload: '{ "branch": "${{ steps.comment-branch.outputs.head_ref }}"}'
--- a/.github/workflows/sdk-regression.yml
+++ b/.github/workflows/sdk-regression.yml
@@ -40,7 +40,7 @@ jobs:
         uses: convictional/trigger-workflow-and-wait@v1.6.5
         with:
           owner: percy
-          repo: ${{ matrix.repo }}
+          repo: ${{ steps.split.outputs._0 }}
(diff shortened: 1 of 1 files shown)


.github/workflows/sdk-regression.yml
```diff
@@ -21,19 +21,30 @@ jobs:
...
           - percy-protractor
           - percy-selenium-js
           - percy-selenium-dotnet
           - percy-selenium-java
+          - percy-selenium-python
+          - percy-capybara
+          - percy-appium-js
           - gatsby-plugin-percy
     steps:
+      - uses: jungwinter/split@v2
```
itsjwala, line 34: let's add a comment on L14, on how to use this, doesn't seem intuitive on first glance.

.github/workflows/sdk-regression.yml
```diff
@@ -21,19 +21,30 @@ jobs:
...
         uses: convictional/trigger-workflow-and-wait@v1.6.5
         with:
           owner: percy
           repo: ${{ matrix.repo }}
           github_token: ${{ secrets.WORKFLOW_DISPATCH_ACTIONS_TOKEN }}
           workflow_file_name: test.yml
-          ref: master
-          client_payload: '{ "branch": "${{ github.event.pull_request.head.ref || github.ref_name }}"}'
+          ref: ${{ steps.split.outputs._1 || master }}
+          client_payload: '{ "branch": "${{ steps.comment-branch.outputs.head_ref || github.ref_name }}"}'
```
itsjwala, line 49: I missed this before, why do we need `github.ref_name` this? we should fail the build if `comment-branch` fails, we don't want incorrect refs to trigger the build and have the checks green.

itsjwala (changes requested):
@samarsault back to you, link previous PR in the description

.github/workflows/sdk-regression.yml
itsjwala: * documentation around not the actions itself, but how are we using it in ourcase, like what does stuff after `@ ` mean to us if someone is adding a new branch they'll read and straight away add that.
* currently reading `percy-webdriverio` and `percy-webdriverio@v2` doesn't sound very intuitive imo.

samarsault: Added doc

.github/workflows/sdk-regression.yml
```diff
@@ -21,19 +21,32 @@ jobs:
...
           owner: percy
           repo: ${{ matrix.repo }}
           github_token: ${{ secrets.WORKFLOW_DISPATCH_ACTIONS_TOKEN }}
           workflow_file_name: test.yml
-          ref: master
-          client_payload: '{ "branch": "${{ github.event.pull_request.head.ref || github.ref_name }}"}'
-          wait_interval: 15
+          ref: ${{ steps.split.outputs._1 || 'master' }}
+          client_payload: '{ "branch": "${{ steps.comment-branch.outputs.head_ref }}"}'
+          wait_interval: 15
```
itsjwala, line 52: EOF

.github/workflows/sdk-regression.yml
itsjwala: can you move it to L14 maybe ( basically where we are using it) we're not concerned about the implementation details of the actions

itsjwala (approved):

itsjwala (approved):

PR #1227, opened 2023-04-04
Fix - percy upload for Safari browser

hiteshsalavi:
`localhost` keyword not supported in Safari. That causes `percy upload` html resource to be unable to fetch image resource

### DIFF
--- a/packages/cli-upload/src/utils.js
+++ b/packages/cli-upload/src/utils.js
@@ -20,7 +20,7 @@ export async function getImageResources({
   absolutePath
 }) {
   let rootUrl = `http://localhost/${encodeURIComponent(name)}`;
-  let imageUrl = `http://localhost/${encodeURIComponent(relativePath)}`;
+  let imageUrl = `http://local/${encodeURIComponent(relativePath)}`;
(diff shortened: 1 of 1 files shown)


packages/cli-upload/src/utils.js
```diff
@@ -20,7 +20,7 @@ export async function getImageResources({
   absolutePath
 }) {
   let rootUrl = `http://localhost/${encodeURIComponent(name)}`;
-  let imageUrl = `http://localhost/${encodeURIComponent(relativePath)}`;
+  let imageUrl = `http://local/${encodeURIComponent(relativePath)}`;
```
itsjwala, line 23: 1. lets make it consistent throughout -  say `render.percy.local` etc
2. we need to test this properly, we don't want any regression for existing users.

itsjwala (changes requested):

packages/cli-upload/src/utils.js
samarsault: Lets keep it local but consistent across

itsjwala (approved):

PR #1228, opened 2023-04-04
✨ increase interval for polling build

itsjwala:
* Increase polling interval for build to default to 10s.
* polling 1s is unnecessary and also causes stress on API, A build usually doesn't finish in 1s granularity. if less than 1s is given we'll clamp it to 1s

### DIFF
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -200,7 +200,7 @@ export class PercyClient {
     interval = 10_000
   }, onProgress) {
     if (interval < MIN_POLLING_INTERVAL) {
-      this.log.warn(`Considering interval ${MIN_POLLING_INTERVAL}ms, it cannot be less than that.`);
+      this.log.warn(`Ignoring interval since it cannot be less than ${MIN_POLLING_INTERVAL}ms.`);
       interval = MIN_POLLING_INTERVAL;
     }
     if (!project && commit) {
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -15,6 +15,8 @@ import {
 // Default client API URL can be set with an env var for API development
 const { PERCY_CLIENT_API_URL = 'https://percy.io/api/v1' } = process.env;
 const pkg = getPackageJSON(import.meta.url);
+// minimum polling interval milliseconds
+const MIN_POLLING_INTERVAL = 1_000;
 
 // Validate ID arguments
 function validateId(type, id) {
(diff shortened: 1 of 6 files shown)


packages/client/src/client.js
```diff
@@ -239,6 +248,7 @@ export class PercyClient {
 
         // not finished, poll again
         if (pending) {
+          setTimeout(self.log.debug, interval, 'waiting...');
```
samarsault, line 251: Why do we need setTimeout again? Isn't this entire function being called after `interval`? In that case we can just use `log.debug(...)`

yashdsaraf:
As per discussions with product
1. Min allowed interval will be 1000ms (1 sec) and default will be 10000ms (10 secs).
2. We will **not** break if supplied interval is less than allowed interval, simply warn and take the max of both.
3. Need to print a message after each "non-final" update about the ETA to next update.

packages/client/src/client.js
itsjwala: @yashdsaraf I just forgot to mention this, we're using `setTimeout` here because only then its possible to show
 `waiting... (10000ms)`

we can't use it directly otherwise it will show something like

```
waiting... (1ms)
Get Build 1234 (10100ms)
```

itsjwala: removed this log line for now

packages/client/src/client.js
```diff
@@ -188,15 +190,24 @@ export class PercyClient {
     return this.get(`projects/${project}/builds?${qs}`);
   }
 
+  // required so we could mock this method in tests :P
+  isIntervalLessThanThreshold(interval) {
+    return interval < MIN_POLLING_INTERVAL;
+  }
+
```
samarsault, line 197: We're not using this in tests now, so can this be removed?

itsjwala: removed

packages/client/src/client.js
```diff
@@ -195,8 +197,12 @@ export class PercyClient {
     project,
     commit,
     timeout = 10 * 60 * 1000,
-    interval = 1000
+    interval = 10_000
   }, onProgress) {
+    if (interval < MIN_POLLING_INTERVAL) {
+      this.log.warn(`Considering interval ${MIN_POLLING_INTERVAL}ms, it cannot be less than that.`);
```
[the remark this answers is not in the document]
itsjwala: updated

samarsault (approved):

PR #1230, opened 2023-04-07
🐛 cloneNode ensure shallow clone

samarsault:
WebComponents have [lifecycle callbacks](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#using_the_lifecycle_callbacks) which can mess up the structure of the page while trying to create a clone. We'd earlier fixed it for `connectedCallback` and `disconnectedCallback` in https://github.com/percy/cli/pull/1185. `attributeChangedCallback` is called before `connectedCallback` and can also cause side effects when we clone a node. We solve for this effect by asserting the shallowness of the node cloned in this PR

### DIFF
--- a/packages/dom/src/clone-dom.js
+++ b/packages/dom/src/clone-dom.js
@@ -29,9 +29,7 @@ export function cloneNodeAndShadow({ dom, disableShadowDOM }) {
     // shallow clone should not contain children
     // these might get added due to DOM side effects
     if (clone.children) {
-      for (const child of Array.from(clone.children)) {
-        clone.removeChild(child);
-      }
+      Array.from(clone.children).forEach(child => clone.removeChild(child));
     }
 
     // clone shadow DOM
--- a/packages/dom/src/clone-dom.js
+++ b/packages/dom/src/clone-dom.js
@@ -23,6 +23,7 @@ export function cloneNodeAndShadow({ dom, disableShadowDOM }) {
     markElement(node, disableShadowDOM);
 
     let clone = node.cloneNode();
+
     parent.appendChild(clone);
 
     // shalow clone should not contain children
--- a/packages/dom/src/clone-dom.js
+++ b/packages/dom/src/clone-dom.js
@@ -23,8 +23,15 @@ export function cloneNodeAndShadow({ dom, disableShadowDOM }) {
     markElement(node, disableShadowDOM);
 
     let clone = node.cloneNode();
-
     parent.appendChild(clone);
+
+    // shalow clone should not contain children
+    // these might get added due to DOM side effects
+    if (clone.children) {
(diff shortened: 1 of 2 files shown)


packages/dom/src/clone-dom.js
```diff
@@ -26,6 +26,14 @@ export function cloneNodeAndShadow({ dom, disableShadowDOM }) {
 
     parent.appendChild(clone);
 
+    // shalow clone should not contain children
+    // these might get added due to DOM side effects
+    if (clone.children) {
+      for (const child of Array.from(clone.children)) {
+        clone.removeChild(child);
```
itsjwala, line 33: clone.children is already an array I believe.

Use this may be

clone.children.forEach(clone.removeChild)

samarsault: `element.children` returns a [HTMLCollection](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection) type ([ref](https://developer.mozilla.org/en-US/docs/Web/API/Element/children)), which is not iterable by default. Hence we need this

itsjwala: okay, lets do it like this? `Array.from(clone.children).forEach(clone.removeChild)`

samarsault: Screenshot 2023-04-07 at 11 26 39 PM

itsjwala: oops yup syntax, my bad

packages/dom/src/clone-dom.js
```diff
@@ -26,6 +26,14 @@ export function cloneNodeAndShadow({ dom, disableShadowDOM }) {
 
     parent.appendChild(clone);
 
+    // shalow clone should not contain children
+    // these might get added due to DOM side effects
```
itsjwala, line 30:
```suggestion
    // shallow clone should not contain children
    // these might get added due to DOM side effects Eg: connectedCallbacks
```

packages/dom/src/clone-dom.js
```diff
@@ -26,6 +26,12 @@ export function cloneNodeAndShadow({ dom, disableShadowDOM }) {
 
     parent.appendChild(clone);
 
+    // shallow clone should not contain children
+    // these might get added due to DOM side effects
```
itsjwala, line 30:
```suggestion
    // these might get added due to DOM side effects
```
remove this line, or add more descriptive text - what are DOM side effects?

itsjwala (approved):

itsjwala:
This is related to some frameworks causing these callbacks to be triggered, currently patching on our side only seems to be the fastest option. though I'd like DOM APIs to support a parameter in the callback that makes it aware of cloning in general.

PR #1232, opened 2023-04-12
👷 Add storybook & selenium ruby to regression

samarsault:
We fixed storybook and selenium ruby regression issues in https://github.com/percy/percy-storybook/pull/723 and https://github.com/percy/percy-selenium-ruby/pull/9 respectively. Now they can be added to the regression pipeline


itsjwala:
SDK_REGRESSION

PR #1237, opened 2023-04-19
Regression Fix job url and status update

samarsault:
We don't have access to job id needed to build the URL in the action context. We fetch it using the API.


itsjwala (approved):
@samarsault

PR #1238, opened 2023-04-20
:sparkles: support domTransformation in Page.snapshot function

itsjwala:
* `domTransformation` is a feature of `@percy/dom` package where in we can run arbitrary JS to modify the snapshotted DOM post serialization - [ref](https://github.com/percy/cli/tree/master/packages/dom#options)
* `Page.snapshot` is used by `cli-snapshot` and `percy-storybook` packages, other SDKs don't borrow `Page` from the asset discovery browser we inject `PercyDOM` script in the user's test browser. 
* With this PR cli snapshot and storybook SDK will be able to leverage `domTransformation` param.

### DIFF
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -39,6 +39,9 @@ export const configSchema = {
         type: 'boolean',
         default: false
       },
+      domTransformation: {
+        type: 'string'
+      },
       scope: {
         type: 'string'
       }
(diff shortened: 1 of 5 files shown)


itsjwala:
RUN_REGRESSION

itsjwala:
RUN_REGRESSION

itsjwala:
RUN_REGRESSION

PR #1244, opened 2023-04-25
Regression branch validation

samarsault:
Branch input as is can be susceptible to injection attacks

### DIFF
--- a/.github/workflows/sdk-regression.yml
+++ b/.github/workflows/sdk-regression.yml
@@ -6,7 +6,7 @@ jobs:
   regression:
     name: regression
     runs-on: ubuntu-latest
-    if: ${{ github.event.issue.pull_request && github.event.comment.body == 'RUN_REGRESSION' }}
+    if: ${{ github.event.issue.pull_request && github.event.comment.body == 'RUN_REGRESSION' && !github.event.pull_request.head.repo.fork }}
     strategy:
       matrix:
         # Format: repo@branch (default branch is master)
--- a/.github/workflows/sdk-regression.yml
+++ b/.github/workflows/sdk-regression.yml
@@ -34,12 +34,15 @@ jobs:
       - uses: xt0rted/pull-request-comment-branch@v2
         if: ${{ github.event.issue.pull_request }}
         id: comment-branch
-      - name: Validate branch name
-        run: |
-          if [[ ! "${{ steps.comment-branch.outputs.head_ref }}" =~ ^[a-zA-Z0-9_\-]+$ ]]; then
-            echo "Invalid branch name"
-            exit 1
-          fi
+      - uses: actions-ecosystem/action-regex-match@v2
+        id: regex-match
+        with:
+          text: ${{ steps.comment-branch.outputs.head_ref }}
+          regex: '^[a-zA-Z0-9_/\-]+$'
+      - name: Break on invalid branch name
+        run: exit 1
+        if: ${{ steps.regex-match.outputs.match == '' }}
+
       - name: Get Current Job Log URL
         uses: Tiryoh/gha-jobid-action@v0
         id: job-url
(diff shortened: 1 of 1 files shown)


.github/workflows/sdk-regression.yml
```diff
@@ -34,6 +34,12 @@ jobs:
       - uses: xt0rted/pull-request-comment-branch@v2
         if: ${{ github.event.issue.pull_request }}
         id: comment-branch
+      - name: Validate branch name
+        run: |
+          if [[ ! "${{ steps.comment-branch.outputs.head_ref }}" =~ ^[a-zA-Z0-9_\-]+$ ]]; then
```
itsjwala, line 39: lets include forward slashes as well, sometimes we use that

samarsault: Added

.github/workflows/sdk-regression.yml
```diff
@@ -34,6 +34,15 @@ jobs:
       - uses: xt0rted/pull-request-comment-branch@v2
         if: ${{ github.event.issue.pull_request }}
         id: comment-branch
+      - uses: actions-ecosystem/action-regex-match@v2
+        id: regex-match
+        with:
+          text: ${{ steps.comment-branch.outputs.head_ref }}
+          regex: '^[a-zA-Z0-9_/\-]+$'
```
rahul-bs, line 41: is regex tested?

rahul-bs: Screenshot 1945-02-06 at 12 51 09 PM

samarsault: We don't need to escape / for JS regex. On regexr:
Screenshot 2023-04-26 at 5 27 31 PM

.github/workflows/sdk-regression.yml
```diff
@@ -6,7 +6,7 @@ jobs:
   regression:
     name: regression
     runs-on: ubuntu-latest
-    if: ${{ github.event.issue.pull_request && github.event.comment.body == 'RUN_REGRESSION' }}
+    if: ${{ github.event.issue.pull_request && github.event.comment.body == 'RUN_REGRESSION' && !github.event.pull_request.head.repo.fork }}
```
rahul-bs, line 9: is this tested?

.github/workflows/sdk-regression.yml
rahul-bs, line 41: Please add same check in all child repos too
Reason: they run on workflow_dispatch, so anyone can trigger action in child repo manually

rahul-bs: This comment is not addressed

rahul-bs:
just to confirm, since you are adding fork check to not run regression which means external users will not be able to run the regression tests with comments
and only our team would be able to contribute to the project

.github/workflows/sdk-regression.yml
samarsault: Running workflow dispatch in the child repo needs our token

samarsault:
With this change no one will be able to trigger the regression tests on a forked PR. We don't want that. Lets limit the people allowed to run the tests to members with write access

.github/workflows/sdk-regression.yml
rahul-bs: check should be in both places, since shell injection is present in that repo, if run type is changed, the vulnerability can be exploited

.github/workflows/sdk-regression.yml
```diff
@@ -31,9 +31,31 @@ jobs:
           - percy-appium-js
           - gatsby-plugin-percy
     steps:
+      - name: Get user permissions
```
ninadbstack, line 34: isnt this file editable by user in a pr ?

samarsault: Yes but the run would use the workflow file from master unless it is merged

rahul-bs (approved):

itsjwala (approved):

rahul-bs (approved):

PR #1250, opened 2023-05-11
Pp 2059

chinmay-browserstack:


samarsault:
Update PR description

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -35,11 +35,19 @@ export default class GenericProvider {
     return true;
   }
 
-  async screenshot(name) {
+  async screenshot(name, {
+    ignoreRegionXpaths = [],
+    ignoreRegionSelectors = [],
+    ignoreRegionSeleniumElements = [],
+    customIgnoreRegions = []
```
ninadbstack, line 42: while you are giving defaults here, they are usually not useful. Ideally add defaults in first lines of function. The reason being if i call above with 
`screenshot("abc", { ignoreRegionXpaths: ['some', 'paths'] })`
in the function `ignoreRegionSelectors` will be `undefined` and not `[]` as you have given in defaults.

ninadbstack (commented):
Overall comment for logging, in CLI assume that multiple percyScreenshot calls will be processed in parallel, current logging would make it hard to find which request caused it. [ it may not be easily possible right now, but think about it

packages/sdk-utils/test/helpers.js
```diff
@@ -26,6 +26,12 @@ const helpers = {
     return Array.isArray(res.body[what]) ? res.body[what].map(map) : map(res.body);
   },
 
+  async mockGetCurrentUrl() {
+    try {
+      await utils.request('/wd/hub/session');
+    } catch {}
+  },
+
```
samarsault, line 34: Where is this used?

packages/sdk-utils/src/index.js
```diff
@@ -17,7 +18,8 @@ export {
   fetchPercyDOM,
   postSnapshot,
   postComparison,
-  flushSnapshots
+  flushSnapshots,
+  captureAutomateScreenshot
```
samarsault, line 22: This should be `postAutomateScreenshot`

packages/sdk-utils/src/post-screenshot.js
```diff
@@ -0,0 +1,18 @@
+import percy from './percy-info.js';
+import request from './request.js';
+
+// Post screenshot data to the CLI automateScreenshot endpoint. If the endpoint responds with a build error,
+// indicate that Percy has been disabled.
+export async function captureAutomateScreenshot(options, params) {
+  let query = params ? `?${new URLSearchParams(params)}` : '';
```
samarsault, line 7: Can we use a default value for params here instead?

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -92,4 +101,104 @@ export default class GenericProvider {
...
+        const selector = `element: ${index}`;
+
+        const ignoredRegion = await this.ignoreElementObject(selector, elements[index]);
+        ignoredElementsArray.push(ignoredRegion);
+      } catch (e) {
+        log.warn(`Correct Web Element not passed at index ${index}.`);
+        log.debug(e.toString());
+      }
+    }
+  }
```
samarsault, line 179: We should avoid mutations in the array. Lets do so that the functions return an array and we concat all in the end.

Also, all this can be replaced by a single `ignoreRegionsBy(findElementFn, elements)`

packages/sdk-utils/src/index.js
chinmay-browserstack: https://github.com/percy/cli/pull/1250#discussion_r1213158152

packages/sdk-utils/test/helpers.js
chinmay-browserstack: This is used in test case by js SDK.https://github.com/percy/percy-selenium-js/pull/333/files#diff-8a8c49bdbe55241012d9ad47a61d9f11f11a2be0105f967811a8f2361f088b08R84-R85

packages/sdk-utils/src/post-screenshot.js
chinmay-browserstack: it is copied from https://github.com/percy/cli/blob/5f86910347e74cb9125ec8a5fb7d90155af44b04/packages/sdk-utils/src/post-snapshot.js#L7

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -92,4 +101,95 @@ export default class GenericProvider {
...
+    const ignoreElementSelectors = await this.ignoreRegionsBy('css selector', ignoreRegionSelectors);
+    const ignoreElements = await this.ignoreRegionsByElement(ignoreRegionSeleniumElements);
+    const ignoreElementCustom = await this.addCustomIgnoreRegions(customIgnoreRegions);
+
+    const ignoredElementsLocations = {
+      ignoreElementsData: [...ignoreElementXpaths, ...ignoreElementSelectors, ...ignoreElements, ...ignoreElementCustom]
+    };
+
+    return ignoredElementsLocations;
+  }
```
samarsault, line 116:
```suggestion
    return {
      ignoreElementsData: [
       ...ignoreElementXpaths, 
       ...ignoreElementSelectors,
        ...ignoreElements,
         ...ignoreElementCustom
         ]
    };
  }
```

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -92,4 +101,95 @@ export default class GenericProvider {
...
+            right: customLocation.right
+          }
+        };
+        ignoredElementsArray.push(ignoredRegion);
+      } else {
+        log.warn(`Values passed in custom ignored region at index: ${index} is not valid`);
+      }
+    }
+    return ignoredElementsArray;
+  }
```
samarsault, line 194: Rename functions with prefix `get`

samarsault (approved):
LGTM

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -92,4 +101,98 @@ export default class GenericProvider {
...
+      ]
+    };
+  }
+
+  async ignoreElementObject(selector, elementId) {
+    const scaleFactor = parseInt(await this.metaData.devicePixelRatio());
+    const rect = await this.driver.rect(elementId);
+    const location = { x: parseInt(rect.x), y: parseInt(rect.y) };
+    const size = { height: parseInt(rect.height), width: parseInt(rect.width) };
+    const coOrdinates = {
```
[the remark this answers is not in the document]
chinmay-browserstack: not 100% sure but we are doing same in AppPercy and may be api was expecting coordinates in this key.

PR #1265, opened 2023-05-22
✨  Support loading lazy attribute 

ninadbstack:
Context:
- We currently did not drop `loading="lazy"` attribute while serialising dom
- We do not scroll in discovery phase

Issue:
- We do not capture resources, especially `img` src where `loading="lazy"` attr is present and image is out of viewport
- So when we render it on percy, the full page screenshot shows broken images [ for out of first viewport images ]

Fix:
- We now drop `loading` attribute for `img` and `iframe` element

### DIFF
--- a/packages/dom/src/clone-dom.js
+++ b/packages/dom/src/clone-dom.js
@@ -4,6 +4,7 @@
  * https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow#parameters
  */
 import markElement from './prepare-dom';
+import { maybeTranformElement } from './transform-dom';
 
 /**
  * Deep clone a document while also preserving shadow roots
@@ -23,6 +24,8 @@ export function cloneNodeAndShadow({ dom, disableShadowDOM }) {
     markElement(node, disableShadowDOM);
 
     let clone = node.cloneNode();
+
+    maybeTranformElement(clone);
 
     parent.appendChild(clone);
 
--- a/packages/dom/src/clone-dom.js
+++ b/packages/dom/src/clone-dom.js
@@ -4,7 +4,7 @@
  * https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow#parameters
  */
 import markElement from './prepare-dom';
-import { maybeTranformElement } from './transform-dom';
+import applyElementTransformation from './transform-dom';
 
 /**
  * Deep clone a document while also preserving shadow roots
@@ -25,7 +25,8 @@ export function cloneNodeAndShadow({ dom, disableShadowDOM }) {
 
     let clone = node.cloneNode();
 
-    maybeTranformElement(clone);
+    // We apply any element transformations here to avoid another treeWalk
+    applyElementTransformation(clone);
 
     parent.appendChild(clone);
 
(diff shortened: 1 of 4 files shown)


packages/dom/src/clone-dom.js
```diff
@@ -24,6 +25,8 @@ export function cloneNodeAndShadow({ dom, disableShadowDOM }) {
 
     let clone = node.cloneNode();
 
+    maybeTranformElement(clone);
```
samarsault, line 28: remove `maybe`

packages/dom/src/transform-dom.js
```diff
@@ -0,0 +1,13 @@
+// Drop loading attribute. We do not scroll page in discovery stage but we want to make sure that
+// all resources are requested, so we drop loading attribute [as it can be set to lazy]
+function dropLoadingAttribute(domElement) {
+  if (!['img', 'iframe'].includes(domElement.tagName?.toLowerCase())) return;
+  domElement.removeAttribute('loading');
+}
+
+// All transformations that we need to apply for a successful discovery and stable render
+export function applyElementTransformations(domElement) {
```
samarsault, line 9: Remove export from here

packages/dom/src/transform-dom.js
```diff
@@ -0,0 +1,13 @@
+// Drop loading attribute. We do not scroll page in discovery stage but we want to make sure that
+// all resources are requested, so we drop loading attribute [as it can be set to lazy]
+export function dropLoadingAttribute(domElement) {
+  if (!['img', 'iframe'].includes(domElement.tagName?.toLowerCase())) return;
+  domElement.removeAttribute('loading');
+}
+
+// All transformations that we need to apply for a successful discovery and stable render
+function applyElementTransformations(domElement) {
```
itsjwala, line 9: nice

itsjwala (approved):

packages/dom/test/transform-dom.test.js
```diff
@@ -0,0 +1,34 @@
...
+    dropLoadingAttribute(dom.getElementById('frame1'));
+
+    expect(dom.getElementById('frame1').getAttribute('loading')).toBe(null);
+  });
+
+  it('does not drop loading from other tags', () => {
+    dropLoadingAttribute(dom.getElementById('other_tag'));
+
+    expect(dom.getElementById('other_tag').getAttribute('loading')).toBe('lazy');
+  });
```
samarsault, line 33: Can we run these for both shadow and plain container nodes? See https://github.com/percy/cli/blob/master/packages/dom/test/serialize-videos.test.js for example

packages/dom/test/transform-dom.test.js
```diff
@@ -0,0 +1,34 @@
+import { dropLoadingAttribute } from '../src/transform-dom';
+import { withExample } from './helpers';
+
+// Note: applyElementTransformations is tested in serialize-dom tests
+
+describe('dropLoadingAttribute', () => {
```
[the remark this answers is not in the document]
ninadbstack: Sure! i thought it should be the case, but only file i referred to inject polyfill tests and it didn't have it [ checked that files just cause it was small ]

packages/dom/test/transform-dom.test.js
ninadbstack: Sure, added!

itsjwala (approved):

samarsault (approved):
LGTM 🎉

PR #1270, opened 2023-05-26
Feat: Send domInfoSha & PoA optimizations

nilshah98:
Adding Optimized Integrations with automate for PoA
Adding Automate token handling
Adding support for PoA


packages/webdriver-utils/src/util/timing.js
```diff
@@ -0,0 +1,54 @@
+import validations from './validations.js';
```
itsjwala, line 1: have it in `sdk-utils`

packages/cli-exec/src/exec.js
```diff
@@ -110,4 +110,41 @@ async function* spawn(cmd, args) {
   }
 }
 
+const autoStart = command('start', {
```
itsjwala, line 113: we don't need to add a new command here. 

in the callback for  `commad('exec',...)` we'll update the value of 

`percy.projectType, percy.skipDiscovery` based on the token prefix ( have this token logic in package which is required by all `cli-*` commands since we'll need a check in future in those packages.

packages/webdriver-utils/src/providers/automateProvider.js
```diff
@@ -1,11 +1,121 @@
...
+    }
+    return response;
+  }
+
+  async percyScreenshotBegin(name) {
+    return await TimeIt.run('percyScreenshotBegin', async () => {
+      try {
+        let result = await this.browserstackExecutor('percyScreenshot', {
+          name,
+          percyBuildId: process.env.PERCY_BUILD_ID,
```
Amit3200, line 55: this we will change as cli v 1.25.0 had a feature implemented

packages/webdriver-utils/src/util/timing.js
Amit3200: Tried moving basically the import statement breaks because of the nature of `validations.js` file.

packages/cli-exec/src/exec.js
Amit3200: Changed this as suggested.

packages/client/src/client.js
```diff
@@ -366,9 +366,9 @@ export class PercyClient {
     return snapshot;
   }
 
-  async createComparison(snapshotId, { tag, tiles = [], externalDebugUrl, ignoredElementsData } = {}) {
+  async createComparison(snapshotId, { tag, tiles = [], externalDebugUrl, ignoredElementsData, domSha } = {}) {
```
Amit3200, line 369: @itsjwala  I have tried changing the `domSha` to `DOMSha` or `webDOMSha` js is converting it into `domsHa` or `webDomsHa` automatically. This breaks the E2E flow. I guess all the caps will need to be written at last.

itsjwala: yes, there is a config for this check - https://github.com/percy/cli/blob/66b8192c0c24a5e16e3477d1701a201ccede4838/packages/config/src/utils/normalize.js#L5-L9 

we've a mapping which automatically updates it to given format

Amit3200: I tried adding `domSha` -> `DOMSha`, it was failing.

itsjwala: lets have this `domInfoSha`

packages/cli-exec/src/exec.js
```diff
@@ -59,6 +59,8 @@ export const exec = command('exec', {
     log.warn('Percy is disabled');
   } else {
     try {
+      percy.projectType = percy?.tokenType();
+      percy.skipDiscovery = percy.projectType === 'web';
```
itsjwala, line 63:
```suggestion
      percy.projectType = percy.tokenType();
      percy.skipDiscovery = percy.projectType !== 'web';
```

packages/core/src/config.js
```diff
@@ -397,6 +397,7 @@ export const comparisonSchema = {
   properties: {
     name: { type: 'string' },
     externalDebugUrl: { type: 'string' },
+    domSha: { type: 'string' },
```
itsjwala, line 400:
```suggestion
    domInfoSha: { type: 'string' },
```

packages/core/src/percy.js
```diff
@@ -360,6 +360,25 @@ export class Percy {
       }
     }.call(this));
   }
+
+  // decides project type
```
itsjwala, line 364: lets move to `packages/client/src/client.js`

packages/core/src/percy.js
```diff
@@ -360,6 +360,25 @@ export class Percy {
...
+  tokenType() {
+    if (!process.env.PERCY_TOKEN || typeof process.env.PERCY_TOKEN !== 'string') { throw new Error('Missing Percy token'); }
+
+    const type = process.env.PERCY_TOKEN?.split('_')[0];
```
itsjwala, line 368:
```suggestion
```

directly use token getToken()

itsjwala: @Amit3200 lets use this inside `getTokenType`

packages/webdriver-utils/src/util/timing.js
itsjwala: ack, lets check this next time.

packages/webdriver-utils/src/util/cache.js
```diff
@@ -1,6 +1,6 @@
 
-import validations from './validations.js';
-const { Undefined } = validations;
+import utils from '@percy/sdk-utils';
+const Undefined = utils.Undefined;
```
itsjwala, line 3:
```suggestion
import {Undefined} from '@percy/sdk-utils';
```

packages/webdriver-utils/src/providers/automateProvider.js
```diff
@@ -1,11 +1,127 @@
+import utils from '@percy/sdk-utils';
```
itsjwala, line 1: lets import individual functions here which is required.

```
import { TimeIt } from '@percy/sdk-utils';
```

Amit3200: This needs change in `package.json`, we can pick this as in separate task to reconcile this.

packages/webdriver-utils/src/util/cache.js
Amit3200: This needs change in `package.json`, we can pick this as in separate task to reconcile this.

packages/webdriver-utils/src/providers/automateProvider.js
itsjwala: try removing `type: "module"`  does it help or is it a bigger change?

Amit3200: There's not type mentioned in `sdk-utils` and adding there works for the webdriver utils but the sdk-util import statement will start breaking.

packages/webdriver-utils/src/util/cache.js
itsjwala: as discussed

```
const {Undefined} = utils;
```

packages/cli-exec/src/exec.js
```diff
@@ -59,6 +59,8 @@ export const exec = command('exec', {
     log.warn('Percy is disabled');
   } else {
     try {
+      percy.projectType = percy.client.tokenType();
```
itsjwala, line 62: make behavior same for existing users, skip `missing token` check for all when in `--debug` or `--testing` mode ( caveat automate will have asset discovery on, which is fine for now)

more discussion happening internally.

packages/client/src/client.js
```diff
@@ -526,6 +529,12 @@ export class PercyClient {
         return 'web';
     }
   }
+
+  computeSkipDiscoveryState() {
+    const type = this.tokenType();
```
itsjwala, line 534: lets pass an argument to `tokenType` which returns token without raising error, by default it will work how it is working currently.

itsjwala: also lets move this function to `core` since `skipDiscovery` is part of core

and rename it `shouldSkipAssetDiscovery()`

packages/client/src/client.js
```diff
@@ -42,19 +42,22 @@ export class PercyClient {
...
     clientInfo,
     environmentInfo,
+    testing,
     // versioned api url
     apiUrl = PERCY_CLIENT_API_URL
   } = {}) {
     Object.assign(this, { token, apiUrl });
     this.addClientInfo(clientInfo);
     this.addEnvironmentInfo(environmentInfo);
+    this.testing = testing || false;
```
itsjwala, line 60: * this won't be required, below comment is resolved ( code moved to `core` package)

packages/client/src/client.js
```diff
@@ -504,6 +507,25 @@ export class PercyClient {
     await this.finalizeComparison(comparison.data.id);
     return comparison;
   }
+
+  // decides project type
+  tokenType() {
+    let token = this.token || this.env.token || '';
```
itsjwala, line 513:
```suggestion
    let token = this.token || this.env.token || '';
```

lets call `getToken` function, with an argument `{ raiseIfMissing: false}` ( default would be true)
so we still use the function to fetch token.

packages/core/src/percy.js
```diff
@@ -360,6 +360,11 @@ export class Percy {
       }
     }.call(this));
   }
+
+  shouldSkipAssetDiscovery(tokenType) {
+    if (this.testing && JSON.stringify(this.testing) === JSON.stringify({})) { return true; }
```
itsjwala, line 365: * is `testing` ever an empty `{}` dictionary? - that's weird

Amit3200: Yes refer to following here, not changing behaviour as existing flow will need change and tests fails, we can change as part of later though.
https://github.com/percy/cli/blob/9c89383d6e1a9aa41f0c83f8e4b9ddcdf827583b/packages/core/src/percy.js#L83

packages/client/src/client.js
Amit3200: `getToken()` is already a defined function. If you meant passing in `tokenType` a variable even then it won't solve much issue as we would end up writing if/else logic only.

itsjwala: what I meant is

`getToken({{ raiseIfMissing: true}})` this would be by default definition of the function, so it won't affect existing users
and wrap the `raise error` with this condition.

in `tokenType()` function you'll call `getToken({raiseIfMissing: false})` so even if token is missing it won't raise an error.

itsjwala: ack, lets update spec for `getToken`

PR #1275, opened 2023-06-06
✨ Introduce new config option for cli-snapshot for enabling JS in asset discovery page

itsjwala:
* we now introduce a new config alongside the existing `enableJavaScript` called `cliEnableJavaScript`
* purpose for this new `cliEnableJavaScript`  flag is we'll use it to override Asset discovery browser page JS before taking dom Snapshot
  * since we take dom snapshot in CLI for only `cli-snapshot` ( or `percy snapshot` command) this doesn't affect `percySnapshot` ( or `percy exec --`) flow
 * This change is backwards compatible and doesn't affect existing users.
 * Tests are updated for a better understanding of the expectations.

### DIFF
--- a/packages/config/test/index.test.js
+++ b/packages/config/test/index.test.js
@@ -1101,7 +1101,8 @@ describe('PercyConfig', () => {
         'Bar BAZ qux': 'xyzzy',
         'percy-css': '',
         'enable-javascript': false,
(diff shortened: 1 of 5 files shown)


ninadbstack (approved):
LGTM!

PR #1279, opened 2023-06-16
Add ability to pass in `page` to `getBuilds` method

mikedidomizio:
This PR adds the ability to pass in `page` params into the `getBuilds` method.  Before the method only supported `filter`.

I've added it as a third argument because I did not want to break existing functionality.

I've tested this on my own project and works as expected.

### DIFF
--- a/packages/client/README.md
+++ b/packages/client/README.md
@@ -131,6 +131,9 @@ await client.getBuilds(projectSlug)
 
 // get all builds for a project's "master" branch
 await client.getBuilds(projectSlug, { branch: 'master' })
+
+// get all builds for a project with limit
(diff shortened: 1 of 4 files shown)


mikedidomizio:
Regarding the work here, if it's not needed feel free to close it or let me know.  I worked around it via direct calls to the Percy API.  It might be helpful to others but it's no sweat off my back if it doesn't get merged.

itsjwala:
Hey @mikedidomizio 👋 

* Thanks for your contribution, we've been busy with some releases lately.
* Regarding your PR, we'll hold on to this currently since we don't have any internal requirements for these APIs
* As you correctly mentioned, users can use this API on their own using `@percy/client` 

Closing this PR, we can circle back on this when required!

PR #1281, opened 2023-06-19
Adding percyCSS options for PoA [Snapshot level]

Amit3200:
Adding percyCSS option for PoA.
This will allow to have user controlled stabilisations while taking screenshots.
The property will be applied and once screenshot is taken it will be removed.

### DIFF
--- a/packages/webdriver-utils/src/providers/genericProvider.js
+++ b/packages/webdriver-utils/src/providers/genericProvider.js
@@ -53,12 +53,29 @@ export default class GenericProvider {
     }
   }
 
+  async addPercyCSS(userCSS) {
+    const createStyleElement = `const e = document.createElement('style');
+      e.setAttribute('class', 'poa-user-css-injected');
+      e.innerHTML = '${userCSS}';
+      document.head.appendChild(e);`;
+    await this.driver.executeScript({ script: createStyleElement, args: [] });
+  }
+
+  async removePercyCSS() {
+    const removeStyleElement = `const n = document.querySelectorAll('.poa-user-css-injected');
+      n.forEach((e) => {e.remove()});`;
+    await this.driver.executeScript({ script: removeStyleElement, args: [] });
+  }
+
   async screenshot(name) {
     let fullscreen = false;
 
+    const percyCSS = this.options.percyCSS || '';
+    await this.addPercyCSS(percyCSS);
     const tag = await this.getTag();
     const tiles = await this.getTiles(fullscreen);
     await this.setDebugUrl();
+    await this.removePercyCSS();
 
     log.debug(`${name} : Tag ${JSON.stringify(tag)}`);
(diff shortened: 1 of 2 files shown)


packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -53,12 +53,29 @@ export default class GenericProvider {
...
+  async addPercyCSS(userCSS) {
+    const createStyleElement = `const e = document.createElement('style');
+      e.setAttribute('class', 'poa-user-css-injected');
+      e.innerHTML = '${userCSS}';
+      document.head.appendChild(e);`;
+    await this.driver.executeScript({ script: createStyleElement, args: [] });
+  }
+
+  async removePercyCSS() {
+    const removeStyleElement = `const n = document.querySelectorAll('.poa-user-css-injected');
```
Amit3200, line 65: change this to `attribute` or look at the current `percy injected css`

Amit3200: query using `document.querySelector(‘[data-percy-css]’)`

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -53,12 +53,29 @@ export default class GenericProvider {
     }
   }
 
+  async addPercyCSS(userCSS) {
+    const createStyleElement = `const e = document.createElement('style');
+      e.setAttribute('class', 'poa-user-css-injected');
+      e.innerHTML = '${userCSS}';
+      document.head.appendChild(e);`;
```
Amit3200, line 60: `append at body`

packages/webdriver-utils/test/providers/genericProvider.test.js
```diff
@@ -115,4 +121,30 @@ describe('GenericProvider', () => {
...
+  describe('addPercyCSS', () => {
+    beforeEach(() => {
+      spyOn(Driver.prototype, 'executeScript').and.returnValue(Promise.resolve(true));
+    });
+
+    it('should call executeScript to add style', async () => {
+      genericProvider = new GenericProvider('123', 'http:executorUrl', { platform: 'win' }, {}, 'local-poc-poa', 'staging-poc-poa', {});
+      await genericProvider.createDriver();
+      await genericProvider.addPercyCSS('h1{color:green !important;}');
+      expect(genericProvider.driver.executeScript).toHaveBeenCalledTimes(1);
```
itsjwala, line 134: Have expectation to have been called with and pass arguments as well. So in future if this method changed due to package update we fail the tests

Amit3200: This function of `executeScript` has its own test and is being tested that way.

itsjwala (approved):
lgtm, some tests are failing

PR #1283, opened 2023-06-23
✨ Update proxy.js to auto strip double quotes/spaces if any

shantanuk-browserstack:
Added a function `stripQuotesAndSpaces` in `env/utils.js` that removes double quotes or spaces present at starting or ending of a line.
Used it in `client/src/proxy.js` where environment variable are stored in `proxyURL` & `shouldProxy`

### DIFF
--- a/packages/client/src/proxy.js
+++ b/packages/client/src/proxy.js
@@ -63,12 +63,15 @@ export function href(options) {
 // Returns the proxy URL for a set of request options
 export function getProxy(options) {
   let proxyUrl = (options.protocol === 'https:' &&
-    (stripQuotesAndSpaces(process.env.https_proxy) || stripQuotesAndSpaces(process.env.HTTPS_PROXY))) ||
-    (stripQuotesAndSpaces(process.env.http_proxy) || stripQuotesAndSpaces(process.env.HTTP_PROXY));
+    (process.env.https_proxy || process.env.HTTPS_PROXY)) ||
+    (process.env.http_proxy || process.env.HTTP_PROXY);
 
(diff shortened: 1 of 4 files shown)


packages/client/src/proxy.js
```diff
@@ -59,14 +59,22 @@ export function href(options) {
     (path || `${pathname || ''}${search || ''}${hash || ''}`);
 };
 
+// auto strip double quotes/spaces if any
+function stripQuotesAndSpaces(line) {
```
ninadbstack, line 63: some linting issues, but we will also need to do it for other env vars, so maybe move this function to util package

packages/client/src/proxy.js
```diff
@@ -62,11 +63,11 @@ export function href(options) {
 // Returns the proxy URL for a set of request options
 export function getProxy(options) {
   let proxyUrl = (options.protocol === 'https:' &&
-    (process.env.https_proxy || process.env.HTTPS_PROXY)) ||
-    (process.env.http_proxy || process.env.HTTP_PROXY);
+    (stripQuotesAndSpaces(process.env.https_proxy) || stripQuotesAndSpaces(process.env.HTTPS_PROXY))) ||
+    (stripQuotesAndSpaces(process.env.http_proxy) || stripQuotesAndSpaces(process.env.HTTP_PROXY));
```
ninadbstack, line 67: isnt it better to just add a new line below that does 
```
proxyUrl = stripQuotesAndSpaces(proxyUrl)
```

itsjwala:
RUN_REGRESSION

ninadbstack (approved):

PR #1288, opened 2023-06-27
[PER-2063] Percy upload for BYOS

shahidk8:
- Change logic of percy upload based on type of project.
- Allow only Web and Generic project type token for upload.

### DIFF
--- a/packages/cli-upload/src/upload.js
+++ b/packages/cli-upload/src/upload.js
@@ -2,9 +2,14 @@ import fs from 'fs';
 import path from 'path';
 import command from '@percy/cli-command';
 import * as UploadConfig from './config.js';
+import PercyClient from '@percy/client';
 
 const ALLOWED_FILE_TYPES = /\.(png|jpg|jpeg)$/i;
 
+function tokenType() {
+  const percyClient = new PercyClient({ token: null, clientInfo: null, environmentInfo: null });
+  return percyClient.tokenType();
+}
 export const upload = command('upload', {
   description: 'Upload a directory of images to Percy',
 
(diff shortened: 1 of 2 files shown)


packages/client/src/client.js
```diff
@@ -504,6 +504,24 @@ export class PercyClient {
...
+      case 'web':
+        return 'web';
+      case 'app':
+        return 'app';
+      case 'ss':
+        return 'generic';
+      default:
+        return 'web';
+    }
+  }
```
samarsault, line 524: Don't add this here, rebase your branch on top of https://github.com/percy/cli/pull/1270/

packages/cli-upload/package.json
```diff
@@ -34,6 +34,7 @@
   },
   "dependencies": {
     "@percy/cli-command": "1.26.1",
+    "@percy/client": "^1.26.1",
```
samarsault, line 37: Don't need this separately

packages/cli-upload/test/upload.test.js
```diff
@@ -199,4 +200,50 @@ describe('percy upload', () => {
...
+            }])
+          }
+        }
+      }
+    });
+  });
+
+  it('throws error for token type other than web and generic', async () => {
+    process.env.PERCY_TOKEN = 'app_invalid_token';
+    await expectAsync(upload(['./images'])).toBeRejected();
```
itsjwala, line 247:
```suggestion
    await expectAsync(upload(['./images'])).toBeRejected();
```

add rejected with error message expectation.

packages/cli-upload/src/upload.js
```diff
@@ -87,15 +96,28 @@ export const upload = command('upload', {
...
+      } else if (tokenType === 'generic') {
+        percy.upload({
+          name: snapshotName,
+          tag: BYOS_TAG,
+          tiles: [
+            { filepath: img.absolutePath }
+          ]
+        });
+      } else {
+        throw new Error('Invalid Token Type');
```
itsjwala, line 119:
```suggestion
        throw new Error('Invalid Token Type, only web, self managed token type are allowed.');
```

lets ask internally about the wordings here.

packages/cli-upload/test/upload.test.js
samarsault: added

packages/cli-upload/src/upload.js
```diff
@@ -87,6 +88,10 @@ export const upload = command('upload', {
 
   const tokenType = percy.client.tokenType();
 
+  if (!ALLOWED_TOKEN_TYPES.includes(tokenType)) {
+    exit(1, 'Invalid Token Type, only web, self managed token type are allowed.');
```
itsjwala: we don't exit here, we'll throw `error`, thats what current `exec` behaviour is.

we need to stop percy local server. so L130 should be executed somehow.

samarsault: Moved the check before starting percy (no matching files handled similarly)

packages/cli-upload/src/upload.js
```diff
@@ -4,6 +4,14 @@ import command from '@percy/cli-command';
 import * as UploadConfig from './config.js';
 
 const ALLOWED_FILE_TYPES = /\.(png|jpg|jpeg)$/i;
+const ALLOWED_TOKEN_TYPES = ['web', 'generic'];
+
+// All BYOS screenshots have a fixed comparison tag
+export const BYOS_TAG = {
+  name: 'BYOS',
```
itsjwala, line 11:
```suggestion
  name: 'BYOS',
```

lets discuss name field here.

itsjwala (approved):

PR #1292, opened 2023-06-29
✨ Percy on Automate Support

Amit3200:
Percy on Automate Support & Changes, This branch will allow integrating browserstack automate tests with the visual tests. This will unlock visual tests on various combination of OS, browsers and devices. Now one can run their functional tests along with visual tests on any `Mac`, `Windows` versions*, not just that now one can also run their functional tests with visual tests on the various mobile devices `android` and `ios` accessible through automate.

This all can be achieved by adding a simple command - `percyScreenshot(driver, name)` in the script.

Refer to more details here - https://docs.percy.io/docs/integrate-functional-testing-with-visual-testing

`* supported Versions`


packages/webdriver-utils/package.json
```diff
@@ -7,6 +7,10 @@
     "url": "https://github.com/percy/cli",
     "directory": "packages/webdriver-utils"
   },
+  "publishConfig": {
+    "access": "public",
+    "tag": "latest"
```
itsjwala, line 12: lets change all the `tag` to `alpha` ( even though we'll specifiy `--dist-tag` while publishing, this is for sanity)

.github/workflows/test.yml
```diff
@@ -55,6 +55,7 @@ jobs:
           - '@percy/cli-build'
           - '@percy/cli-config'
           - '@percy/sdk-utils'
+          - '@percy/webdriver-utils'
```
itsjwala, line 58: need to add same for windows

karma.config.cjs
```diff
@@ -23,7 +23,9 @@ module.exports = async config => {
       { pattern: 'test/**/*.test.js', type: 'module', watched: false },
       { pattern: 'test/assets/**', watched: false, included: false }
     ],
-
+    exclude: [
```
itsjwala, line 26: why is this removed?

Amit3200: There was a test I guess thats why @nilshah98  had to remove it.

nilshah98: > First sdk-utils/request , here we have added support for fully qualified urls and also for https server.

> Hence, to support that while testing, created a local instance of https server and wrote tests accordingly, which passed.

> The current issue is, that sdk-utils also run their tests on browsers, where it tries to create this https server and fails.

> As a workaround I have added test/request.test.js in the exclude parameter in karma.config.js

itsjwala:
RUN_REGRESSION

packages/sdk-utils/test/request.test.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+  };
+}
+
+describe('Utils Requests', () => {
+  let server;
+
+  // Adding below env variables to support self signed certs
+  beforeAll(() => {
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
itsjwala, line 85: since anyway we're doing this, we can pass rogue certs no need to have a file committed ( possible to mock this part?)

packages/core/src/api.js
```diff
@@ -3,7 +3,9 @@ import path from 'path';
 import { createRequire } from 'module';
 import logger from '@percy/logger';
 import { normalize } from '@percy/config/utils';
-import { getPackageJSON, Server } from './utils.js';
+import { getPackageJSON, Server, percyAutomateRequestHandler } from './utils.js';
+// TODO Remove below esline disable once we publish webdriver-util
```
itsjwala, line 7:
```suggestion
```

packages/core/src/api.js
```diff
@@ -3,7 +3,9 @@ import path from 'path';
...
+import WebdriverUtils from '@percy/webdriver-utils'; // eslint-disable-line import/no-extraneous-dependencies
```
itsjwala, line 8:
```suggestion
import WebdriverUtils from '@percy/webdriver-utils';
```

itsjwala (commented):
fix all the TODO comments

karma.config.cjs
itsjwala: I think it would be due to the SSL error it's breaking, lets try fixing this. Excluding it here doesn't seem clean. lets sync on this.

packages/webdriver-utils/src/metadata/desktopMetaData.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  // showing major version
+  osVersion() {
+    return this.capabilities.osVersion.toLowerCase();
+  }
+
+  // combination of browserName + browserVersion + osVersion + osName
+  deviceName() {
+    return this.browserName() + '_' + this.browserVersion() + '_' + this.osVersion() + '_' + this.osName();
```
Amit3200, line 30: Lets change this to:
`this.osName() + '_' + this.osVersion() + '_' + this.browserName() + '_' + this.browserVersion();`

packages/sdk-utils/src/post-screenshot.js
```diff
@@ -0,0 +1,18 @@
+import percy from './percy-info.js';
```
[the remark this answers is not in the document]
itsjwala: This would mean, all the JS based SDKs will now use `webdriver-utils` as dependency instead of `sdk-utils`

packages/core/src/api.js
itsjwala: add `webdriver-utils` as dependency of `core` package

packages/sdk-utils/src/post-screenshot.js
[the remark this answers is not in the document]
itsjwala: its fine to have it here as `post-snapshot` is present in same directory.

karma.config.cjs
itsjwala: For now add a comment here and request.test.js why we're skipping it.

nilshah98: Done 👍

packages/sdk-utils/test/request.test.js
nilshah98: Tried with just `abcd` the server creation fails in `request.test.js`.
Using self signed certs.

itsjwala:
lets update the description with documentation and maybe landing page so it looks neat.

packages/cli-app/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@percy/cli-app",
-  "version": "1.26.3",
+  "version": "1.27.0-beta.1",
```
itsjwala, line 3: bump version in separate PR (check master merge commits to understand better)

itsjwala: publishConfig { tag : } needs to be `beta`

itsjwala (approved):

PR #1296, opened 2023-06-30
Added env var support for timeouts

ninadbstack:
Added support for 2 flags as follows
1. `PERCY_NETWORK_IDLE_WAIT_TIMEOUT`: this flag is used to configure maximum time that percy cli would wait for network idle timeout to hit. i.e. if `networkIdleTimeout` is 600ms we would wait at most PERCY_NETWORK_IDLE_WAIT_TIMEOUT time to get a successful `networkIdleTimeout`.
2. `PERCY_PAGE_LOAD_TIMEOUT`: This is timeout for page_load event to fire on the page when we load page in the chromium.

Both of above are defaulted to 30000ms [ 30 sec ] and both would print a warning once per session when its set to over 60000ms [ 60 sec ].

We are adding this functionality so that in rare cases when someone wants to confirm that increasing timeout will solve the asset discovery issues for the page, they can confirm it by increasing the time.

### DIFF
--- a/packages/core/src/network.js
+++ b/packages/core/src/network.js
@@ -10,7 +10,7 @@ const ALLOWED_RESOURCES = ['Document', 'Stylesheet', 'Image', 'Media', 'Font', '
 // The Interceptor class creates common handlers for dealing with intercepting asset requests
 // for a given page using various devtools protocol events and commands.
 export class Network {
-  static TIMEOUT = 30000;
+  static TIMEOUT = undefined;
 
   log = logger('core:discovery');
 
@@ -29,6 +29,7 @@ export class Network {
       page.session.browser.version.userAgent.replace('Headless', '');
     this.intercept = options.intercept;
     this.meta = options.meta;
+    this._initializeNetworkIdleWaitTimeout();
   }
 
   watch(session) {
@@ -236,6 +237,18 @@ export class Network {
     }
 
     this._forgetRequest(request);
+  }
+
+  _initializeNetworkIdleWaitTimeout() {
+    if (Network.TIMEOUT) return;
+
+    Network.TIMEOUT = parseInt(process.env.PERCY_NETWORK_IDLE_WAIT_TIMEOUT) || 30000;
+
+    if (Network.TIMEOUT > 60000) {
+      this.log.warn('Setting PERCY_NETWORK_IDLE_WAIT_TIMEOUT over 60000ms is not recommended. ' +
+        'If your page needs more than 60000ms to idle due to CPU/Network load, ' +
+        'its recommended to increase CI resources where this cli is running.');
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/core/test/discovery.test.js
```diff
@@ -837,15 +838,28 @@ describe('Discovery', () => {
...
+
+    it('shows a warning when idle wait timeout is set over 60000ms', async () => {
+      process.env.PERCY_NETWORK_IDLE_WAIT_TIMEOUT = 80000;
+
+      await percy.snapshot({
+        name: 'test idle',
+        url: 'http://localhost:8000'
+      });
+
+      expect(logger.stderr).toContain(jasmine.stringMatching(
```
itsjwala, line 850: can you add expectation with for this line as well?

https://github.com/percy/cli/blob/c12668d70252186f061399f00f4f2aebbcc06d8c/packages/core/src/network.js#L66

packages/core/test/discovery.test.js
```diff
@@ -874,6 +889,20 @@ describe('Discovery', () => {
...
+    it('shows a warning when page load timeout is set over 60000ms', async () => {
+      percy.loglevel('debug');
+      process.env.PERCY_PAGE_LOAD_TIMEOUT = 80000;
+
+      await percy.snapshot({
+        name: 'navigation idle',
+        url: 'http://localhost:8000'
+      });
+
+      expect(logger.stderr).toContain(jasmine.stringMatching(
```
itsjwala, line 902: same comment as previous one

packages/core/src/network.js
```diff
@@ -10,7 +10,7 @@ const ALLOWED_RESOURCES = ['Document', 'Stylesheet', 'Image', 'Media', 'Font', '
 // The Interceptor class creates common handlers for dealing with intercepting asset requests
 // for a given page using various devtools protocol events and commands.
 export class Network {
-  static TIMEOUT = 30000;
+  static TIMEOUT = undefined;
```
itsjwala, line 13: we have something for `networkIdleTimeout` which can have max value of `750ms` which can be provided by users. 

https://github.com/percy/cli/blob/c12668d70252186f061399f00f4f2aebbcc06d8c/packages/core/src/network.js#L75-L78

itsjwala: so we'll need to verify once for heavy pages, is `TIMEOUT` honoured over `networkIdleTimeout`

ninadbstack: networkIdleTimeout is different, is the event, this TIMEOUT is amount of time we will wait for `networkIdleTimeout` to trigger. 

i.e. by default wait for 30 sec for networkIdleTimeout to hit, which will only be hit if network is idle for 750ms

packages/core/test/discovery.test.js
ninadbstack: I can add it, but its not related to current test though, this line is related to networkIdleTimeout

itsjwala (approved):

samarsault:
RUN_REGRESSION

PR #1299, opened 2023-07-05
[PER-2252] Blob assets not being serialized by Percy

shantanuk-browserstack:
Code Changes to capture the Blob CSS assets.

### DIFF
--- a/packages/dom/src/serialize-cssom.js
+++ b/packages/dom/src/serialize-cssom.js
@@ -62,6 +62,32 @@ export function serializeCSSOM({ dom, clone, resources, cache }) {
       clone.prepend(styleLink);
     }
   }
+
+  // Capture blob css assets
+  for (let styleSheet of dom.styleSheets) {
+    if (styleSheet instanceof CSSStyleSheet && styleSheet.href && styleSheet.href.startsWith("blob:")) {
+      const styleLink = document.createElement('link');
+      styleLink.setAttribute('rel', 'stylesheet');
+
+      if (!cache.has(styleSheet)) {
+        const styles = Array.from(styleSheet.cssRules)
+          .map(cssRule => cssRule.cssText).join('\n');
+        let resource = resourceFromText(uid(), 'text/css', styles);
+        resources.add(resource);
+        cache.set(styleSheet, resource.url);
+      }
+      styleLink.setAttribute('data-percy-adopted-stylesheets-serialized', 'true');
+      styleLink.setAttribute('data-percy-serialized-attribute-href', cache.get(styleSheet));
+
+      /* istanbul ignore next: tested, but coverage is stripped */
+      if (clone.constructor.name === 'HTMLDocument' || clone.constructor.name === 'DocumentFragment') {
+        // handle document and iframe
+        clone.body.prepend(styleLink);
+      } else if (clone.constructor.name === 'ShadowRoot') {
+        clone.prepend(styleLink);
+      }
+    }
+  }
 }
 
 export default serializeCSSOM;
--- a/packages/dom/src/serialize-cssom.js
+++ b/packages/dom/src/serialize-cssom.js
@@ -18,6 +18,14 @@ function styleSheetsMatch(sheetA, sheetB) {
   return true;
 }
 
+function createStyleResource(styleSheet, resources) {
+  const styles = Array.from(styleSheet.cssRules)
+    .map(cssRule => cssRule.cssText).join('\n');
+  let resource = resourceFromText(uid(), 'text/css', styles);
+  resources.add(resource);
+  return resource;
+}
+
 export function serializeCSSOM({ dom, clone, resources, cache }) {
   // in-memory CSSOM into their respective DOM nodes.
   for (let styleSheet of dom.styleSheets) {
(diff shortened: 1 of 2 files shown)


packages/dom/src/serialize-cssom.js
```diff
@@ -62,6 +62,33 @@ export function serializeCSSOM({ dom, clone, resources, cache }) {
       clone.prepend(styleLink);
     }
   }
+
+  // Capture blob css assets
+  for (let styleSheet of dom.styleSheets) {
```
samarsault, line 67: We can do this in above loop for in memory CSSOM, no need to loop multiple times

packages/dom/src/serialize-cssom.js
```diff
@@ -62,6 +62,33 @@ export function serializeCSSOM({ dom, clone, resources, cache }) {
...
+      styleLink.setAttribute('rel', 'stylesheet');
+      const styles = Array.from(styleSheet.cssRules)
+        .map(cssRule => cssRule.cssText).join('\n');
+
+      if (cache.has(styles)) { continue; }
+
+      let resource = resourceFromText(uid(), 'text/css', styles);
+      resources.add(resource);
+      cache.set(styles, resource.url);
+
```
samarsault, line 79: Can refactor this to a function `createStyleResource`

packages/dom/test/serialize-css.test.js
```diff
@@ -158,5 +158,58 @@ describe('serializeCSSOM', () => {
...
+        return;
+      }
+      withExample('<div>BlobStyle</div>', { withShadow: false });
+
+      function generateBlobUrl(cssStyle) {
+        const blob = new window.Blob([cssStyle], { type: 'text/css' });
+        return window.URL.createObjectURL(blob);
+      }
+
+      function createLinkElement(blobURL) {
```
samarsault, line 173: Lets rename it `createStyleLinkElement`, `createLinkElement` suggests it might be a hyperlink

packages/dom/test/serialize-css.test.js
```diff
@@ -158,5 +158,58 @@ describe('serializeCSSOM', () => {
...
+        const capture = serializeDOM();
+        let $ = parseDOM(capture, 'plain');
+        expect($('body')[0].innerHTML).toMatch(`<link rel="stylesheet" data-percy-blob-stylesheets-serialized="true" href="${capture.resources[0].url}">`);
+        dom.head.removeChild(linkElement1);
+        dom.head.removeChild(linkElement2);
+        dom.head.removeChild(linkElement3);
+        URL.revokeObjectURL(blobUrl1);
+        URL.revokeObjectURL(blobUrl2);
+      }
+    });
```
samarsault, line 213: Is it possible to not use event listeners for this? Instead, we can simply assert that the blob assets are present in the end

packages/dom/src/serialize-cssom.js
```diff
@@ -35,6 +43,21 @@ export function serializeCSSOM({ dom, clone, resources, cache }) {
 
       cloneOwnerNode.parentNode.insertBefore(style, cloneOwnerNode.nextSibling);
       cloneOwnerNode.remove();
+    } else if (styleSheet instanceof window.CSSStyleSheet && styleSheet.href && styleSheet.href.startsWith('blob:')) {
```
samarsault, line 46: Why do we need `styleSheet instanceof window.CSSStyleSheet` check here? Isn't every stylesheet in `dom.styleSheets` of this type?

packages/dom/test/serialize-css.test.js
```diff
@@ -158,5 +159,50 @@ describe('serializeCSSOM', () => {
...
+
+      dom.head.appendChild(linkElement1);
+      dom.head.appendChild(linkElement2);
+      dom.head.appendChild(linkElement3);
+
+      await when(() => {
+        const capture = serializeDOM();
+        let $ = parseDOM(capture, 'plain');
+        expect($('body')[0].innerHTML).toMatch(`<link rel="stylesheet" data-percy-blob-stylesheets-serialized="true" href="${capture.resources[0].url}">`);
+      }, 5000);
```
samarsault, line 200: Move the assertion outside of `when`, use `when` only to wait for all styles to load

```javascript
await when( () => everything loaded, 5000)
const capture = serializeDOM()
```

packages/dom/src/serialize-cssom.js
```diff
@@ -35,6 +42,22 @@ export function serializeCSSOM({ dom, clone, resources, cache }) {
 
       cloneOwnerNode.parentNode.insertBefore(style, cloneOwnerNode.nextSibling);
       cloneOwnerNode.remove();
+    } else if (styleSheet.href && styleSheet.href.startsWith('blob:')) {
```
samarsault, line 45:
```suggestion
    } else if (styleSheet.href?.startsWith('blob:')) {
```

packages/dom/test/serialize-css.test.js
```diff
@@ -158,5 +159,52 @@ describe('serializeCSSOM', () => {
...
+      dom.head.appendChild(linkElement1);
+      dom.head.appendChild(linkElement2);
+      dom.head.appendChild(linkElement3);
+
+      await when(() => {
+        assert(dom.styleSheets.length === 3);
+      }, 5000);
+      const capture = serializeDOM();
+      let $ = parseDOM(capture, 'plain');
+      expect($('body')[0].innerHTML).toMatch(`<link rel="stylesheet" data-percy-blob-stylesheets-serialized="true" href="${capture.resources[0].url}">`);
```
samarsault, line 201: Also add expectation for the presence of the rest of the stylesheets

packages/dom/test/serialize-css.test.js
```diff
@@ -158,5 +159,54 @@ describe('serializeCSSOM', () => {
...
+      expect($('body')[0].innerHTML).toMatch(`<link rel="stylesheet" data-percy-blob-stylesheets-serialized="true" href="${capture.resources[1].url}">`);
+      expect($('body')[0].innerHTML).toMatch(`<link rel="stylesheet" data-percy-blob-stylesheets-serialized="true" href="${capture.resources[2].url}">`);
```
samarsault, line 203: We can merge this into a single expect

samarsault (approved):
LGTM

ninadbstack (approved):

Issue #1290, opened 2023-06-29
Incorrect export of core types


Techn1x:
## The problem

When I run `tsc --build` I get this error

```
tests/helpers/snapshot.ts:2:38 - error TS7016: Could not find a declaration file for module '@percy/core'. '/MY_APP_PATH/node_modules/.pnpm/@percy+core@1.26.1/node_modules/@percy/core/dist/index.js' implicitly has an 'any' type.
  There are types at '/MY_APP_PATH/node_modules/@percy/core/types/index.d.ts', but this result could not be resolved when respecting package.json "exports". The '@percy/core' library may need to update its package.json or typings.

2 import type { SnapshotOptions } from '@percy/core'
```

In other words, it knows types exist for this package, but due to how the package.json is setup it's not supposed to know :stuck_out_tongue: 

I think this is due to typescript updating how it expects types to be exported (in particular a config update here https://github.com/tsconfig/bases/pull/197 ) - it's to do with these lines
https://github.com/percy/cli/blob/9c89383d6e1a9aa41f0c83f8e4b9ddcdf827583b/packages/core/package.json#L24-L32

I believe the index types file should be listed under `exports` `"."` _as well as_ the `types` entry in package.json (and the `types` entry should be kept as fallback for older TS versions)
This stack overflow post explains it pretty well https://stackoverflow.com/a/76212193

---

The same issue is probably present for a handful of other packages in this monorepo that also export types, but I'm only using `core` so that's all I've encountered.

## Environment

- Node version: 18.6.0
- `@percy/cli` version: 1.26.1

Techn1x:
Thanks @itsjwala I have opened that PR here https://github.com/percy/cli/pull/1301

PR #1301, opened 2023-07-06
chore: fix percy types exports for newer typescript

Techn1x:
Fixes https://github.com/percy/cli/issues/1290

I have tested this via my own application, it resolves the error I was seeing detailed in the issue. I'm no expert at this but this makes sense to me.

Please let me know if there are other steps needed to have this released & published

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -24,7 +24,10 @@
   "types": "./types/index.d.ts",
   "type": "module",
   "exports": {
-    ".": "./dist/index.js",
+    ".": {
+      "types": "./types/index.d.ts",
+      "default": "./dist/index.js"
+    },
     "./utils": "./dist/utils.js",
     "./config": "./dist/config.js",
     "./install": "./dist/install.js",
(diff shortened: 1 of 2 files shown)


packages/core/package.json
```diff
@@ -24,7 +24,10 @@
   "types": "./types/index.d.ts",
```
Techn1x, line 24: the original "types" remains here for backwards compatibility reasons with older TS

itsjwala:
@Techn1x please make the changes in all the packages so it's uniform.

Techn1x:
@itsjwala which package are you referring to specifically? I already applied the change to all packages in this monorepo that have types. I checked again just now in case I missed one.

That test that failed seems unrelated? Flaky?

itsjwala (approved):
@Techn1x 

* Apologies I didn't see the PR fully - config and core packages only contain types so that should do
*  yup windows tests are flakey, it should be fine to merge without it being green.

thanks for your contribution 🎉

PR #1303, opened 2023-07-06
Adding Dpr fixes

Amit3200:
Get Resolution details for mobile + Enhancements (x DPR)

### DIFF
--- a/packages/webdriver-utils/src/metadata/mobileMetaData.js
+++ b/packages/webdriver-utils/src/metadata/mobileMetaData.js
@@ -43,6 +43,12 @@ export default class MobileMetaData {
     return { width, height };
   }
 
+  async screenResolution() {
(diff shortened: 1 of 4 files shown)


packages/webdriver-utils/src/metadata/mobileMetaData.js
```diff
@@ -43,6 +43,12 @@ export default class MobileMetaData {
     return { width, height };
   }
 
+  async screenResolution() {
```
itsjwala, line 46: Ideally we should move this to parent class ( currently there isn't one currently)

itsjwala (approved):

packages/webdriver-utils/src/metadata/mobileMetaData.js
Amit3200: Currently we don't have it we can create in upcoming times though.

PR #1305, opened 2023-07-07
feat: added support for extracting and passing header and footer heights

nilshah98:
**Context -**
- When capturing screenshots for mobile devices we have observed instances where we are also adding `nav` / `status` / `url` bars to the screenshot image.
- We want to remove these extra heights as they are **not** the part of the webpage.
- To remove these elements, we need some way to identify the device / platform / browser combination and subsequently find the heights for the sections to remove.

### DIFF
--- a/packages/webdriver-utils/src/providers/genericProvider.js
+++ b/packages/webdriver-utils/src/providers/genericProvider.js
@@ -174,7 +174,7 @@ export default class GenericProvider {
     const orientation = this.metaData.orientation();
     [this.header, this.footer] = await this.getHeaderFooter();
     // for android window size only constitutes of browser viewport, hence adding nav / status / url bar heights
-    height = this.capabilities.platformName && this.capabilities.platformName.toLowerCase() === 'android' ? height + this.header + this.footer : height;
+    height = this.metaData.osName() === 'android' ? height + this.header + this.footer : height;
     return {
       name: this.metaData.deviceName(),
       osName: this.metaData.osName(),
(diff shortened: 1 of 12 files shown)


packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -164,9 +169,12 @@ export default class GenericProvider {
 
   async getTag() {
     if (!this.driver) throw new Error('Driver is null, please initialize driver with createDriver().');
-    const { width, height } = await this.metaData.windowSize();
+    let { width, height } = await this.metaData.windowSize();
     const resolution = await this.metaData.screenResolution();
     const orientation = this.metaData.orientation();
+    [this.header, this.footer] = await this.getHeaderFooter();
+    // for android window size only constitutes of browser viewport, hence adding nav / status / url bar heights
+    height = this.capabilities.platformName && this.capabilities.platformName.toLowerCase() === 'android' ? height + this.header + this.footer : height;
```
Amit3200, line 177: Not all the devices will have this please confirm that this will not break for `ios` and `desktop`. Also do we not need to do this on `ios`?

Amit3200: Check `resolver` once where we check divide the flow between `mobileResolver` and `desktopResolver`, check that flow once as well.

nilshah98: Using `metaData.osName()` now.
verified that it is working as expected and is present for both desktop and mobile metadata resolvers.

Amit3200 (approved):

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -278,4 +286,16 @@ export default class GenericProvider {
     }
     return ignoredElementsArray;
   }
+
+  async getHeaderFooter() {
+    const devicesConfig = (await request(DEVICES_CONFIG_URL)).body;
```
itsjwala, line 291: lets call this once and keep it present in the local SDK server, this way per session this file would be cached.

itsjwala (approved):

Amit3200 (approved):

PR #1309, opened 2023-07-10
Make Freeze Animation false by default

Amit3200:
Making `freezeAnimation` false by default

### DIFF
--- a/packages/webdriver-utils/src/providers/genericProvider.js
+++ b/packages/webdriver-utils/src/providers/genericProvider.js
@@ -33,6 +33,36 @@ export default class GenericProvider {
     this.debugUrl = null;
   }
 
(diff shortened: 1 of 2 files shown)


packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -33,6 +33,36 @@ export default class GenericProvider {
     this.debugUrl = null;
   }
 
+  addDefaultOptions() {
+    this.options.freezeAnimation = this.options.freezeAnimation || false;
+  }
+
+  defaultPercyCSS() {
```
itsjwala, line 40: lets add tests, we don't want changes in this to go unnoticed.

itsjwala (approved):

PR #1311, opened 2023-07-12
Added condition to check if dom supports adoptedStyleSheets

shantanuk-browserstack:
adoptedStyleSheets is supported in Safari versions from v16.4 and in automate, Safari v16.3 is only available for the latest mac os.
So if anyone uses Safari from automate to run a test for percy, the test fails with the following error:
[percy] JavascriptError: A JavaScript exception occured: undefined is not an object (evaluating 'dom.adoptedStyleSheets')

Now a check is added to verify if adoptedStyleSheets is supported. If not supported, it will show a warning and continue processing.

### DIFF
--- a/packages/dom/src/serialize-cssom.js
+++ b/packages/dom/src/serialize-cssom.js
@@ -63,24 +63,27 @@ export function serializeCSSOM({ dom, clone, resources, cache }) {
 
   // clone Adopted Stylesheets
   // Regarding ordering of the adopted stylesheets - https://github.com/WICG/construct-stylesheets/issues/93
-  for (let sheet of dom.adoptedStyleSheets) {
-    const styleLink = document.createElement('link');
-    styleLink.setAttribute('rel', 'stylesheet');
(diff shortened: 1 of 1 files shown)


packages/dom/src/serialize-cssom.js
```diff
@@ -63,24 +63,27 @@ export function serializeCSSOM({ dom, clone, resources, cache }) {
 
   // clone Adopted Stylesheets
   // Regarding ordering of the adopted stylesheets - https://github.com/WICG/construct-stylesheets/issues/93
-  for (let sheet of dom.adoptedStyleSheets) {
-    const styleLink = document.createElement('link');
-    styleLink.setAttribute('rel', 'stylesheet');
+  /* istanbul ignore next: tested, but coverage is stripped */
+  if (dom.adoptedStyleSheets) {
```
ninadbstack, line 67: We also need a warning message here if this is being skipped

ninadbstack (approved):

PR #1321, opened 2023-07-19
Support Old App Percy Tokens

Amit3200:
Support Old App Percy Tokens along with PoA projects.
We have removed the check if the projectType is of `app` for dynamic fetching of projectType. We should re-enable it when we plan from moving away from `app:exec` to `exec` for app projects.

### DIFF
--- a/packages/cli-exec/src/exec.js
+++ b/packages/cli-exec/src/exec.js
@@ -59,8 +59,12 @@ export const exec = command('exec', {
     log.warn('Percy is disabled');
   } else {
     try {
-      percy.projectType = percy.client.tokenType();
-      percy.skipDiscovery = percy.shouldSkipAssetDiscovery(percy.projectType);
+      // Skip this for app because they are triggered as app:exec
+      // Remove this once they move to exec command as well
+      if (percy.projectType !== 'app') {
+        percy.projectType = percy.client.tokenType();
+        percy.skipDiscovery = percy.shouldSkipAssetDiscovery(percy.projectType);
+      }
       yield* percy.yield.start();
     } catch (error) {
       if (error.name === 'AbortError') throw error;
--- a/packages/cli-exec/src/exec.js
+++ b/packages/cli-exec/src/exec.js
@@ -62,9 +62,10 @@ export const exec = command('exec', {
       // Skip this for app because they are triggered as app:exec
       // Remove this once they move to exec command as well
       if (percy.projectType !== 'app') {
-        log.info('Percy project attribute calculation');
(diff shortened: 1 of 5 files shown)


packages/webdriver-utils/src/index.js
```diff
@@ -33,11 +33,16 @@ export default class WebdriverUtils {
...
+    try {
+      this.log.info('Starting automate screenshot ...');
+      const automate = ProviderResolver.resolve(this.sessionId, this.commandExecutorUrl, this.capabilities, this.sessionCapabilites, this.clientInfo, this.environmentInfo, this.options, this.buildInfo);
+      this.log.debug('Resolved provider ...');
+      await automate.createDriver();
+      this.log.debug('Created driver ...');
+      return await automate.screenshot(this.snapshotName, this.options);
+    } catch (e) {
+      this.log.error(`Error: ${e.message}`);
+      this.log.error(`Error Log: ${e.toString()}`);
```
ninadbstack, line 45: Add Snapshot name to the logs, so that its easy while Debugging

packages/cli-exec/src/exec.js
```diff
@@ -59,8 +59,12 @@ export const exec = command('exec', {
     log.warn('Percy is disabled');
   } else {
     try {
-      percy.projectType = percy.client.tokenType();
-      percy.skipDiscovery = percy.shouldSkipAssetDiscovery(percy.projectType);
+      // Skip this for app because they are triggered as app:exec
+      // Remove this once they move to exec command as well
+      if (percy.projectType !== 'app') {
```
itsjwala, line 64: This will always be true right?

projectType is defined (check above)

Can you explain when will it be false?

itsjwala, line 64: Also this is not correct way to handle this, we need to throw error and exit from here.

Ideally do at top of the function, and then run the usual flow.

Amit3200: Why we want to throw error?, if its `app:exec` the current flow should run as it is and no override should take place. Else override the projectType which we were doing before on the basis of tokenType.

itsjwala: Understood, lets update test to reflect this cross usage behaviour.

app:exec with old and new token

exec with old and new app token

Amit3200: @itsjwala  I tried `log.debug` however we don't log `log.debug` in any category neither in `stdout` nor in `stderr`. Hence I didn't have anything to match against expectation.

itsjwala (approved):
lgtm

PR #1322, opened 2023-07-19
🧹  Fix docs for override -> options

ninadbstack:
As per https://github.com/percy/cli/pull/819 `overrides` was renamed to `options` but README and some comment references were missed. Fixing it in current PR


samarsault (approved):

itsjwala (approved):
lgtm, might require updates on docs.percy.io

PR #1323, opened 2023-07-23
Adding support for legacy protocol

Amit3200:
We have 2 protocols, which boils down to dialects `W3C` and `OSS` now `OSS` & `W3C` have different response. To make consistent and increase coverage for now we will normalize the data by adding a layer.

### DIFF
--- a/packages/webdriver-utils/src/metadata/normalizeData.js
+++ b/packages/webdriver-utils/src/metadata/normalizeData.js
@@ -1,0 +1,34 @@
+export default class NormalizeData {
+  osRollUp(os) {
+    if (os.toLowerCase().startsWith('win')) {
+      return 'Windows';
+    } else if (os.toLowerCase().startsWith('mac')) {
+      return 'OS X';
+    } else if (os.toLowerCase().includes('iphone') || os.toLowerCase().startsWith('ios')) {
+      return 'iOS';
+    } else if (os.toLowerCase().startsWith('android')) {
+      return 'Android';
+    }
+    return os;
+  }
+
+  browserRollUp(browserName, device) {
+    if (device) {
+      if (browserName?.toLowerCase().includes('chrome')) {
+        return 'chrome';
+      } else if ((browserName?.toLowerCase().includes('iphone') ||
+          browserName?.toLowerCase().includes('ipad'))) {
+        return 'safari';
+      }
+    }
+    return browserName;
+  }
+
+  browserVersionRollUp(browserVersion, deviceName, device) {
+    if (device) {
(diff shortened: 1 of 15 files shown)


packages/webdriver-utils/src/metadata/normalizeData.js
```diff
@@ -0,0 +1,34 @@
...
+    return browserName;
+  }
+
+  browserVersionRollUp(browserVersion, deviceName, device) {
+    if (device) {
+      // return `${this.osRollUp(os)} ${osVersion?.split('.')[0]}`;
+      return deviceName;
+    }
+    return browserVersion?.split('.')[0];
+  }
```
ninadbstack, line 33: Didnt understand why are we using deviceName for browserVersionRollUp ?

packages/webdriver-utils/test/metadata/normalizeData.test.js
```diff
@@ -0,0 +1,84 @@
...
+
+    describe('should return browsers as it is for desktop', () => {
+      it('should return the browser as detected', () => {
+        expect(normalizeDataObj.browserRollUp('Chrome', false)).toBe('Chrome');
+      });
+    });
+  });
+
+  describe('browserVersionRollUp', () => {
+    describe('should rollup browser version for devices', () => {
```
ninadbstack, line 69: Maybe we should rename the function ? or use two separate here ?

packages/webdriver-utils/src/metadata/normalizeData.js
Amit3200: Reason being that in case of devices we don't really get the `browserVersion` properly. In capabilities we are getting the `browserVersion` as the `iphone` or `device_name` itself in certain devices. Responses differ when using different appium and selenium Version. And from the UI perspective to keep it consistent we moved it to keep `device_name` this ensures no UI change is required and nothing is breaking.

packages/webdriver-utils/test/metadata/normalizeData.test.js
Amit3200: Didn't get it properly? Open for new function rename as well. As of now I kept it as for devices `browserVersion` will be deviceName (initially we were thinking we will show OS version, however things didn't seemed consistent on UI), and for devices we are showing the actual `browserVersion`.

packages/webdriver-utils/src/metadata/normalizeData.js
ninadbstack: Got it, then can we just rename function to `browserVersionOrDeviceNameRollup` and add the comment on top ? [ the same comment as what you have explained above

ninadbstack (approved):

PR #1329, opened 2023-07-27
Added graceful handling of aborted requests

ninadbstack:
What ?
- Added graceful handling of aborted requests
- Now we keep track of aborted requests and whenever we are going to make a `session.send` call we check if we have already received abortion event for same request id 
- If request was already aborted we dont print it out as `Error: Protocol error (Fetch.fulfillRequest): Invalid InterceptionId.`
- We now print two new logs under debug mode, one for aborted request event and once when we ignore error above due to the same.

Why?
- There are cases when a browser triggers a network request and cancels it before its resolved. It usually happens when a resource was originally required, but is now not required [ say you have an `srcset` with multiple entries and you change the width of the browser fast enough such that it starts loading one of the resources, but now it doesn't need it ]
- In this case we would first receive `Network.requestWillBeSent` event, which we would eventually respond with `Fetch.continueRequest` or `Fetch.fulfillRequest` etc, but before we respond if the request gets cancelled [which invalidates `Interception Id`] we get above Protocol error. 

Test:
- This was hard to write test for as its hard to reproduce this race [ or browser cancelling a request ] to begin with
- So we are using `fetchWithTimeout` helper so that we can mimic a cancellation from browser
- In order to reproduce the race condition we specifically delay `Fetch.continueRequest` response by 1 sec so that request triggered by `fetch` is cancelled
- Which is equivalent to calling a `Fetch.*` method with a request/interception Id of a request thats already cancelled

### DIFF
--- a/packages/core/src/network.js
+++ b/packages/core/src/network.js
@@ -6,6 +6,7 @@ import { normalizeURL, hostnameMatches, createResource, waitFor } from './utils.
 const MAX_RESOURCE_SIZE = 25 * (1024 ** 2); // 25MB
 const ALLOWED_STATUSES = [200, 201, 301, 302, 304, 307, 308];
 const ALLOWED_RESOURCES = ['Document', 'Stylesheet', 'Image', 'Media', 'Font', 'Other'];
+const ABORTED_MESSAGE = 'Request was aborted by browser';
 
 // The Interceptor class creates common handlers for dealing with intercepting asset requests
 // for a given page using various devtools protocol events and commands.
@@ -18,6 +19,7 @@ export class Network {
   #requests = new Map();
   #intercepts = new Map();
   #authentications = new Set();
+  #aborted = new Set();
 
   constructor(page, options) {
     this.page = page;
@@ -87,6 +89,25 @@ export class Network {
     });
   }
 
+  // Validates that requestId is still valid as sometimes request gets cancelled and we have already executed
+  // _forgetRequest for the same, but we still attempt to make a call for it and it fails
+  // with Protocol error (Fetch.failRequest): Invalid InterceptionId.
+  async send(session, method, params) {
+    /* istanbul ignore else: currently all send have requestId */
+    if (params.requestId) {
+      /* istanbul ignore if: race condition, very hard to mock this */
+      if (this.isAborted(params.requestId)) {
+        throw new Error(ABORTED_MESSAGE);
+      }
+    }
+
+    return await session.send(method, params);
+  }
+
+  isAborted(requestId) {
+    return this.#aborted.has(requestId);
+  }
+
   // Throw a better network timeout error
   _throwTimeoutError(msg, filter = () => true) {
     if (this.log.shouldLog('debug')) {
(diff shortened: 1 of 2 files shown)


packages/core/src/network.js
```diff
@@ -87,6 +89,25 @@ export class Network {
     });
   }
 
+  // Validates that requestId is still valid as sometimes request gets cancelled and we have already executed
+  // _forgetRequest for the same, but we still attempt to make a call for it and it fails
+  // with Protocol error (Fetch.failRequest): Invalid InterceptionId.
+  async send(session, method, params) {
+    /* istanbul ignore else: currently all send have requestId */
+    if (params.requestId) {
```
samarsault, line 97: If all send have requestId then why are we checking this?

packages/core/src/network.js
```diff
@@ -264,6 +292,7 @@ async function sendResponseResource(network, request, session) {
   let log = network.log;
   let url = originURL(request);
   let meta = { ...network.meta, url };
+  let send = async (method, params) => await network.send(session, method, params);
```
samarsault, line 295: Optional
```suggestion
  let send = (method, params) => network.send(session, method, params);
```

packages/core/src/network.js
```diff
@@ -87,6 +89,25 @@ export class Network {
...
+  // _forgetRequest for the same, but we still attempt to make a call for it and it fails
+  // with Protocol error (Fetch.failRequest): Invalid InterceptionId.
+  async send(session, method, params) {
+    /* istanbul ignore else: currently all send have requestId */
+    if (params.requestId) {
+      /* istanbul ignore if: race condition, very hard to mock this */
+      if (this.isAborted(params.requestId)) {
+        throw new Error(ABORTED_MESSAGE);
+      }
+    }
```
itsjwala, line 102: what if the requestId is not present?

packages/core/test/discovery.test.js
```diff
@@ -748,6 +751,81 @@ describe('Discovery', () => {
...
+    await percy.snapshot({
+      name: 'cancelled request',
+      url: 'http://localhost:8000',
+      enableJavaScript: true
+    });
+
+    await percy.idle();
+
+    expect(logger.stdout).toEqual(jasmine.arrayContaining([
+      '[percy:core] Snapshot taken: cancelled request'
```
itsjwala, line 812: we're still taking snapshot in this case? or is it that this log line comes first?

packages/core/src/network.js
ninadbstack: There are other events that dont, and send is supposed to be used for all current and future events.

packages/core/src/network.js
ninadbstack: Just reduction in number of params, code looks worse to read [ and more off from how it looks for `session.send` calls ]

packages/core/src/network.js
ninadbstack: there is a check on line 97, not sure if you are referring to anything else

packages/core/test/discovery.test.js
ninadbstack: We will be taking snapshot anyway, this is not an error, it was just falsely shown as one. 
For customers this just avoid error logs in debug logs.

packages/core/src/network.js
itsjwala: that check is for if the requestId is present, if it's not present are we expecting to hit the condition for "invalid inteception id"?

itsjwala (approved):
lgtm, nice!

samarsault (approved):

PR #1338, opened 2023-08-04
Sending windowHeight to API for footer calculation

rishigupta1599:


packages/webdriver-utils/src/providers/automateProvider.js
```diff
@@ -144,7 +146,7 @@ export default class AutomateProvider extends GenericProvider {
         sha: tileData.split('-')[0] // drop build id
       }));
     }
-    return { tiles: tiles, domInfoSha: tileResponse.dom_sha };
+    return { tiles: tiles, domInfoSha: tileResponse.dom_sha, windowHeight: windowHeight * dpr };
```
nilshah98, line 149: `windowHeight * dpr` can be floating point correct ?
I understand that we are doing `to_i` in API, but code in CLI can be misrepresenting that API accepts floating point number.
Hence, I would suggest to make this int instead of float.

Amit3200: Then that should be done at all the places. floating digit comes for mobile as of now.

rishigupta1599: Ack I would add ceil function instead as we have seen height gets ceiled in images.

PR #1339, opened 2023-08-04
✨  Checks mime for incorrect content type fonts

ninadbstack:
Context:
- We know that browser does not return correct body for font responses to us and so we need to fetch it directly
- Now if server is not returning correct mimetype for font files [ if its sending `application/octate-stream` for example ] we didnt fetch font again as mime didnt match
- This caused errors while rendering font later in percy

Change:
- We now infer mime for all requests from url [ irrespective of original content type ] and if its font then we fetch it directly
- We dont tamper with original content type to honor browser behaviour when this gets rendered

### DIFF
--- a/packages/core/src/network.js
+++ b/packages/core/src/network.js
@@ -397,13 +397,17 @@ async function saveResponseResource(network, request) {
         return log.debug(`- Skipping disallowed resource type [${request.type}]`, meta);
       }
 
+      let detectedMime = mime.lookup(response.url);
       let mimeType = (
         // ensure the mimetype is correct for text/plain responses
-        response.mimeType === 'text/plain' && mime.lookup(response.url)
+        response.mimeType === 'text/plain' && detectedMime
       ) || response.mimeType;
 
-      // font responses from the browser may not be properly encoded, so request them directly
-      if (mimeType?.includes('font')) {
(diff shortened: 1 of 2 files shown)


packages/core/src/network.js
```diff
@@ -397,13 +397,17 @@ async function saveResponseResource(network, request) {
...
+        response.mimeType === 'text/plain' && detectedMime
       ) || response.mimeType;
 
-      // font responses from the browser may not be properly encoded, so request them directly
-      if (mimeType?.includes('font')) {
+      // if we detect a font mime, we dont want to override it as different browsers may behave
+      // differently for incorrect mimetype in font response, but we want to treat it as a
+      // font anyway as font responses from the browser may not be properly encoded,
+      // so request them directly.
+      if (mimeType?.includes('font') || (detectedMime && detectedMime.includes('font'))) {
```
samarsault, line 410:
```suggestion
      if (mimeType?.includes('font') || detectedMime?.includes('font')) {
```

ninadbstack: We cant do that, because when mime type is not found `detectedMime` is `false` and `false?. includes(..` throws exception as `?` only protects against null/undefined, not `false` [ learnt hard way :P ]

samarsault (approved):

itsjwala (approved):
Nice 🔥

PR #1343, opened 2023-08-11
✨ changes for creating executable binaries of `percy` CLI

pankaj443:
1. We are creating executable for all platforms,  win, mac and linux.
2. Using executable we are removing node dependency.
3. scripts/executable.sh
- Downloads required dependencies
- pkg doesn’t support ES6 so we are converting code using babel and some sed commands
- Use pkg to build executables for all platofrms
- codesign and notarize macos executable
5. Added Steps to upload executables as artifacts.


.github/workflows/build.yml
```diff
@@ -0,0 +1,27 @@
...
+          name: macos executable
+          path: run-macos
+      - uses: actions/upload-artifact@v3
+        with:
+          name: linux executable
+          path: run-linux
+      - uses: actions/upload-artifact@v3
+        with:
+          name: win executable
+          path: run-win.exe
```
itsjwala, line 27: we need to package both `arm64, amd64` variant for unix 

Eg: https://github.com/github/gh-ost/releases

Also the name of the package should be `percy`

itsjwala: while I was testing the executables on my machine. 

`./run-macos` works without any issues, though

if I run using absolute path, it breaks.

scripts/build.sh
```diff
@@ -0,0 +1,66 @@
+mkdir temp && cd temp
+
+npm init -y
+
+npm install --save-dev @percy/cli
```
itsjwala, line 5: we'll be building of existing commit, so no install here.

scripts/build.sh
```diff
@@ -0,0 +1,66 @@
...
+cp -R ./temp//node_modules/@percy/* packages/
+
+sed -i '' '/"type": "module",/d' ./package.json
+
+cd packages && sed -i '' '/"type": "module",/d' ./*/package.json && cd ..
+
+echo "import { cli } from '@percy/cli';\
+$(cat ./packages/cli/dist/percy.js)" > ./packages/cli/dist/percy.js
+
+sed -i '' '/Imports and returns compatibile CLI commands from various sources/a \
```
itsjwala, line 22: lets add comments for each of the hacks we'd to do here.

scripts/build.sh
```diff
@@ -0,0 +1,66 @@
...
+  if (outdir.charAt(0) == "/") {\
+    outdir = outdir.replace("/", "");\
+  }
+' ./packages/core/dist/install.js
+
+npm run build_cjs
+
+cp -R ./build/* packages/
+
+pkg ./packages/cli/bin/run.js -d
```
itsjwala, line 62: [optional]
since shell scripts can get difficult to maintain as we go, I'd recommend using node script itself 
Eg: https://github.com/google/zx

.github/workflows/build.yml
ninadbstack: One more thing, lets rename executables correctly. it should be 

`percy` for unix/linux and `percy.exe` on windows

scripts/build.sh
pankaj443: we need dist folder to convert code to cjs and use it but we don't have `bundle.js` in dom package in src folder. so I installed cli. is there another way to do so?

.github/workflows/build.yml
pankaj443: as decided we will update docs. if another cli is installed at that place it might be reason for it to break

scripts/build.sh
pankaj443: this would be significant effort, lets skip for now

prklm10 (approved):

scripts/build.sh
itsjwala:
```
yarn build
```

this generates the dist folder.

pankaj443: thanks, fixed

scripts/build.sh
```diff
@@ -0,0 +1,83 @@
...
+  }\
+  outdir = outdir.replace("C:\\\\","");\
+  archive = archive.replace("C:\\\\","");
+' ./packages/core/dist/install.js
+
+sed -i '' '/let outdir/a \
+  if (outdir.charAt(0) == "/") {\
+    outdir = outdir.replace("/", "");\
+  }
+' ./packages/core/dist/install.js
```
itsjwala, line 48: can we please maintain this as part of the code itself behind an `ENV`?  this will grow unmaintainable in future.

scripts/build.sh
```diff
@@ -0,0 +1,83 @@
...
+mv run-macos percy-macos
+mv run-linux percy-linux
+mv run-win.exe percy-win.exe
+
+# Cleanup temp folder
+rm -rf temp
+rm -rf build
+
+# Sign & Notrize mac app
+echo "$APPLE_DEV_CERT" | base64 -d > AppleDevIDApp.p12
```
itsjwala, line 67: confirm this doesn't logs the output?

scripts/build.sh
```diff
@@ -0,0 +1,83 @@
...
+security set-keychain-settings -t 3600 -l ~/Library/Keychains/percy.keychain
+security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k percy ~/Library/Keychains/percy.keychain-db
+
+codesign  --force --verbose=4 -s "Developer ID Application: BrowserStack Inc (763K6K6H44)" --options runtime --entitlements scripts/files/entitlement.plist --keychain ~/Library/Keychains/percy.keychain percy-macos
+
+zip percy-macos.zip percy-macos
+cat scripts/files/notarize_config.json.tmpl | sed -e "s/{{APPLE_ID_USERNAME}}/$APPLE_ID_USERNAME/" | sed -e "s/{{APPLE_ID_KEY}}/$APPLE_ID_KEY/" > notarize_config.json
+gon -log-level=error -log-json notarize_config.json
+
+security delete-keychain percy.keychain
```
itsjwala, line 83: lets delete the p12 certs as well here, can you make a `cleanup()` function and move all the cleanup in there?

scripts/files/entitlement.plist
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+<key>com.apple.security.cs.allow-jit</key>
+<true/>
+<key>com.apple.security.cs.allow-unsigned-executable-memory</key>
+<true/>
+</dict>
+</plist>
```
itsjwala, line 10: EOF

scripts/files/notarize_config.json.tmpl
```diff
@@ -0,0 +1,10 @@
+{
+  "apple_id": {
+    "username" : "{{APPLE_ID_USERNAME}}",
+    "password":  "{{APPLE_ID_KEY}}"
+  },
+  "notarize" : [{
+    "path": "percy-macos.zip",
+    "bundle_id": "percy.io.cli"
```
itsjwala, line 8:
```suggestion
    "bundle_id": "io.percy.cli"
```

not sure if this java style package name?, coz - `cli.percy.io` entails to `io.percy.cli` package.

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
+name: Build Executables
+on:
+  release:
+    types: [published]
```
itsjwala, line 4: [temporary] lets try doing a publish from branch, to test e2e?

scripts/build.sh
pankaj443: yes confirmed.

.github/workflows/build.yml
pankaj443: i have tested on branch already. https://github.com/percy/cli/actions/runs/5998822360

package.json
```diff
@@ -9,6 +9,7 @@
     "url": "https://github.com/percy/cli"
   },
   "scripts": {
+    "build_cjs": "BABEL_ENV=dev babel packages -d build",
```
itsjwala, line 12: can this be linked with lerna command?

lerna is mono repo package manager, check all the commands below for reference.

itsjwala: [optional] hook with existing `build` command 

so following without `ENV` builds normally
```
yarn build
```

and following builds with the hacks present in `build.sh` ( we'll have `if else` in codebase which is fine and testable)

```
BABEL_ENV=executable yarn build
```

pankaj443: i don’t think this is possible `babel packages -d build ` command is converting everything present inside packages folder of cli to cjs format supported by `pkg`.

scripts/build.sh
pankaj443: added cleanup fn

scripts/build.sh
```diff
@@ -0,0 +1,52 @@
...
+
+yarn install
+
+# Remove type from package.json files
+sed -i '' '/"type": "module",/d' ./package.json
+cd packages && sed -i '' '/"type": "module",/d' ./*/package.json && cd ..
+
+echo "import { cli } from '@percy/cli';\
+$(cat ./packages/cli/dist/percy.js)" > ./packages/cli/dist/percy.js
+
```
itsjwala, line 22: can we move L16-L21 also to code?

packages/cli/src/commands.js
```diff
@@ -140,7 +148,13 @@ export async function importCommands() {
       pkgs.set(pkg.name, () => Promise.all(
         pkg['@percy/cli'].commands.map(async cmdPath => {
           let modulePath = path.join(pkgPath, cmdPath);
-          let module = await import(url.pathToFileURL(modulePath).href);
+          // Below code is used in scripts/build.sh to update href
+          let module = null;
+          if (process.env.ENVIRONMENT === 'executable') {
```
itsjwala, line 153: hmm, this would mean we'll have to package .env as well in executable.

what we want is while building the code, default skip/remove the codepath for else {}

lets think more about the solutioning here.

pankaj443: yes we will have to create a .env file and use dotenv package. that should work

scripts/build.sh
pankaj443: these lines can’t be moved behind env

itsjwala: [optional] ideally we can using build script ( `yarn build` ) we can use the same `env` to differentiate.

Give it a try, otherwise we can do in followup PR

packages/cli/src/commands.js
itsjwala: yeah, we already have the package under `percy/env` we can extend it there (`packages/env/src/environment.js`)

scripts/build.sh
```diff
@@ -0,0 +1,54 @@
...
+# Rename executables
+mv run-macos percy-macos
+mv run-linux percy-linux
+mv run-win.exe percy-win.exe
+
+# Sign & Notrize mac app
+echo "$APPLE_DEV_CERT" | base64 -d > AppleDevIDApp.p12
+
+security create-keychain -p percy percy.keychain
```
itsjwala, line 40: Add in confluence, link existing if already present.

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+          brew install mitchellh/gon/gon
+      - run: ./scripts/build.sh
+        env:
+          APPLE_DEV_CERT: ${{secrets.APPLE_DEV_CERT}}
+          APPLE_ID_USERNAME: ${{secrets.APPLE_ID_USERNAME}}
+          APPLE_ID_KEY: ${{secrets.APPLE_ID_KEY}}
+      - uses: actions/upload-artifact@v3
+        with:
+          name: macos executable
+          path: percy-macos
```
itsjwala, line 26: zip file can be

`percy-darwin.zip`
`percy-linux.zip`
`percy-win.zip`

all should unzip to `percy` executable (percy.exe for win)

pankaj443: updated

packages/cli/src/commands.js
pankaj443: .env will work only when file is present from location we are executing, i have created a new file with will be entrypoint for executable and will set nodeenv

scripts/build.sh
pankaj443: it might take time let do in followup PR

packages/cli/bin/executable.js
```diff
@@ -0,0 +1,14 @@
+// This file is used as a entry point for executable with NODE_ENV=executable
+process.env.NODE_ENV = 'executable';
```
itsjwala, line 2: remove this file.

we can use in `executable.sh` itself.
```
export NODE_ENV = 'executable';
```

packages/cli/src/commands.js
```diff
@@ -92,6 +92,14 @@ function importLegacyCommands(commandsPath) {
   });
 }
 
+function formatFilepath(filepath) {
+  /* istanbul ignore next */
+  if (!(process.platform === 'win32' || process.platform === 'win64')) {
```
itsjwala, line 97:
```suggestion
  if (!process.platform.startsWith('win')) {
```

packages/core/src/install.js
```diff
@@ -63,7 +64,22 @@ export async function download({
...
+    if (outdir.charAt(0) === '/') {
+      outdir = outdir.replace('/', '');
+    }
+  }
+
+  let command = 'pwd';
   let archive = path.join(outdir, decodeURIComponent(url.split('/').pop()));
+  if (process.env.NODE_ENV === 'executable') {
+    /* istanbul ignore next */
+    if (process.platform === 'win32' || process.platform === 'win64') {
```
itsjwala, line 77: same comment as previous.

.github/workflows/build.yml
itsjwala: `percy` isn't executable currently it seems, does the binary need `chmod +x` while building?

packages/cli/bin/executable.js
```diff
@@ -0,0 +1,14 @@
...
+// ensure that we're running within a supported node version
```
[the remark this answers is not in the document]
itsjwala: Also include the asset discovery browser in executable - [ref](https://docs.percy.io/docs/caching-asset-discovery-browser-in-ci) make it default.

[optional] create separate variant of executable which doesn't includes browser, so user can specify browser themselves if they need instead of downloading bloated binary.

.github/workflows/build.yml
pankaj443: updated. we need to convert to tar format to retain permissions.

scripts/build.sh
pankaj443: Added in https://browserstack.atlassian.net/wiki/spaces/ENG/pages/3818786920/Percy+SDK+Engineering+Handbook

packages/cli/bin/executable.js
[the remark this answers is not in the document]
pankaj443: the current executable doesn’t include pre-downloade browser, I tried creating one with installed browser by adding export PERCY_POSTINSTALL_BROWSER=true before yarn install step but it didn’t work for me. also if we want include browser we will have to create diff github actions for each platform right so that it downloads correct chromium browser?

packages/cli/bin/run.cjs
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+// Update NODE_ENV for executable
```
itsjwala, line 3:
```suggestion
// DO NOT REMOVE: Update NODE_ENV for executable
```

scripts/executable.sh
```diff
@@ -0,0 +1,55 @@
...
+  security delete-keychain percy.keychain
+}
+
+npm install -g pkg
+
+yarn install
+yarn build
+
+# Remove type from package.json files
+sed -i '' '/"type": "module",/d' ./package.json
```
itsjwala, line 16: don't fail silently, if match not found.

scripts/executable.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
itsjwala, line 1: add default bash flags, `set -eu pipefail`?

.github/workflows/executable.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Install gon via HomeBrew for code signing and app notarization
+        run: |
+          brew tap mitchellh/gon
+          brew install mitchellh/gon/gon
+      - run: ./scripts/executable.sh
+        env:
+          APPLE_DEV_CERT: ${{secrets.APPLE_DEV_CERT}}
+          APPLE_ID_USERNAME: ${{secrets.APPLE_ID_USERNAME}}
+          APPLE_ID_KEY: ${{secrets.APPLE_ID_KEY}}
+      - name: Tar file osx
```
itsjwala, line 23: move all tar under single `run:`

pankaj443: updated

scripts/executable.sh
pankaj443: sed does not exit 1 even if match is not found.

itsjwala: you'll need to use GNU sed - [ref](https://stackoverflow.com/a/66365284/9485283) 

```bash
➜  echo "Bash Scripting Language" | gsed 's/Bash/Perl/;tx;q1;:x'
Perl Scripting Language
➜  echo $?
0
➜  echo "Bas Scripting Language" | gsed 's/Bash/Perl/;tx;q1;:x'
Bas Scripting Language
➜  echo $?
1
```

packages/cli/bin/executable.js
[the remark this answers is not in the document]
itsjwala: hmm, lets tackle this in followup.

package.json
itsjwala: lets tackle in followup PR.

.github/workflows/executable.yml
```diff
@@ -21,6 +21,7 @@ jobs:
       - name: Create tar files
         run: |
           tar -cvf percy-osx.tar percy
+          ./percy --version
```
itsjwala, line 26: swap linux with macos step, and run the executable as separate `run:` 

we usually check end of the step and skip checking intermediate steps.

scripts/executable.sh
pankaj443: updated

itsjwala (approved):
LGTM!

* update description ( remove internal link as it won't be accessible by outside folks)
* note down all the followup items in the task, we can/will pickup later.

PR #1350, opened 2023-08-25
Adding CLI Metrics to measure client duration

Amit3200:
Capturing Screenshot Timings and Metrics.

### DIFF
--- a/packages/webdriver-utils/src/index.js
+++ b/packages/webdriver-utils/src/index.js
@@ -34,12 +34,16 @@ export default class WebdriverUtils {
 
   async automateScreenshot() {
     try {
+      const startTime = Date.now();
       this.log.info(`[${this.snapshotName}] : Starting automate screenshot ...`);
       const automate = ProviderResolver.resolve(this.sessionId, this.commandExecutorUrl, this.capabilities, this.sessionCapabilites, this.clientInfo, this.environmentInfo, this.options, this.buildInfo);
       this.log.debug(`[${this.snapshotName}] : Resolved provider ...`);
       await automate.createDriver();
       this.log.debug(`[${this.snapshotName}] : Created driver ...`);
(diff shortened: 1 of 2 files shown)


packages/webdriver-utils/src/index.js
```diff
@@ -34,12 +34,16 @@ export default class WebdriverUtils {
...
+      const startTime = Date.now();
       this.log.info(`[${this.snapshotName}] : Starting automate screenshot ...`);
       const automate = ProviderResolver.resolve(this.sessionId, this.commandExecutorUrl, this.capabilities, this.sessionCapabilites, this.clientInfo, this.environmentInfo, this.options, this.buildInfo);
       this.log.debug(`[${this.snapshotName}] : Resolved provider ...`);
       await automate.createDriver();
       this.log.debug(`[${this.snapshotName}] : Created driver ...`);
-      return await automate.screenshot(this.snapshotName, this.options);
+      const comparisonData = await automate.screenshot(this.snapshotName, this.options);
+      comparisonData.metadata.cliScreenshotStartTime = startTime;
+      comparisonData.metadata.cliScreenshotEndTime = Date.now();
```
ninadbstack, line 45: Not sure if you need actual start/end time or only duration, as only duration is used in the event. You can just send `screenshotDuration` instead of start and end time

Amit3200: The reason we are sending both the times not the duration is if we want to get full comparison time duration we can subtract the end time from differ and start time from here. This helps in getting stepWise time taken for various things.

ninadbstack: Well, if we need instrumentation on that level, we should be creating a separate pattern for it. This doesn't scale well and requires a lot of boilerplate repetition.  

Anyway, its fine for now, lets discuss its refactor separately.

ninadbstack (approved):

PR #1362, opened 2023-09-01
🏗️  Upload executables as assets

pankaj443:
- Upload executables as assets when releasing
- While uploading assets zip also retains executable permissions no need to convert to tar

### DIFF
--- a/.github/workflows/executable.yml
+++ b/.github/workflows/executable.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Verify executable
         run: ./percy --version
       - name: Upload assets
-        uses: softprops/action-gh-release@v1
+        uses: softprops/action-gh-release@c9b46fe7aad9f02afd89b12450b780f52dacfb2d
         with:
           files: |
             percy-osx.zip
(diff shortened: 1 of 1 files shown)


.github/workflows/executable.yml
```diff
@@ -20,23 +20,20 @@ jobs:
           APPLE_DEV_CERT: ${{secrets.APPLE_DEV_CERT}}
           APPLE_ID_USERNAME: ${{secrets.APPLE_ID_USERNAME}}
           APPLE_ID_KEY: ${{secrets.APPLE_ID_KEY}}
-      - name: Create tar files
+      - name: Create zip files
         run: |
-          tar -cvf percy-linux.tar percy
+          zip percy-linux.zip percy
```
itsjwala, line 25: why not `tar.gz` file since zip looses the executable permission?

```bash
tar -czvf percy-linux.tar.gz percy`
```

pankaj443: it does not in this case. it was loosing when we were uploading as artifact.

itsjwala (approved):

PR #1371, opened 2023-09-13
Percy on Automate: Making async create comparisons

Amit3200:
- Enhancement
✨ Making upload comparisons async in order to attain better speed.

### DIFF
--- a/packages/core/src/api.js
+++ b/packages/core/src/api.js
@@ -118,8 +118,9 @@ export function createPercyServer(percy, port) {
     }))
     .route('post', '/percy/automateScreenshot', async (req, res) => {
       req = percyAutomateRequestHandler(req, percy.build);
+      percy.upload(await new WebdriverUtils(req.body).automateScreenshot());
       res.json(200, {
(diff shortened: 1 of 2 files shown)


packages/core/src/api.js
```diff
@@ -118,8 +118,9 @@ export function createPercyServer(percy, port) {
     }))
     .route('post', '/percy/automateScreenshot', async (req, res) => {
       req = percyAutomateRequestHandler(req, percy.build);
+      percy.upload(await new WebdriverUtils(req.body).automateScreenshot());
```
itsjwala, line 121: check #1369 

I've tried this, though this doesn't make it async, since we need to have a way to make that `await` removed which will require some more code changes. though this version is much simplified then what it is in master.

Amit3200: Yes this is not full await however this unblocks PoA SDK. CLI can keep on waiting but if we add await which is master as of now that blocks the PoA. So if we have this thing then we do see speed improvements.

itsjwala: you can close this PR, as I've updated this in the linked PR. 

Also can you please review other bit in that PR?

PR #1377, opened 2023-09-22
Removing PercyCSS from CLI for performance benefit

Amit3200:
Removing percyCSS execution from CLI for increasing speed. The percyCSS will be executed on the required Infra to avoid latency issues.

### DIFF
--- a/packages/webdriver-utils/src/providers/genericProvider.js
+++ b/packages/webdriver-utils/src/providers/genericProvider.js
@@ -42,32 +42,6 @@ export default class GenericProvider {
     this.options.freezeAnimation = this.options.freezeAnimation || false;
   }
 
-  defaultPercyCSS() {
-    return `*, *::before, *::after {
-      -moz-transition: none !important;
-      transition: none !important;
-      -moz-animation: none !important;
-      animation: none !important;
-      animation-duration: 0 !important;
-      caret-color: transparent !important;
-      content-visibility: visible !important;
-    }
-    html{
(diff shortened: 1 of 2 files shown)


packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -120,9 +80,7 @@ export default class GenericProvider {
 
     this.addDefaultOptions();
 
-    const percyCSS = (this.defaultPercyCSS() + (this.options.percyCSS || '')).split('\n').join('');
-    log.debug(`[${name}] : Applying the percyCSS - ${this.options.percyCSS}`);
-    await this.addPercyCSS(percyCSS);
+    this.options.percyCSS = (this.options.percyCSS || '').split('\n').join('');
```
itsjwala, line 83: query: why are we replacing new line with nothing?

itsjwala (approved):

packages/webdriver-utils/src/providers/genericProvider.js
Amit3200: Yes its kind of serializing, I tested once earlier it was not working, so i kept this flow as it is. I guess if we are not adding anything it should run fine. When we are adding things then it needs such thing.

itsjwala: weird 😕  anyways..

shantanuk-browserstack (approved):

PR #1378, opened 2023-09-25
:sparkles: Introduce new config option enableLayout for enabling layout testing.

prklm10:
Adding following to global config
 - enableLayout (default false)

### DIFF
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -44,6 +44,10 @@ export const configSchema = {
         type: 'boolean',
         default: false
       },
+      enableLayout: {
+        type: 'boolean',
+        default: false
+      },
       domTransformation: {
         type: 'string'
       },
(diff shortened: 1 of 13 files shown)


packages/dom/README.md
```diff
@@ -38,6 +38,7 @@ const domSnapshot = await page.evaluate(() => PercyDOM.serialize(options))
 - `enableJavaScript` — When true, does not serialize some DOM elements
 - `domTransformation` — Function to transform the DOM after serialization
 - `disableShadowDOM` — disable shadow DOM capturing, this option can be passed to `percySnapshot` its part of per-snapshot config.
+- `enableLayout` - When true, It enables layout testing.
```
ninadbstack, line 41:
```suggestion
- `enableLayout` - When true, It enables layout testing on entire screenshot
```

packages/core/src/config.js
```diff
@@ -44,6 +44,10 @@ export const configSchema = {
         type: 'boolean',
         default: false
       },
+      enableLayout: {
+        type: 'boolean',
+        default: false
```
itsjwala, line 49:
```suggestion
```

lets not add default for now. we can interpret it as false when nothing is sent.

packages/dom/README.md
itsjwala: This documentation won't be part of `dom` package, as we don't use it here.

prklm10: done

packages/core/src/config.js
prklm10: removed default value.

itsjwala (approved):

PR #1388, opened 2023-10-06
✨ Added token support in percy config

ninadbstack:
- Added support for token in percy config, under `percy.token`
- `PERCY_TOKEN` env var would still override token from config


prklm10 (approved):
🍍  LGTM

PR #1400, opened 2023-10-17
[PER-2582] Do not execute git commands when `PERCY_SKIP_GIT_CHECK` is set to True

shantanuk-browserstack:
When git is not installed locally,  percy-cli triggers git popup.

image

We have introduced a env var `PERCY_SKIP_GIT_CHECK`. When this is set to True, git commands will not be executed locally when using percy-cli.

### DIFF
--- a/packages/env/test/defaults.test.js
+++ b/packages/env/test/defaults.test.js
@@ -173,7 +173,6 @@ describe('Defaults', () => {
   });
 
   describe('env PERCY_SKIP_GIT_CHECK', () => {
-  
     beforeEach(() => {
       process.env.PERCY_SKIP_GIT_CHECK = true;
     });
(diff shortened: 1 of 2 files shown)


packages/env/test/defaults.test.js
```diff
@@ -171,4 +171,20 @@ describe('Defaults', () => {
...
+  describe('env PERCY_SKIP_GIT_CHECK', () => {
+    beforeEach(() => {
+      process.env.PERCY_SKIP_GIT_CHECK = true;
+    });
+
+    afterAll(() => {
+      process.env.PERCY_SKIP_GIT_CHECK = false;
+    });
+
+    it('does not make git commands, if there are any', () => {
```
ninadbstack, line 184: Don't we always execute git ? Is there a condition for 'if there are any' ?

shantanuk-browserstack: updated the description

ninadbstack (approved):

PR #1401, opened 2023-10-18
✨ adding endpoint for sending failed event

prklm10:
Adding endpoint for sending failed event.

### DIFF
--- a/packages/cli-exec/src/exec.js
+++ b/packages/cli-exec/src/exec.js
@@ -118,7 +118,8 @@ async function* spawn(cmd, args, percy) {
     proc.on('close', code => {
       closed = code;
       if (code !== 0) {
-        // Only send event when there is a global error code
+        // Only send event when there is a global error code and
+        // percy token is present
         if (process.env.PERCY_TOKEN) {
           const myObject = {
             errorKind: 'cli',
--- a/packages/cli-exec/src/exec.js
+++ b/packages/cli-exec/src/exec.js
@@ -57,6 +57,7 @@ export const exec = command('exec', {
   // attempt to start percy if enabled
   if (!percy) {
     log.warn('Percy is disabled');
+    // we can send this data as well.
   } else {
     try {
       // Skip this for app because they are triggered as app:exec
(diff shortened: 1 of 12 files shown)


shantanuk-browserstack (approved):

packages/cli-exec/src/exec.js
```diff
@@ -94,15 +94,41 @@ export const exec = command('exec', {
...
+      });
+    }
+
     proc.on('error', err => (error = err));
 
+    proc.on('close', code => {
+      closed = code;
+      if (code !== 0) {
+        // Only send event when there is a global error code
+        if (process.env.PERCY_TOKEN) {
```
pankaj443, line 122: Can you add a comment around this condition?

prklm10: added a comment

packages/client/src/client.js
```diff
@@ -516,6 +516,21 @@ export class PercyClient {
     return comparison;
   }
 
+  async sendFailedEvents(buildId, { errorKind = 'sdk', client = null, clientVersion = null, cliVersion = null, errorMessage = null } = {}) {
+    validateId('build', buildId);
+    this.log.debug('Sending FailedEvents');
+    return this.post(`builds/${buildId}/failed-events`, {
```
ninadbstack, line 522: does this only send failed events ? so we can't calculate percentage ? we were going to make a reusable instrumentation endpoint right ?

packages/client/src/client.js
```diff
@@ -516,6 +516,21 @@ export class PercyClient {
...
+      data: {
+        buildId: buildId,
```
ninadbstack, line 524: why do we need this if its already sent in endpoint path ?

prklm10: it was added because in api pr it was missing, removed and added build_id in api pr.

packages/core/src/utils.js
```diff
@@ -60,6 +60,35 @@ export function percyAutomateRequestHandler(req, percy) {
   req.body.buildInfo = percy.build;
 }
 
+// Returns the body for sendEvent structure
```
ninadbstack, line 63: This is not returning anything

packages/core/src/utils.js
```diff
@@ -60,6 +60,35 @@ export function percyAutomateRequestHandler(req, percy) {
...
+export function percyBuildEventHandler(req, cliVersion) {
```
ninadbstack, line 64: You seem to be updating incoming request body object here, its usually not a good practice, its better to pluck the data that you want from incoming body and create a new object [ that automatically excludes unintended/malicious data

packages/core/src/api.js
```diff
@@ -121,6 +121,12 @@ export function createPercyServer(percy, port) {
       percy.upload(await WebdriverUtils.automateScreenshot(req.body));
       res.json(200, { success: true });
     })
+  // Recieves events from sdk's.
+    .route('post', '/percy/events', async (req, res) => {
+      const body = percyBuildEventHandler(req, pkg.version);
+      await percy.client.sendBuildEvents(percy.build?.id, body);
```
ninadbstack, line 127: [Optional] or for future, this can be made async. Currently this call is blocking customer script i believe, unless sdks dont wait on the response

ninadbstack (approved):

packages/core/src/api.js
prklm10: we can do this in future also. As we are only sending the event from catch block of sdk. (when any screenshot is failing) so it shouldn't have any performance implications.

PR #1408, opened 2023-10-22
Fixing FindRegion + Refactor

Amit3200:
- Fixing `getRegionObject` - Considering the scroll positions + Header NavBar for devices.
-  Fixing Xpath and Selector FindBy Logic.

### DIFF
--- a/packages/webdriver-utils/src/driver.js
+++ b/packages/webdriver-utils/src/driver.js
@@ -77,4 +77,25 @@ export default class Driver {
     const response = JSON.parse((await request(baseUrl, options)).body);
     return response.value;
   }
+
+  async findElementBoundingBox(using, value) {
+    if (using === 'xpath') {
+      return await this.findElementXpath(value);
+    } else if (using === 'css selector') {
+      return await this.findElementSelector(value);
+    }
+  }
+
+  async findElementXpath(xpath) {
+    const command = { script: `return document.evaluate('${xpath}', document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue.getBoundingClientRect();`, args: [] };
+    const response = await this.executeScript(command);
+    return response.value;
+  }
+
+  async findElementSelector(selector) {
+    selector = selector.replace('\\', '\\\\');
+    const command = { script: `return document.querySelector('${selector}').getBoundingClientRect();`, args: [] };
+    const response = await this.executeScript(command);
+    return response.value;
+  }
 }
--- a/packages/webdriver-utils/src/driver.js
+++ b/packages/webdriver-utils/src/driver.js
@@ -10,7 +10,7 @@ export default class Driver {
     this.passedCapabilities = passedCapabilities;
   }
 
-  requestPostOptions(command) {
+  static requestPostOptions(command) {
     return {
       method: 'POST',
       headers: {
@@ -53,7 +53,7 @@ export default class Driver {
     if (!command.script.includes('browserstack_executor')) {
       command.script = `/* percy_automate_script */ \n ${command.script}`;
     }
-    const options = this.requestPostOptions(command);
+    const options = Driver.requestPostOptions(command);
     const baseUrl = `${this.executorUrl}/session/${this.sessionId}/execute/sync`;
     const response = JSON.parse((await request(baseUrl, options)).body);
     return response;
@@ -72,7 +72,7 @@ export default class Driver {
   }
 
   async findElement(using, value) {
-    const options = this.requestPostOptions({ using, value });
+    const options = Driver.requestPostOptions({ using, value });
     const baseUrl = `${this.executorUrl}/session/${this.sessionId}/element`;
     const response = JSON.parse((await request(baseUrl, options)).body);
     return response.value;
--- a/packages/webdriver-utils/src/driver.js
+++ b/packages/webdriver-utils/src/driver.js
@@ -87,6 +87,7 @@ export default class Driver {
   }
 
   async findElementXpath(xpath) {
+    xpath = xpath.replace(/'/g, '"');
     const command = { script: `return document.evaluate('${xpath}', document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue.getBoundingClientRect();`, args: [] };
     const response = await this.executeScript(command);
     return response.value;
--- a/packages/webdriver-utils/src/driver.js
+++ b/packages/webdriver-utils/src/driver.js
@@ -10,6 +10,16 @@ export default class Driver {
     this.passedCapabilities = passedCapabilities;
   }
 
+  requestPostOptions(command) {
+    return {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json;charset=utf-8'
+      },
+      body: JSON.stringify(command)
+    };
+  }
+
   async getCapabilites() {
     return await Cache.withCache(Cache.caps, this.sessionId, async () => {
       try {
@@ -43,13 +53,7 @@ export default class Driver {
     if (!command.script.includes('browserstack_executor')) {
       command.script = `/* percy_automate_script */ \n ${command.script}`;
     }
-    const options = {
-      method: 'POST',
-      headers: {
-        'Content-Type': 'application/json;charset=utf-8'
-      },
-      body: JSON.stringify(command)
-    };
+    const options = this.requestPostOptions(command);
     const baseUrl = `${this.executorUrl}/session/${this.sessionId}/execute/sync`;
     const response = JSON.parse((await request(baseUrl, options)).body);
     return response;
@@ -68,13 +72,7 @@ export default class Driver {
   }
 
   async findElement(using, value) {
-    const options = {
-      method: 'POST',
-      headers: {
-        'Content-Type': 'application/json;charset=utf-8'
-      },
-      body: JSON.stringify({ using, value })
-    };
+    const options = this.requestPostOptions({ using, value });
     const baseUrl = `${this.executorUrl}/session/${this.sessionId}/element`;
     const response = JSON.parse((await request(baseUrl, options)).body);
     return response.value;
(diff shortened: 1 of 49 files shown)


packages/webdriver-utils/src/metadata/desktopMetaData.js
```diff
@@ -49,7 +49,7 @@ export default class DesktopMetaData {
 
   async screenResolution() {
     return await Cache.withCache(Cache.resolution, this.driver.sessionId, async () => {
-      const data = await this.driver.executeScript({ script: 'return [(window.screen.width * window.devicePixelRatio).toString(), (window.screen.height * window.devicePixelRatio).toString()];', args: [] });
+      const data = await this.driver.executeScript({ script: 'return [parseInt(window.screen.width * window.devicePixelRatio).toString(), parseInt(window.screen.height * window.devicePixelRatio).toString()];', args: [] });
```
nilshah98, line 52: Why did we add `parseInt` here ?

Amit3200: resolutions should come as proper integer not as decimals. since this was desktop it was always coming as integer as dpr values are 1,2 however added it for consistent behaviour with respect to mobile.

nilshah98: Got it.
In that case, does `parseInt` take ceil or floor ?

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -191,16 +203,27 @@ export default class GenericProvider {
...
+
   async getRegionObject(selector, elementId) {
-    const scaleFactor = parseInt(await this.metaData.devicePixelRatio());
+    const scaleFactor = await this.metaData.devicePixelRatio();
     const rect = await this.driver.rect(elementId);
     const location = { x: rect.x, y: rect.y };
     const size = { height: rect.height, width: rect.width };
+    // Update YFactor Element is not visible in viewport
+    // In case of iOS if the element is not visible in viewport it gives 0,0 as coordinate.
+    this.updateYFactor(location);
```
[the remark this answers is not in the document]
Amit3200: @nilshah98  It was important to get the functionality getting tested. Separating the function makes the function readability look better imo  and testing becomes much easier for the various cases. So tomorrow if we find a new behaviour it can be handled in this function.

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -191,16 +203,27 @@ export default class GenericProvider {
     ];
   }
 
+  updateYFactor(location) {
+    if (this.currentOs === 'iOS') {
+      if (location.y === 0) {
```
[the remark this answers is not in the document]
Amit3200: `location.y === 0` element is not in the viewport or element may be behind the status bar at top. If its at top there might be a height which **might** be visible in viewport. Hence adjusting with `scaleYFactor` refer to the builds shared.

packages/webdriver-utils/src/metadata/desktopMetaData.js
Amit3200: `parseInt` takes floor values

packages/webdriver-utils/src/providers/genericProvider.js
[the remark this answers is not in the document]
Amit3200: Refactored the logic to save the network call and clubbed all the page shifting mechanism limited to one function.

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -36,6 +36,11 @@ export default class GenericProvider {
     this.debugUrl = null;
     this.header = 0;
     this.footer = 0;
+    this.pageYShiftFactor = 0;
+    this.pageXShiftFactor = 0;
+    this.scrollXFactor = 0;
+    this.scrollYFactor = 0;
+    this.currentOs = null;
```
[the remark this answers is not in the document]
Amit3200: using `currentOperatingSystem`.

packages/webdriver-utils/src/driver.js
```diff
@@ -43,13 +53,7 @@ export default class Driver {
...
       command.script = `/* percy_automate_script */ \n ${command.script}`;
     }
-    const options = {
-      method: 'POST',
-      headers: {
-        'Content-Type': 'application/json;charset=utf-8'
-      },
-      body: JSON.stringify(command)
-    };
+    const options = this.requestPostOptions(command);
```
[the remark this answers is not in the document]
Amit3200: done.

packages/webdriver-utils/src/driver.js
```diff
@@ -10,6 +10,16 @@ export default class Driver {
     this.passedCapabilities = passedCapabilities;
   }
 
+  requestPostOptions(command) {
```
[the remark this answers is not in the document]
Amit3200: done.

packages/webdriver-utils/src/providers/genericProvider.js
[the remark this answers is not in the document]
Amit3200: This is further simplified to a simple function and logic is shifted to single function.

itsjwala (approved):

itsjwala (approved):

prklm10 (approved):

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -33,6 +33,12 @@ export default class GenericProvider {
     this.debugUrl = null;
     this.header = 0;
     this.footer = 0;
+    this.statusBarHeight = 0;
+    this.pageXShiftFactor = 0;
+    this.pageYShiftFactor = 0;
+    this.currentTag = null;
+    this.removeElementShiftFactor = 50000;
```
nilshah98, line 40: What's the use of `removeElementShiftFactor` ?

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -172,30 +193,67 @@ export default class GenericProvider {
...
-      ...customRegions
-    ];
+    let isRegionPassed = [xpaths, selectors, elements, customLocations].some(regions => regions.length > 0);
+    if (isRegionPassed) {
+      await this.doTransformations();
+      const xpathRegions = await this.getSeleniumRegionsBy('xpath', xpaths);
+      const selectorRegions = await this.getSeleniumRegionsBy('css selector', selectors);
+      const elementRegions = await this.getSeleniumRegionsByElement(elements);
+      const customRegions = await this.getSeleniumRegionsByLocation(customLocations);
+      await this.undoTransformations('.poa-injected');
```
nilshah98, line 233: Have we made sure there are no other scripts or styles having the class `poa-injected` ?
Since those will also get removed.

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -172,30 +193,67 @@ export default class GenericProvider {
...
-  async getRegionObject(selector, elementId) {
-    const scaleFactor = parseInt(await this.metaData.devicePixelRatio());
-    const rect = await this.driver.rect(elementId);
-    const location = { x: rect.x, y: rect.y };
-    const size = { height: rect.height, width: rect.width };
+  async getRegionObjectFromBoundingBox(selector, element) {
+    const scaleFactor = await this.metaData.devicePixelRatio();
+    let headerAdjustment = 0;
+    if (this.currentTag.osName === 'iOS') {
+      headerAdjustment = this.statusBarHeight;
```
nilshah98, line 250: I think in older iphones, we might need footerAdjustment as well

nilshah98 (commented):
If this passes QA, can you please -
1. Create confluence doc for the new workflows ?
2. Create a P3 ticket to optimise this ?

packages/webdriver-utils/src/providers/genericProvider.js
Amit3200: All the `.poa-injected` things will need to be removed by this point they all are removed on Terminal. So we used same selector for consistency making sure that we don't impact anything else.

packages/webdriver-utils/src/providers/genericProvider.js
Amit3200: We had ran older builds on Iphone as far as we see there was no any big impact noted till now.

packages/webdriver-utils/src/providers/genericProvider.js
Amit3200: This is to make the coordinates incorrect when the element is not visible on screen. So when scroll values don't match the top and bottom becomes negative causing it to be 0 and hence not drawing it.

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -191,16 +198,30 @@ export default class GenericProvider {
     ];
   }
 
+  async updatePageShiftFactor(location) {
+    const scrollFactors = await this.driver.executeScript({ script: 'return [parseInt(window.scrollX * window.devicePixelRatio), parseInt(window.scrollY * window.devicePixelRatio)];', args: [] });
+    this.pageYShiftFactor = this.currentOperatingSystem === 'iOS' ? this.statusBarHeight : (this.statusBarHeight - scrollFactors.value[1]);
+    this.pageXShiftFactor = this.currentOperatingSystem === 'iOS' ? 0 : (-scrollFactors.value[0]);
```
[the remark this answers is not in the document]
Amit3200: Verified few builds on it however exhaustive runs were not done.

nilshah98 (approved):

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -63,6 +69,18 @@ export default class GenericProvider {
     }
   }
 
+  async getInitialPosition() {
+    if (this.currentTag.osName === 'iOS') {
```
prklm10, line 73: can do this to lower case and check for `ios`. So that it doesn't get missed on some future upgrade that changes this osName.

Amit3200: @prklm10  `iOS` is being set by us manually in the normalization layer.
Here - https://github.com/percy/cli/blob/b78c8d644e43da7eefce79b7a6ff8a9714d4398f/packages/webdriver-utils/src/metadata/normalizeData.js#L8

prklm10: got it 🍍.

itsjwala (approved):

PR #1427, opened 2023-11-03
Fixing percyScreenshotEnd response

rishigupta1599:
In percyScreenshotEnd we currently check for `response?.body?.link` which is returned as response of postComparisons. In App Percy it works because they create comparisons before and send it's response back whereas in POA the case is not same and we create comparison later hence we send different response which does not have link causing status to be marked as failure everytime.
Now made check based on error object that if errorObj is present then only mark as failure else success. As we don't have percyScreenshotUrl in POA when this is called we are passing build url as alternative.

### DIFF
--- a/packages/webdriver-utils/src/providers/automateProvider.js
+++ b/packages/webdriver-utils/src/providers/automateProvider.js
@@ -69,7 +69,7 @@ export default class AutomateProvider extends GenericProvider {
       error = e;
       throw e;
     } finally {
-      await this.percyScreenshotEnd(name, response?.body?.link, `${error}`);
+      await this.percyScreenshotEnd(name, error);
     }
     return response;
   }
@@ -93,14 +93,14 @@ export default class AutomateProvider extends GenericProvider {
     });
   }
 
-  async percyScreenshotEnd(name, percyScreenshotUrl, statusMessage = null) {
+  async percyScreenshotEnd(name, error) {
     return await TimeIt.run('percyScreenshotEnd', async () => {
       try {
         await this.browserstackExecutor('percyScreenshot', {
           name,
-          percyScreenshotUrl,
-          status: percyScreenshotUrl ? 'success' : 'failure',
-          statusMessage,
+          percyScreenshotUrl: this.buildInfo.url,
+          status: error ? 'failure' : 'success',
+          statusMessage: error ? `${error}` : '',
           state: 'end'
         });
       } catch (e) {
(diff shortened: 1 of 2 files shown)


Amit3200 (approved):

packages/webdriver-utils/src/providers/automateProvider.js
```diff
@@ -93,14 +93,14 @@ export default class AutomateProvider extends GenericProvider {
...
-  async percyScreenshotEnd(name, percyScreenshotUrl, statusMessage = null) {
+  async percyScreenshotEnd(name, error) {
     return await TimeIt.run('percyScreenshotEnd', async () => {
       try {
         await this.browserstackExecutor('percyScreenshot', {
           name,
-          percyScreenshotUrl,
-          status: percyScreenshotUrl ? 'success' : 'failure',
-          statusMessage,
+          percyScreenshotUrl: this.buildInfo.url,
```
Amit3200, line 101: Just thinking if this should be `null` safe.

packages/webdriver-utils/test/providers/automateProvider.test.js
```diff
@@ -29,21 +29,21 @@ describe('AutomateProvider', () => {
     });
 
     it('throws Error when called without initializing driver', async () => {
-      let automateProvider = new AutomateProvider('1234', 'https://localhost/command-executor', { platform: 'win' }, {}, {}, 'client', 'environment', {}, percyBuildInfo);
+      let automateProvider = new AutomateProvider('1234', 'https://localhost/command-executor', { platform: 'win' }, {}, {}, 'client', 'environment', percyBuildInfo);
```
Amit3200, line 32: Do we want to change this?, the options ideally are the screenshot options, we should not remove it because
this will make `options` = `percyBuildInfo` where as `percyBuildInfo` = `undefined` unless we name make it named parameter. This gives an unclear code flow view.

rishigupta1599: AutomateProvider constructor expects 8 parameters whereas we were sending 9, now percyBuildInfo is passed correctly at 8th position.
I checked locally using logging the value for percyBuildInfo was coming as `{}` whereas we were passing it but at wrong position.

packages/webdriver-utils/src/providers/automateProvider.js
rishigupta1599: Done

Amit3200 (approved):

shantanuk-browserstack (approved):

itsjwala (approved):

PR #1440, opened 2023-11-22
Resolving race condition in snapshots flush

rishigupta1599:
Current flushing logic wait till only the last dequeued element to be resolve the promise and as soon as that happens we take early exit. Now it can be the case that there are still the snapshots in the queue which did not got flushed. Now if queue is marked as end, we will finalize the build, but as snpashots queue still has some elements it will try to finalize those after build is finalized till the time queue is cleared.

### DIFF
--- a/packages/core/test/unit/queue.test.js
+++ b/packages/core/test/unit/queue.test.js
@@ -1,11 +1,13 @@
 import { AbortController, generatePromise, waitForTimeout } from '../../src/utils.js';
 import Queue from '../../src/queue.js';
+import { logger, setupTest } from '../helpers/index.js';
 
 describe('Unit / Tasks Queue', () => {
   let q;
 
-  beforeEach(() => {
+  beforeEach(async () => {
     q = new Queue('test');
+    await setupTest();
   });
 
   it('has a customizable concurrency', () => {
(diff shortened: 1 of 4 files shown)


ninadbstack (approved):
LGTM

packages/core/test/unit/queue.test.js
```diff
@@ -401,24 +401,32 @@ describe('Unit / Tasks Queue', () => {
...
 
-    resolve2();
-    await expectAsync(p2).toBeResolved();
+  it('empty flush', async () => {
+    q.set({ concurrency: 1 });
+    await q.start();
+    q.readyState = 2;
+    expect(q.size).toBe(0);
+    let promise = generatePromise(q.flush());
+    await expectAsync(promise).toBePending();
```
itsjwala, line 428: Update expectation with debug logs added

itsjwala (approved):

packages/core/test/unit/queue.test.js
rishigupta1599: Done

packages/cli-app/package.json
```diff
@@ -9,7 +9,7 @@
   },
   "publishConfig": {
     "access": "public",
-    "tag": "beta"
+    "tag": "alpha"
```
itsjwala, line 12: revert package.json changes before merging

itsjwala (approved):

PR #1441, opened 2023-11-23
♻️  Refactor: added defaultmap + locking mechanism for intercepts queue

nilshah98:
**Context-**
- Percy relies on CDP messages to intercept and capture network requests.
- CDP uses websockets, and since websockets run on TCP, at network layer it is guaranteed that the packets will always be transmitted in order.
- But at application level, packets can be processed out of order which can cause issues.
- Hence, adding a locking mechanism so that events are always processed in order at Application layer as well.
- The ideal flow of messages should be -
- - `requestWillBeSent` -> `requestPaused` -> `responseReceived` -> `loadingFinished` / `loadingFailed`

**Note:** Also merged #1443 in this PR.

### DIFF
--- a/packages/core/src/network.js
+++ b/packages/core/src/network.js
@@ -13,10 +13,10 @@ const ABORTED_MESSAGE = 'Request was aborted by browser';
 // ServiceWorker flow:  requestWillBeSent -> responseReceived -> loadingFinished / loadingFailed
 class RequestLifeCycleHandler {
   constructor() {
-    this.releaseRequest = null;
-    this.releaseResponse = null;
-    this.checkRequestSent = new Promise((resolve) => (this.releaseRequest = resolve));
-    this.checkResponseSent = new Promise((resolve) => (this.releaseResponse = resolve));
+    this.resolveRequestWillBeSent = null;
+    this.resolveResponseReceived = null;
+    this.requestWillBeSent = new Promise((resolve) => (this.resolveRequestWillBeSent = resolve));
+    this.responseReceived = new Promise((resolve) => (this.resolveResponseReceived = resolve));
   }
 }
 // The Interceptor class creates common handlers for dealing with intercepting asset requests
@@ -167,7 +167,7 @@ export class Network {
     let { networkId: requestId, requestId: interceptId, resourceType } = event;
 
     // wait for request to be sent
-    await this.#requestsLifeCycleHandler.get(requestId).checkRequestSent;
+    await this.#requestsLifeCycleHandler.get(requestId).requestWillBeSent;
     let pending = this.#pending.get(requestId);
     this.#pending.delete(requestId);
 
@@ -192,7 +192,7 @@ export class Network {
     // release request
     // note: we are releasing this, even if intercept is not set for network.js
     // since, we want to process all-requests in-order doesn't matter if it should be intercepted or not
-    this.#requestsLifeCycleHandler.get(requestId).releaseRequest();
+    this.#requestsLifeCycleHandler.get(requestId).resolveRequestWillBeSent();
   }
 
   // Called when a pending request is paused. Handles associating redirected requests with
@@ -225,7 +225,7 @@ export class Network {
     let { requestId, response } = event;
     // await on requestWillBeSent
     // no explicitly wait on requestWillBePaused as we implictly wait on it, since it manipulates the lifeCycle of request using Fetch module
-    await this.#requestsLifeCycleHandler.get(requestId).checkRequestSent;
+    await this.#requestsLifeCycleHandler.get(requestId).requestWillBeSent;
     let request = this.#requests.get(requestId);
     /* istanbul ignore if: race condition paranioa */
     if (!request) return;
@@ -236,7 +236,7 @@ export class Network {
       return Buffer.from(result.body, result.base64Encoded ? 'base64' : 'utf-8');
     };
     // release response
-    this.#requestsLifeCycleHandler.get(requestId).releaseResponse();
+    this.#requestsLifeCycleHandler.get(requestId).resolveResponseReceived();
   }
 
   // Called when a request streams events. These types of requests break asset discovery because
@@ -244,7 +244,7 @@ export class Network {
   _handleEventSourceMessageReceived = async event => {
     let { requestId } = event;
     // wait for request to be sent
-    await this.#requestsLifeCycleHandler.get(requestId).checkRequestSent;
+    await this.#requestsLifeCycleHandler.get(requestId).requestWillBeSent;
     let request = this.#requests.get(requestId);
     /* istanbul ignore else: race condition paranioa */
     if (request) this._forgetRequest(request);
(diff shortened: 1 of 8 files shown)


packages/core/src/network.js
```diff
@@ -8,16 +8,50 @@ const ALLOWED_STATUSES = [200, 201, 301, 302, 304, 307, 308];
 const ALLOWED_RESOURCES = ['Document', 'Stylesheet', 'Image', 'Media', 'Font', 'Other'];
 const ABORTED_MESSAGE = 'Request was aborted by browser';
 
+// DefaultMap, which returns a default value for an uninitialized key
+// Similar to defaultDict in python
+class DefaultMap extends Map {
```
ninadbstack, line 13: Should move this to a common package/utils

packages/core/src/network.js
```diff
@@ -8,16 +8,50 @@ const ALLOWED_STATUSES = [200, 201, 301, 302, 304, 307, 308];
...
+
+// RequestLifeCycleHandler handles life cycle of a requestId
+// Ideal flow:          requestWillBeSent -> requestPaused -> responseReceived -> loadingFinished / loadingFailed
+// ServiceWorker flow:  requestWillBeSent -> responseReceived -> loadingFinished / loadingFailed
+class RequestLifeCycleHandler {
+  constructor() {
+    this.releaseRequest = null;
+    this.releaseResponse = null;
+    this.checkRequestSent = new Promise((resolve) => (this.releaseRequest = resolve));
+    this.checkResponseSent = new Promise((resolve) => (this.releaseResponse = resolve));
```
ninadbstack, line 41: Can we try to make these names similar to actual lifecycle events ? so that its easier to understand. 

await requestWillBeSent; is better to read than await checkRequestSent imo. [ and reduces efforts needed to map it back to actual

packages/core/src/network.js
```diff
@@ -237,7 +270,10 @@ export class Network {
   // Called when a request has finished loading which triggers the this.onrequestfinished
   // callback. The request should have an associated response and be finished with any redirects.
   _handleLoadingFinished = async event => {
-    let request = this.#requests.get(event.requestId);
+    let { requestId } = event;
+    // wait for upto 2 seconds or check if response has been sent
+    await Promise.any([this.#requestsLifeCycleHandler.get(requestId).checkResponseSent, new Promise((resolve) => setTimeout(resolve, 2000))]);
```
ninadbstack, line 275: shouldnt _handleLoadingFailed wait for an event as well ?

packages/core/src/network.js
```diff
@@ -8,16 +8,50 @@ const ALLOWED_STATUSES = [200, 201, 301, 302, 304, 307, 308];
...
+  }
+}
 // The Interceptor class creates common handlers for dealing with intercepting asset requests
 // for a given page using various devtools protocol events and commands.
 export class Network {
   static TIMEOUT = undefined;
 
   log = logger('core:discovery');
 
+  #requestsLifeCycleHandler = new DefaultMap(() => new RequestLifeCycleHandler());
```
ninadbstack, line 51: when is something deleted from this map ?
also considering we have this, we dont need any state sets right ?

nilshah98: > when is something deleted from this map ?

Nothing is ever deleted from `#requestsLifeCycleHandler` since it is only used to manage lifecycles.
We can ideally delete `requestId` once all the locks are released from it's object, ie. it has reached end of lifecycle.
But, I don't see the point of it, as this is a very light map.

> also considering we have this, we dont need any state sets right ?

Need `#pending` to communicate from `requestWillBeSent` to `requestPaused` states.
Can remove `#intercepts` completely.
Need `#requests` as it keeps mapping of all active requests and is what `idle` depends on.
Need `#authentication` as this keeps track of all requests that have already been authenticated.
Need `#aborted` to keep track of aborted requests.

In the current PR, I am planning to only remove `#intercepts` queue as it was used for communicating from `requestPaused` to `requestWillBeSent` but that should never happen now.

packages/core/src/network.js
nilshah98: Yep, will add these in the next iteration.
Wanted to get a thumbs up on the implementation before proceeding further.

packages/core/src/network.js
nilshah98: Sure, let name them as - `checkRequestWillBeSent` and `releaseRequestWillBeSent` ?

ninadbstack: `await requestWillBeSent` means "wait until `requestWillBeSent` event". So we dont need to add `check` prefix before it. In fact check makes it slightly confusing as it reads as "wait until checkRequestWillBeSent event". So my preference remains fro first [ even though its not blocking ] 

Similarly instead of `releaseRequestWillBeSent` would prefer `resolveRequestWillBeSent` as it means the same but one uses same word as in general used in code.

packages/core/src/network.js
ninadbstack: `as this is a very light map.` do not fully agree with this as we could have thousands of requests in a single page load [ in fact we do for many storybook builds ] [ even though not blocking as it would be still few kbs max ]

`_requestPaused` could `await requestWillBeSent`, so it doesnt need to read from `#pending` as far as i understand. Because we are also storing event in pending with request id, you can just `resolve` with `event` for request will be sent 
`#requests` agree we need it for now at least
`#authentication` could be done by checking if auth promise is already resolved [ but i dont mind using set either ]
`#aborted` is not necessary as you could just set `aborted` on `#request` object [ or in lifecycle like lifecyle.aborted ] it was done the way it is because we could process `aborted` event before `requestWillBeSent` and it wont exist in requests at all but as `loadingFailed` would await on `requestWillBeSent` we can be sure that request already exists in requests

packages/core/src/network.js
```diff
@@ -1,23 +1,35 @@
...
   #pending = new Map();
   #requests = new Map();
-  #intercepts = new Map();
+  // #intercepts = new Map();
```
ninadbstack, line 32: lets remove it entirely before merge

packages/core/src/network.js
```diff
@@ -213,8 +220,11 @@ export class Network {
 
   // Called when a response has been received for a specific request. Associates the response with
   // the request data and adds a buffer method to fetch the response body when needed.
-  _handleResponseReceived = (session, event) => {
+  _handleResponseReceived = async (session, event) => {
     let { requestId, response } = event;
+    // await on requestWillBeSent
+    // no explicitly wait on requestWillBePaused as we implictly wait on it, since it manipulates the lifeCycle of request using Fetch module
+    await this.#requestsLifeCycleHandler.get(requestId).requestWillBeSent;
```
ninadbstack, line 227:
```suggestion
    await this.#requestsLifeCycleHandler.get(requestId).requestWillBeSent;
```
[optional] add a method `lifecycle(requestId)` to shorten `this.#requestsLifeCycleHandler.get(requestId)`

ninadbstack (approved):
LGTM! with an optional comment

itsjwala (approved):

PR #1443, opened 2023-11-27
Capture service worker intercepted requests

nilshah98:
**Changes**:
- [x] Added `captureMockedServiceWorker` flag under `discovery` for capturing service worker requests
- [x]  Set `bypass: false` for `Network.setBypassServiceWorker`  so that service-worker controller can be added and we can access the MSW mocks customer has added.
- [x] Added `serviceWorker` parameter, to handleRequests that are not intercepted _(since service worker intercepted and will resolve them)_

**Detailed Changes & Reasoning**:
- Added `Network.setBypassServiceWorker` to `bypass: false` so that we can add service-worker controller and access service worker that mocks customer's data.
- Initially tried to access service-worker requests using `Network.getResponseBody`, but over here `body` is returned as `""` _(empty string)_ randomly.
- - Tried to look on the internet about why this happens, but no luck :(
- Finally, decided to call `Network.getResponseBody` on the original request which service-worker intercepted.
- - `Network.getResponseBody` is same as what appears in the `Response` tab in Network section in devtools.
- - For requests that were intercepted by service-worker, the `Response` tab is **always** populated.
- - But the service-worker which makes the request upstream or returns mocked data randomly returns `body: ""` when `Network.getResponseBody` is called on it.
- - Hence, relying on calling `Network.getResponseBody` on the original request itself.

**Related PRs**:
https://github.com/percy/percy-storybook/pull/845

### DIFF
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -52,6 +52,7 @@ function debugSnapshotOptions(snapshot) {
   debugProp(snapshot, 'discovery.requestHeaders', JSON.stringify);
   debugProp(snapshot, 'discovery.authorization', JSON.stringify);
   debugProp(snapshot, 'discovery.disableCache');
+  debugProp(snapshot, 'discovery.captureServiceWorker');
   debugProp(snapshot, 'discovery.userAgent');
   debugProp(snapshot, 'clientInfo');
   debugProp(snapshot, 'environmentInfo');
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -289,6 +289,7 @@ export function createDiscoveryQueue(percy) {
         requestHeaders: snapshot.discovery.requestHeaders,
         authorization: snapshot.discovery.authorization,
         userAgent: snapshot.discovery.userAgent,
+        captureServiceWorker: snapshot.discovery.captureServiceWorker,
         meta: snapshot.meta,
 
         // enable network inteception
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -52,7 +52,7 @@ function debugSnapshotOptions(snapshot) {
   debugProp(snapshot, 'discovery.requestHeaders', JSON.stringify);
   debugProp(snapshot, 'discovery.authorization', JSON.stringify);
   debugProp(snapshot, 'discovery.disableCache');
-  debugProp(snapshot, 'discovery.captureServiceWorker');
+  debugProp(snapshot, 'discovery.captureMockedServiceWorker');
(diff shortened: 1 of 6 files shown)


packages/core/src/discovery.js
```diff
@@ -288,6 +289,7 @@ export function createDiscoveryQueue(percy) {
         requestHeaders: snapshot.discovery.requestHeaders,
         authorization: snapshot.discovery.authorization,
         userAgent: snapshot.discovery.userAgent,
+        captureServiceWorker: snapshot.discovery.captureServiceWorker,
```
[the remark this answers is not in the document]
itsjwala: `snapshot.discovery.captureServiceWorker` will contain global value if, can you verify once?

packages/core/src/snapshot.js
```diff
@@ -114,6 +114,7 @@ function getSnapshotOptions(options, { config, meta }) {
       requestHeaders: config.discovery.requestHeaders,
       authorization: config.discovery.authorization,
       disableCache: config.discovery.disableCache,
+      captureServiceWorker: config.discovery.captureServiceWorker,
```
itsjwala, line 117: okay we're overriding it here so no issues.

itsjwala (commented):
can you link locking CLI pr?

nilshah98:
> can you link locking CLI pr?

The base branch for this one is https://github.com/percy/cli/tree/refactor-network-js.
Hence, for merging as well, the PR is created for merging into https://github.com/percy/cli/tree/refactor-network-js.

packages/core/test/discovery.test.js
```diff
@@ -1970,4 +1970,68 @@ describe('Discovery', () => {
...
+        event.waitUntil(clients.claim());
+      });
+      `]);
+
+      server.reply('/app.js', () => [200, 'text/javascript', dedent`
+      const registerServiceWorker = async () => {
+        await navigator.serviceWorker.register('sw.js',{ scope: './', });
+      };
+      
+      await registerServiceWorker();
```
ninadbstack, line 1996: didnt know we can call await in base js context

nilshah98: The script declaration is done as -
```html
    <script type="module" src="app.js"></script>
```

Here, `type="module"` allows for using `await` at top-level.
If we remove `type="module"` it throws error as -

`Uncaught SyntaxError: await is only valid in async functions and the top level bodies of modules (at app.js:5:1)`

ninadbstack: Oh nice, did not know this :P

ninadbstack (approved):

PR #1447, opened 2023-11-28
Added support to set network interface to bind

ninadbstack:


itsjwala (approved):

misterrick:
@itsjwala @ninadbstack Just a heads up that the change from `localhost` to `0.0.0.0` broke our CI pipeline. Maybe the default should be kept as `localhost` or this should have been published as a breaking change. We've fixed it for now by adding `PERCY_SERVER_HOST=localhost`, but this may affect others, too.

PR #1458, opened 2023-12-08
Add support for Scrollable single element screenshot

chinmay-browserstack:
Added new option scope-options in which user can pass scroll: true/false to take scrollable single element screenshot properly

### DIFF
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -300,6 +300,7 @@ export class PercyClient {
     name,
     widths,
     scope,
+    scopeOptions,
     minHeight,
     enableJavaScript,
     enableLayout,
@@ -329,6 +330,7 @@ export class PercyClient {
           name: name || null,
           widths: widths || null,
           scope: scope || null,
+          'scope-options': scopeOptions || {},
           'minimum-height': minHeight || null,
           'enable-javascript': enableJavaScript || null,
           'enable-layout': enableLayout || false
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -253,13 +253,7 @@ export const snapshotSchema = {
         domTransformation: { $ref: '/config/snapshot#/properties/domTransformation' },
         enableLayout: { $ref: '/config/snapshot#/properties/enableLayout' },
         reshuffleInvalidTags: { $ref: '/config/snapshot#/properties/reshuffleInvalidTags' },
-        scopeOptions: {
-          type: 'object',
-          additionalProperties: false,
-          properties: {
-            scroll: { $ref: '/config/snapshot#/properties/scopeOptions/properties/scroll' }
-          }
-        },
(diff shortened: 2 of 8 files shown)


packages/core/src/config.js
```diff
@@ -59,6 +59,15 @@ export const configSchema = {
       scope: {
         type: 'string'
       },
+      scopeOptions: {
+        type: 'object',
```
ninadbstack, line 63: For all objects we should define valid keys for the same, so that users can understand if they have typos

chinmay-browserstack: This are valid keys for given option https://github.com/percy/cli/pull/1458/files#diff-e7c1269ddb64f1de14c2c79cac204c16cfc3f8d8535512d218f8e4ad8efc63b6R66-R68. anything else need to be added?

packages/core/src/config.js
```diff
@@ -59,6 +59,15 @@ export const configSchema = {
...
+        additionalProperties: false,
+        properties: {
+          scroll: {
+            type: 'boolean'
```
itsjwala, line 67: can you update typescript definition index.ts?

packages/core/src/config.js
```diff
@@ -59,6 +59,15 @@ export const configSchema = {
...
+          }
+        }
+      },
```
itsjwala, line 70: add specs for same, with and without this config.

packages/core/src/discovery.js
```diff
@@ -42,6 +42,7 @@ function debugSnapshotOptions(snapshot) {
   debugProp(snapshot, 'deviceScaleFactor');
   debugProp(snapshot, 'waitForTimeout');
   debugProp(snapshot, 'waitForSelector');
+  debugProp(snapshot, 'scopeOptions.scroll');
```
itsjwala, line 45: update specs

packages/client/src/client.js
```diff
@@ -329,6 +330,7 @@ export class PercyClient {
           name: name || null,
           widths: widths || null,
           scope: scope || null,
+          'scope-options': scopeOptions || {},
```
itsjwala, line 333: add specs when value given

packages/core/src/config.js
```diff
@@ -244,6 +253,13 @@ export const snapshotSchema = {
         domTransformation: { $ref: '/config/snapshot#/properties/domTransformation' },
         enableLayout: { $ref: '/config/snapshot#/properties/enableLayout' },
         reshuffleInvalidTags: { $ref: '/config/snapshot#/properties/reshuffleInvalidTags' },
+        scopeOptions: {
```
itsjwala, line 256: it would have same definition as L62 ( make it similar to above lines)

chinmay-browserstack: It is same, and this is how it is defined for other properties

itsjwala (approved):
we'll check test failure separately as its not related to this PR, merging to unblock

PR #1466, opened 2023-12-14
Fixing ignoreRegions for POA Full Page

rishigupta1599:
Refactoring code to support ignoreRegions for Full Page Screenshot.
In full page screenshot Webpage can be at any scrolling state like (x,y), but the screenshot comes from the top (0,0), in that case proper scrolling factor need to be added in coordinates to get them wrt top of the DOM.

### DIFF
--- a/packages/webdriver-utils/src/providers/genericProvider.js
+++ b/packages/webdriver-utils/src/providers/genericProvider.js
@@ -69,11 +69,15 @@ export default class GenericProvider {
     }
   }
 
+  isIOS() {
+    return this.currentTag?.osName === 'iOS';
+  }
+
   async getScrollDetails() {
     return await this.driver.executeScript({ script: 'return [parseInt(window.scrollX), parseInt(window.scrollY)];', args: [] });
   }
 
-  async getInitialScrollPosition() {
+  async getInitialScrollLocation() {
     if (this.initialScrollLocation) {
       return this.initialScrollLocation;
     }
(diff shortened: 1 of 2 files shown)


packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -69,15 +69,27 @@ export default class GenericProvider {
     }
   }
 
+  async getScrollDetails() {
+    return await this.driver.executeScript({ script: 'return [parseInt(window.scrollX), parseInt(window.scrollY)];', args: [] });
+  }
+
+  async getInitialScrollFactor() {
```
ninadbstack, line 76: why are we calling it scroll factor ? factor is usually ratio right ? we are just storing initial scroll location

rishigupta1599: Done changed name to getInitialScrollLocation

packages/webdriver-utils/src/providers/genericProvider.js
```diff
@@ -69,15 +69,27 @@ export default class GenericProvider {
...
+  }
+
   async getInitialPosition() {
     if (this.currentTag.osName === 'iOS') {
-      this.initialScrollFactor = await this.driver.executeScript({ script: 'return [parseInt(window.scrollX), parseInt(window.scrollY)];', args: [] });
+      this.initialScrollLocation = await this.getInitialScrollPosition();
     }
   }
 
-  async scrollToInitialPosition(x, y) {
```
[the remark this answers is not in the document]
rishigupta1599: Done

ninadbstack (approved):

PR #1478, opened 2023-12-22
✨ Add file size logging for tiles and resources

itsjwala:
* Adds logs of following format for `resources` and `tiles` upload endpoints. 

```
[percy:client] Uploading 692B resource:
```

* Note: it logs file size post `base64` encoding is done, which is 30% increase in file size.
* we've `25mb` file size per resource limit restriction so this log line will further help in identifying the slow uploading resource.

### DIFF
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -274,6 +274,7 @@ export class PercyClient {
     let encodedContent = base64encode(content);
 
     this.log.debug(`Uploading ${formatBytes(encodedContent.length)} resource: ${url}...`);
+    this.mayBeLogUploadSize(encodedContent.length);
 
     return this.post(`builds/${buildId}/resources`, {
       data: {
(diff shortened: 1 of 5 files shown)


packages/client/src/client.js
```diff
@@ -269,15 +270,17 @@ export class PercyClient {
   // created from `content` if one is not provided.
   async uploadResource(buildId, { url, sha, filepath, content } = {}) {
     validateId('build', buildId);
-    this.log.debug(`Uploading resource: ${url}...`);
     if (filepath) content = await fs.promises.readFile(filepath);
+    let encodedContent = base64encode(content);
+
+    this.log.debug(`Uploading ${formatBytes(encodedContent.length)} resource: ${url}...`);
```
ninadbstack, line 276: Can we also add a warning if this is over 20MB and an error text warning if its over 25M?

prklm10 (approved):
lgtm 👍🏼

prklm10 (approved):

packages/client/src/client.js
itsjwala: added

PR #1485, opened 2024-01-02
sending cli error message in send events and fixing stdin bug

prklm10:
sending cli error message in send events and fixing stdin bug
- Not sending cli error when `PERCY_CLIENT_ERROR_LOGS` is set to `false`.
- Otherwise sending error to send events.
- Fixing stdin bug introduced  in `1.27.4`.


packages/cli-exec/src/exec.js
```diff
@@ -121,13 +125,17 @@ async function* spawn(cmd, args, percy) {
     proc.on('close', code => {
       closed = code;
       if (code !== 0) {
+        // Only send cli error when PERCY_CLIENT_ERROR_LOGS is set to true
+        if (process.env.PERCY_CLIENT_ERROR_LOGS && process.env.PERCY_CLIENT_ERROR_LOGS === 'false') {
```
ninadbstack, line 129:
```suggestion
        if (process.env.PERCY_CLIENT_ERROR_LOGS === 'false') {
```

packages/cli-exec/src/exec.js
```diff
@@ -121,13 +125,17 @@ async function* spawn(cmd, args, percy) {
...
+          errorMessage = '1';
```
ninadbstack, line 130:
```suggestion
          errorMessage = 'Log sharing is disabled';
```

packages/cli-exec/test/test_prog.js
```diff
@@ -0,0 +1,9 @@
+// Error
```
ninadbstack, line 1: move this to test/test_data folder

packages/cli-exec/src/exec.js
prklm10: updated.

packages/cli-exec/src/exec.js
prklm10: Updated

packages/cli-exec/test/test_prog.js
prklm10: done

PR #1486, opened 2024-01-04
🔖 v1.27.7-beta.0

prklm10:
version bump v1.27.7-beta.0


shantanuk-browserstack:
we'll also need to update tag from latest to beta  as we are releasing a beta version.

shantanuk-browserstack (approved):

PR #1487, opened 2024-01-05
feat: first commit: rough changes, added support for SYNC CLI

this-is-shivamsingh:
- This is the support for SYNC CLI
- Need to await till the time snapshot processing is completed and then send the snapshot details to the user.

Changes - 
1.  added promise if enableSync is passed
2. attached resolve, reject promise to snapshot object, that will be used after get processing status as true, and fetched the snapshot data.
3. send it snapshot to `percySyncQueue`, after createSnapshot is done

### DIFF
--- a/packages/core/src/api.js
+++ b/packages/core/src/api.js
@@ -91,9 +91,9 @@ export function createPercyServer(percy, port) {
     })
   // post one or more snapshots, optionally async
     .route('post', '/percy/snapshot', async (req, res) => {
-      let snapshot = req.body.enableSync
+      let snapshot = req.body['enableSync']
         ? await new Promise((resolve, reject) => percy.snapshot(req.body, { resolve, reject }))
-        : percy.snapshot(req.body);
+        : percy.snapshot(req.body)
 
       if (!req.url.searchParams.has('async')) await snapshot;
       return res.json(200, { success: true });
--- a/packages/core/src/api.js
+++ b/packages/core/src/api.js
@@ -91,7 +91,10 @@ export function createPercyServer(percy, port) {
     })
   // post one or more snapshots, optionally async
     .route('post', '/percy/snapshot', async (req, res) => {
-      let snapshot = percy.snapshot(req.body);
+      let snapshot = req.body['enableSync']
+        ? await new Promise((resolve, reject) => percy.snapshot(req.body, { resolve, reject }))
+        : percy.snapshot(req.body)
+
       if (!req.url.searchParams.has('async')) await snapshot;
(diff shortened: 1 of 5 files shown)


packages/core/src/api.js
```diff
@@ -15,6 +15,10 @@ function encodeURLSearchParams(subj, prefix) {
   )).join('&') : `${prefix}=${encodeURIComponent(subj)}`;
 }
 
+function promiseWrapper(fn, ...params) {
```
ninadbstack, line 18:
```suggestion
function promisify(fn, ...params) {
```

packages/core/src/api.js
```diff
@@ -91,14 +95,20 @@ export function createPercyServer(percy, port) {
     })
   // post one or more snapshots, optionally async
     .route('post', '/percy/snapshot', async (req, res) => {
-      let snapshot = percy.snapshot(req.body);
+      let snapshot = req.body.enableSync
+        ? await promiseWrapper(percy.snapshot, req.body)
+        : percy.snapshot(req.body);
```
ninadbstack, line 100: We dont need to call in 2 ways, we can just call once and choose to await depending on sync. It makes it easier to maintain when more params are added

packages/core/src/api.js
```diff
@@ -91,14 +95,20 @@ export function createPercyServer(percy, port) {
...
+
       if (!req.url.searchParams.has('async')) await snapshot;
-      return res.json(200, { success: true });
+      return res.json(200, { success: true, data: snapshot });
```
ninadbstack, line 103: this will return incorrect values in default mode. need to validate it

packages/core/src/percy.js
```diff
@@ -262,7 +262,7 @@ export class Percy {
   // Takes one or more snapshots of a page while discovering resources to upload with the resulting
   // snapshots. Once asset discovery has completed for the provided snapshots, the queued task will
   // resolve and an upload task will be queued separately.
-  snapshot(options) {
+  snapshot(options, callback = null) {
```
ninadbstack, line 265: incorrect usage of standard terms. `callbacks` are fundamentally different than promises in how they should be used and what they do. Dont use that word here.  Use something else, maybe just resolve, reject as two args

packages/core/src/percy.js
```diff
@@ -322,7 +324,7 @@ export class Percy {
   }
 
   // Uploads one or more snapshots directly to the current Percy build
-  upload(options) {
+  upload(options, callback = null) {
```
ninadbstack, line 327: same here.

PR #1501, opened 2024-01-24
Add timeout in sync cli

chinmay-browserstack:
Added timeout for comparison of 90 seconds and 08: 30 min for snapshot

### DIFF
--- a/packages/core/src/wait-for-job.js
+++ b/packages/core/src/wait-for-job.js
@@ -4,6 +4,7 @@ const MIN_POLLING_INTERVAL = 5_000;
 // Poll atleast once in 2 min
 const MAX_POLLING_INTERVAL_SECONDS = 120;
 const THRESHOLD_OPTIMAL_POLL_TIME = 5;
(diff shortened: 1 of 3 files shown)


packages/core/src/wait-for-job.js
```diff
@@ -4,6 +4,7 @@ const MIN_POLLING_INTERVAL = 5_000;
 // Poll atleast once in 2 min
 const MAX_POLLING_INTERVAL_SECONDS = 120;
 const THRESHOLD_OPTIMAL_POLL_TIME = 5;
+const JOB_TIMEOUT = Number(process.env.SYNC_TIMEOUT) || 90_000;
```
ninadbstack, line 7: is 90 sec enough for full page ?

chinmay-browserstack: This time doesn't include time to capture screenshot. this is only waiting for api to respond. 90 second is enough for api to compare

ninadbstack (approved):

PR #1507, opened 2024-01-30
Adding support of shadowrootmode

rishigupta1599:
From Chromium 121 onwards shadowroot attribute support is removed and replaced with shadowrootmode so need to give support for it.

### DIFF
--- a/packages/dom/test/inject-polyfill.test.js
+++ b/packages/dom/test/inject-polyfill.test.js
@@ -18,6 +18,18 @@ describe('injectDeclarativeShadowDOMPolyfill', () => {
                 </div>
         </template>
         </div>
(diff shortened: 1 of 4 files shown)


packages/dom/test/inject-polyfill.test.js
```diff
@@ -7,9 +7,13 @@ describe('injectDeclarativeShadowDOMPolyfill', () => {
   beforeEach(() => {
     dom = withExample(`
         <div data-percy-shadow-host>
-        <template shadowroot="open">
```
ninadbstack, line 10: considering that we have code for both, we should have test for both

ninadbstack (approved):

PR #1510, opened 2024-01-31
🧹 Added logging of body for failed requests

ninadbstack:
We are seeing cases when a mitm proxy in between is either sending 403s or just blocking directly. Currently due to insufficient logging it becomes hard to understand if the 403 was due to incorrect/invalid percy token or due to a proxy in between, we are adding response body to the logs so that logs can be used to debug error right away.


pankaj443 (approved):
LGTM

PR #1522, opened 2024-02-17
Adding check for correct SDK for percy product

rishigupta1599:
Added validation for the case where the correct token and cli flow is used but sdk used is wrong.
For exp: -> App Percy Token is used and cli server is also started using app:exec which will create correct App Percy Build but as percy_screenshot is common in both App Percy and Percy on Automate it can be possible to trigger percy_screenshot using Automate SDK which should not be the case ideally.

We are relying on build type and the endpoint called to determine which SDK would have been used and in which build

### DIFF
--- a/packages/core/src/snapshot.js
+++ b/packages/core/src/snapshot.js
@@ -313,6 +313,7 @@ export function createSnapshotsQueue(percy) {
         let { data } = await percy.client.createBuild({ projectType: percy.projectType });
         let url = data.attributes['web-url'];
         let number = data.attributes['build-number'];
+        percy.client.buildType = data.attributes?.type;
         Object.assign(build, { id: data.id, url, number });
         // immediately run the queue if not delayed or deferred
         if (!percy.delayUploads && !percy.deferUploads) queue.run();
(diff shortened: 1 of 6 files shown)


packages/core/src/snapshot.js
```diff
@@ -361,6 +362,11 @@ export function createSnapshotsQueue(percy) {
     .handle('task', async function*({ resources, ...snapshot }) {
       let { name, meta } = snapshot;
 
+      if (percy.client.screenshotFlow === 'automate' && percy.client.buildType !== 'automate') {
+        throw new Error(`Cannot run automate screenshots in ${percy.client.buildType} project. Please use automate project token`);
+      } else if (percy.client.screenshotFlow === 'app' && percy.client.buildType !== 'app') {
+        throw new Error(`Cannot run App Percy screenshots in ${percy.client.buildType} project. Please use App Percy project token`);
```
prklm10, line 368: We can simplify this to below
```suggestion
     const errorMessage = percy.client.screenshotFlow === 'automate' ? 'automate' : 'App Percy';
const projectType = percy.client.screenshotFlow === 'automate' ? 'automate' : 'app';

if (percy.client.buildType !== projectType) {
  throw new Error(`Cannot run ${errorMessage} screenshots in ${percy.client.buildType} project. Please use ${errorMessage} project token`);
}
```

rishigupta1599: Did not do that reason being web projects also goes through this flow, and didn't wanted to touch web projects as it can have two flows snapshot and exec and will increase the blast radius and we already have a validation for it on sdk side and api side.

prklm10: Ack.

prklm10 (approved):

chinmay-browserstack:
So we already have checks in all SDK's if command called and projectType is same. if it's not same it should throw error. so this might not be required.

chinmay-browserstack (approved):

PR #1524, opened 2024-02-20
Throwing error when sha, filepath or content is not present

prklm10:
Throwing error when sha, filepath or content is not present when creating a comparison.
This fixes the below error, when sha is empty, for app build.
```
Error: Finalizing build 32741978 failed: cannot finalize before all snapshots have comparisons created. This is likely an SDK error, please make sure that comparisons are created before calling finalize.
```


rishigupta1599 (approved):
LGTM

PR #1527, opened 2024-02-21
Test Case implementation

prklm10:
Test Case implementation
Adding support for same snapshot name across different test-cases.


packages/core/src/discovery.js
```diff
@@ -317,7 +317,7 @@ export function createDiscoveryQueue(percy) {
     .handle('error', ({ name, meta }, error) => {
       if (error.name === 'AbortError' && queue.readyState < 3) {
         // only error about aborted snapshots when not closed
-        percy.log.error('Received a duplicate snapshot, ' + (
+        percy.log.error('Received a duplicate snapshot within same test case, ' + (
```
ninadbstack, line 320: not everyone would be using testcases, so the wording should depend on if testCase was null or not

packages/core/test/snapshot.test.js
```diff
@@ -837,14 +837,62 @@ describe('Snapshot', () => {
...
+      'dom-snapshot': testDOM
+    }, {
+      url: 'http://localhost:8000/one',
+      testCase: 'test-case-4',
+      domSnapshot: JSON.stringify({ html: testDOM })
+    }]);
+
+    expect(logger.stderr).toEqual([]);
+    expect(logger.stdout).toEqual([
+      '[percy] Snapshot taken: /one',
```
ninadbstack, line 889: we need to update logging and `meta` to include testcase name

prklm10: updated meta as well as logging.

packages/core/src/discovery.js
prklm10: Fixed logging

PR #1530, opened 2024-02-22
Added support for capturing image srcset

chinmay-browserstack:
Added support for capturing image srcset using `captureAllSrcsetUrls` flag.

### DIFF
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -187,6 +187,10 @@ export const configSchema = {
         type: 'boolean',
         default: false
       },
+      catureImageSrcSet: {
+        type: 'boolean',
+        default: false
+      },
       requestHeaders: {
         type: 'object',
         normalize: false,
@@ -272,6 +276,7 @@ export const snapshotSchema = {
             authorization: { $ref: '/config/discovery#/properties/authorization' },
             disableCache: { $ref: '/config/discovery#/properties/disableCache' },
             captureMockedServiceWorker: { $ref: '/config/discovery#/properties/captureMockedServiceWorker' },
+            catureImageSrcSet: { $ref: '/config/discovery#/properties/catureImageSrcSet' },
             userAgent: { $ref: '/config/discovery#/properties/userAgent' },
             devicePixelRatio: { $ref: '/config/discovery#/properties/devicePixelRatio' }
           }
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -187,7 +187,7 @@ export const configSchema = {
         type: 'boolean',
         default: false
       },
-      catureImageSrcSet: {
+      captureAllSrcsetURLs: {
         type: 'boolean',
         default: false
       },
@@ -276,7 +276,7 @@ export const snapshotSchema = {
             authorization: { $ref: '/config/discovery#/properties/authorization' },
             disableCache: { $ref: '/config/discovery#/properties/disableCache' },
             captureMockedServiceWorker: { $ref: '/config/discovery#/properties/captureMockedServiceWorker' },
-            catureImageSrcSet: { $ref: '/config/discovery#/properties/catureImageSrcSet' },
+            captureAllSrcsetURLs: { $ref: '/config/discovery#/properties/captureAllSrcsetURLs' },
             userAgent: { $ref: '/config/discovery#/properties/userAgent' },
             devicePixelRatio: { $ref: '/config/discovery#/properties/devicePixelRatio' }
           }
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -190,7 +190,7 @@ export const configSchema = {
         type: 'boolean',
         default: false
       },
-      captureAllSrcsetUrls: {
+      captureSrcset: {
         type: 'boolean',
         default: false
       },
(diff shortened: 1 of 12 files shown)


packages/core/src/discovery.js
```diff
@@ -94,6 +97,42 @@ function parseDomResources({ url, domSnapshot }) {
   }, new Map([[rootResource.url, rootResource]]));
 }
 
+async function parseImageSrcset({ imageLinks }, { resources, meta }, { allowedHostnames }) {
+  let log = logger('core:snapshot');
+  const promiseList = [];
+
+  imageLinks = imageLinks.filter((link) => hostnameMatches(allowedHostnames, link));
```
Amit3200, line 104: will it cover picture, video, source tags as well?

packages/core/src/utils.js
```diff
@@ -18,6 +19,14 @@ export function hostname(url) {
   return new URL(url).hostname;
 }
 
+// Return mime-type from a URL.
+export function detectMimeType(link) {
+  // mime package does not handle query params
+  let urlObj = new URL(link);
+  let urlWithoutSearchParams = urlObj.origin + urlObj.pathname;
+  return mime.lookup(urlWithoutSearchParams);
```
Amit3200, line 27: This doesn't fail correct?, if it fails what happens?

packages/dom/test/serialize-image-srcset.test.js
```diff
@@ -0,0 +1,51 @@
...
+    it(`${platform}: capture url from source of picture`, async () => {
+      withExample(`
+      <picture>
+        <source srcset='//locahost:9876/base/test/assets/example.webp, //localhost:9876/base/test/assets/example.png 2x' />
+        <source srcset='//locahost:9876/base/test/assets/example.jpeg 100px, //locahost:9876/base/test/assets/example1.jpeg 200px' />
+      </picture>
+    `);
+
+      serialized = serializeDOM();
+      console.log(serialized.imageLinks);
```
Amit3200, line 27: remove this.

packages/core/src/utils.js
chinmay-browserstack: This is existing logic which is used in network.js for each resource so it shouldn't fail.

packages/core/src/discovery.js
chinmay-browserstack: video tags are handled separately where we capture poster. poster can only be one. picture case is handled

packages/core/src/config.js
```diff
@@ -190,6 +190,10 @@ export const configSchema = {
         type: 'boolean',
         default: false
       },
+      captureAllSrcsetUrls: {
```
ninadbstack, line 193:
```suggestion
      captureSrcset: {
```
easier to type and conveys same [ as we dont plan to introduce capture a few from src set ]

packages/core/src/discovery.js
```diff
@@ -95,6 +98,45 @@ function parseDomResources({ url, domSnapshot }) {
   }, new Map([[rootResource.url, rootResource]]));
 }
 
+async function parseImageSrcset(snapshot, discovery) {
+  let log = logger('core:snapshot');
```
ninadbstack, line 102: shouldnt logger be discovery ?

packages/core/src/discovery.js
```diff
@@ -95,6 +98,45 @@ function parseDomResources({ url, domSnapshot }) {
...
+  const { resources, meta, domSnapshot } = snapshot;
+  const { allowedHostnames } = discovery;
+  let { imageLinks } = domSnapshot;
+
+  imageLinks = imageLinks.filter((link) => hostnameMatches(allowedHostnames, link));
+  imageLinks = imageLinks.filter((link) => !resources.get(link));
+
+  log.debug(`Capturing image src set: ${JSON.stringify(imageLinks)}`, meta);
+  for (const link of imageLinks) {
+    const promise = makeRequest(link, { buffer: true, retries: 0 }, (body, response) => {
```
ninadbstack, line 113: This may or may not work depending on the Auth requirements of the backend. I would recommend doing this in discovery browser as that would auto take care of auth requirements, else we will have to replicate this. 

Also `request` by default uses proxy if set, which may or may not be required for browser [ in fact could have issues if requests go via proxy ]

packages/core/src/discovery.js
```diff
@@ -95,6 +98,45 @@ function parseDomResources({ url, domSnapshot }) {
   }, new Map([[rootResource.url, rootResource]]));
 }
 
+async function parseImageSrcset(snapshot, discovery) {
```
ninadbstack, line 101:
```suggestion
async function gatherSrcsetResources(snapshot, discovery) {
```
Its not parsing srcset for image urls

packages/dom/src/serialize-image-srcset.js
```diff
@@ -0,0 +1,33 @@
+// Return list of link of all srcset images to be captured
+export function serializeImageSrcSet({ dom }) {
```
ninadbstack, line 2: This requires extra queries on dom and extra iterations on dom. So far we have always tried to add to cloneNode function which anyway's does one tree walk on the node to avoid loading same nodes again. That could potentially give slight performance boost.

packages/core/src/discovery.js
chinmay-browserstack: Actually calling this urls using fetch api inside browser is causing CORS issue, that's y written logic in such a way

packages/core/src/discovery.js
chinmay-browserstack: All logger in this file uses `core:snapshot`

packages/dom/src/serialize-image-srcset.js
chinmay-browserstack: resolved

PR #1533, opened 2024-02-28
Force quit CLI post 10s

pankaj443:
JIRA: https://browserstack.atlassian.net/browse/PER-3007
We noticed CI getting stuck due to cp.spawn processes
Added a 10s timout post which we force quit.


packages/cli-exec/src/exec.js
```diff
@@ -92,7 +92,14 @@ export const exec = command('exec', {
   await percy?.stop(!!error);
 
   // forward any returned status code
-  if (status) exit(status, error);
+  if (status) {
+    exit(status, error);
+  } else {
+    // force exit post timeout
+    setTimeout(function() {
+      process.exit(0);
```
ninadbstack, line 100: we still need to exit with forwarded status code

packages/cli-exec/src/exec.js
```diff
@@ -92,7 +92,14 @@ export const exec = command('exec', {
...
+    }, 10000);
+  }
```
ninadbstack, line 102: better promisify and await on sleep time of 10 secs. Then we can do standard process.exit

like 
```
await 10 sec 
process.exit(0)
```

also dont need to put it in `else`

ninadbstack (approved):
LGTM

PR #1536, opened 2024-02-28
Added support for capturing asset in multiple dpr and width combination

chinmay-browserstack:
Added new flag `captureResponsiveAssets` which will capture assets in all dpr, width combination supported by current project


packages/core/src/snapshot.js
```diff
@@ -118,6 +118,7 @@ function getSnapshotOptions(options, { config, meta }) {
       disableCache: config.discovery.disableCache,
       captureMockedServiceWorker: config.discovery.captureMockedServiceWorker,
       captureSrcset: config.discovery.captureSrcset,
+      captureResponsiveAssets: config.discovery.captureResponsiveAssets,
```
Amit3200, line 121: This should not be needed, we should call this explicitly again if there is `enableJS` true. If we want to keep it then we need confirmation from product.

ninadbstack: Ideally we should do this automatically.

packages/core/src/discovery.js
```diff
@@ -138,11 +139,24 @@ function processSnapshotResources({ domSnapshot, resources, ...snapshot }) {
   return { ...snapshot, resources };
 }
 
+async function responsiveDomParams(snapshot, client) {
```
ninadbstack, line 142: Not something we want to call per snapshot, move it to the snapshot config validation and cache the response in percy.js so that it can be applied to all snapshots in current build

packages/client/test/client.test.js
```diff
@@ -353,6 +353,20 @@ describe('PercyClient', () => {
...
   });
 
+  describe('#getDeviceDetails()', () => {
+    it('in case of error return []', async () => {
+      api.reply('/device-details', () => [500]);
+      await expectAsync(client.getDeviceDetails()).toBeResolvedTo([]);
+    });
+
+    it('gets device details', async () => {
+      api.reply('/device-details', () => [200, { data: '<<device-data-without-build-id>>' }]);
```
ninadbstack, line 363:
```suggestion
      api.reply('/device-details', () => [200, { data: ['<<device-data-without-build-id>>' ]}]);
```

packages/core/src/discovery.js
```diff
@@ -317,7 +324,8 @@ export function createDiscoveryQueue(percy) {
       try {
         yield* captureSnapshotResources(page, snapshot, {
           captureWidths: !snapshot.domSnapshot && percy.deferUploads,
-          capture: callback
+          capture: callback,
+          captureResponsiveDom: percy.deviceDetails || []
```
ninadbstack, line 328: captureForDevices

packages/core/src/discovery.js
```diff
@@ -198,6 +198,16 @@ async function* captureSnapshotResources(page, snapshot, options) {
       }
     }
 
+    if (captureResponsiveDom) {
+      for (const device of captureResponsiveDom) {
+        yield waitForDiscoveryNetworkIdle(page, discovery);
```
ninadbstack, line 203:
```suggestion
        // We are not adding these widths and pixels ratios in loop above because we want to explicitly reload the page after resize which we dont do above
        yield waitForDiscoveryNetworkIdle(page, discovery);
```

packages/core/src/discovery.js
```diff
@@ -212,13 +222,10 @@ async function* captureSnapshotResources(page, snapshot, options) {
...
-  if (deviceScaleFactor !== discovery.devicePixelRatio) {
-    yield waitForDiscoveryNetworkIdle(page, discovery);
-
-    yield* captureSnapshotResources(page, snapshot, {
-      deviceScaleFactor: discovery.devicePixelRatio,
-      mobile: true
-    });
+  if (discovery.devicePixelRatio) {
+    const log = logger('core:discovery');
+    log.warn('discovery.devicePixelRatio is deprecated percy will now auto capture resource in all devicePixelRatio, Ignoring configuration');
```
ninadbstack, line 227:
```suggestion
    log.deprecate('discovery.devicePixelRatio is deprecated percy will now auto capture resource in all devicePixelRatio, Ignoring configuration');
```

packages/client/src/client.js
```diff
@@ -217,6 +217,18 @@ export class PercyClient {
     return this.get(`job_status?sync=true&type=${type}&id=${ids.join()}`);
   }
 
+  // Returns device details enabled on project associated with given token
+  async getDeviceDetails(buildId) {
+    try {
+      let url = 'device-details';
+      if (buildId) url += `?build_id=${buildId}`;
+      const { data } = await this.get(url);
```
Amit3200, line 225: This returns list only correct?

chinmay-browserstack: yes

ninadbstack (approved):

PR #1538, opened 2024-02-29
serialize base64 data URLs

pankaj443:
JIRA: https://browserstack.atlassian.net/browse/PER-3004

Customer was getting errors for large payload due to base64 images in dom. To fix it we have serialized base64 images into seperate resources which will reduce the main dom size.

Verified builds:
https://percy.io/percy/context-test-prod/builds/32908409/changed/1805493533?browser=chrome&browser_ids=56&subcategories=approved&viewLayout=side-by-side&viewMode=new&width=1280&widths=1280
https://percy.io/percy/context-test-prod/builds/32909537/changed/1805546877?browser=chrome&browser_ids=56&subcategories=approved&viewLayout=side-by-side&viewMode=new&width=1280&widths=1280


packages/dom/src/serialize-base64-images.js
```diff
@@ -0,0 +1,51 @@
...
+  for (let s in signatures) {
+    if (base64String.indexOf(s) === 0) {
+      return signatures[s];
+    }
+  }
+  return null;
+}
+
+// Captures the current frame of videos and sets the poster image
+export function serializeBase64Images(dom, resources) {
```
ninadbstack, line 31: Can we generalize this for all data urls ? instead of specifically only doing for base 64 and images ?

ninadbstack: We also have svgs in the same fashion with utf8 data

packages/dom/src/serialize-base64-images.js
```diff
@@ -0,0 +1,51 @@
+import { resourceFromText } from './utils';
+import { uid } from './prepare-dom';
+
+const signatures = {
+  JVBERi0: 'application/pdf',
+  R0lGODdh: 'image/gif',
+  R0lGODlh: 'image/gif',
+  iVBORw0KGgo: 'image/png',
+  '/9j/': 'image/jpg',
+  Qk02U: 'image/bmp'
```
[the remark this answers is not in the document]
pankaj443: this is only required for mimetype for extension, other way is to get string b/w data & `;base64,`

packages/dom/src/clone-dom.js
```diff
@@ -13,11 +14,12 @@ import applyElementTransformations from './transform-dom';
 
 const ignoreTags = ['NOSCRIPT'];
 
-export function cloneNodeAndShadow({ dom, disableShadowDOM }) {
+export function cloneNodeAndShadow({ dom, disableShadowDOM, resources }) {
   // clones shadow DOM and light DOM for a given node
   let cloneNode = (node, parent) => {
     let walkTree = (nextn, nextp) => {
       while (nextn) {
+        serializeBase64(nextn, resources);
```
ninadbstack, line 22: This is changing the node in original test browser. We want to update cloned node so that client test dom is not affected. Please move this under like 36 as previously discussed on call.

pankaj443: updated

ninadbstack (approved):

PR #1556, opened 2024-03-26
Bumping up chromium version to v123

Amit3200:
- Bumping up chromium version to v123
- This should update asset-discovery browser and should resolve shadow-root related issues.


pankaj443 (approved):
LGTM 🏎️

PR #1558, opened 2024-03-27
Remove `checked` attr for radio-button by default

shantanuk-browserstack:
In the new behaviour, if any of the radio-button have checked attribute present in the dom by default, 
it will be removed by default.

Only those radio-buttons will have `checked` attribute present in their rendered dom which are currently in checked state i.e. `element.checked` returning `true`

### DIFF
--- a/packages/dom/src/serialize-inputs.js
+++ b/packages/dom/src/serialize-inputs.js
@@ -7,6 +7,7 @@ export function serializeInputElements({ dom, clone, warnings }) {
     switch (elem.type) {
       case 'checkbox':
       case 'radio':
+        elem.removeAttribute("checked");
         if (elem.checked) {
           cloneEl.setAttribute('checked', '');
         }
--- a/packages/dom/src/serialize-inputs.js
+++ b/packages/dom/src/serialize-inputs.js
@@ -11,10 +11,11 @@ export function serializeInputElements({ dom, clone, warnings }) {
           here we are removing the checked attr if present by default,
           so that only the current selected radio-button will have the checked attr present in the dom
 
-          this happens because in html, when the checked attribute is present in the multiple radio-buttons by default,
-          the browser will only render the last checked radio-button as when the user is selecting any particular button,
-          the checked attribute on other buttons is not removed,
-          hence sometimes it shows inconsistent state even though the `element.checked` attribute returns correct state
+          this happens because in html, 
+          when the checked attribute is present in the multiple radio-buttons for which only one can be selected at a time,
+          the browser will only render the last checked radio-button by default,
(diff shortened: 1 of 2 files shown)


packages/dom/src/serialize-inputs.js
```diff
@@ -7,6 +7,7 @@ export function serializeInputElements({ dom, clone, warnings }) {
     switch (elem.type) {
       case 'checkbox':
       case 'radio':
+        elem.removeAttribute('checked');
```
ninadbstack, line 10: Need to add comment on why we are doing this

packages/dom/src/serialize-inputs.js
```diff
@@ -7,6 +7,16 @@ export function serializeInputElements({ dom, clone, warnings }) {
     switch (elem.type) {
       case 'checkbox':
       case 'radio':
+        /*
+          here we are removing the checked attr if present by default,
+          so that only the selected radio-button will have the checked attr present in the dom
+
+          we need to removed `checked` from both cloneEl & elem as if it is only removed from cloneEl,
+          then if any radio-button has `checked` attr present by default,
+          then it will be shown as checked even though it was not explicitly selected before taking snapshot
```
ninadbstack, line 16: Not very clear why it needs to be removed from both, we only serialize clonedel

packages/dom/test/serialize-inputs.test.js
```diff
@@ -87,6 +90,14 @@ describe('serializeInputs', () => {
       expect($('#radio')[0].checked).toBe(true);
     });
 
+    it(`${platform}: removes checked attr from radio-buttons if present by default`, () => {
```
ninadbstack, line 93: we also need to verify that the actually checked radio has correct property set

packages/dom/test/serialize-inputs.test.js
```diff
@@ -84,9 +91,20 @@ describe('serializeInputs', () => {
...
       expect($('#radio')[0].checked).toBe(true);
     });
 
+    it(`${platform}: removes checked attr from radio-buttons if present in dom by default but radio is not checked`, () => {
+      // validates `checked=""` is removed from the dom for option1 as it is not checked
+      expect($('#option1')[0].outerHTML).not.toContain('checked=""');
+      expect($('#option1')[0].checked).toBe(false);
+
+      // validates `checked=""` is not removed from the dom for option2 as it is in checked state
+      expect($('#option2')[0].outerHTML).toContain('checked=""');
```
ninadbstack, line 104: In our example, the 2nd anyway shows up as checked. We want to simulate a condition that post this someone selects first radio button and the same is selected in the serialized dom

ninadbstack (approved):

PR #1566, opened 2024-04-02
GZIP resources before uploading

pankaj443:
JIRA: https://browserstack.atlassian.net/browse/PER-3058
Now we are gziping resources before uploading if PERCY_GZIP env var is set. This is useful in cases where dom resources are greater than 25mb and large payload error is received.


packages/client/test/client.test.js
```diff
@@ -589,6 +590,27 @@ describe('PercyClient', () => {
         }
       });
     });
+
+    it('can upload a resource from a local path in GZIP format', async () => {
+      process.env.PERCY_GZIP = true;
```
ninadbstack, line 595: this should be moved to before and after block so that it can be cleared after current test.

ninadbstack (approved):

Issue #1560, opened 2024-03-27
Using PERCY_ENABLE can lead to NPE


SemaLuna:
## The problem

When using `PERCY_ENABLE` to disable Percy, the `exec` command unfortunately still has logic for trying to send buildEvents upstream (when there is an error).

So if the command that percy wraps on errors, we get a NPE.

Eventually leading to an NPE:
```
file:///PATH_TO_PROJECT/node_modules/.pnpm/@percy+cli-exec@1.28.0_typescript@5.4.2/node_modules/@percy/cli-exec/dist/exec.js:145
          percy.client.sendBuildEvents(percy.build.id, myObject);
                ^
TypeError: Cannot read properties of undefined (reading 'client')
    at ChildProcess.<anonymous> (file:///PATH_TO_PROJECT/node_modules/.pnpm/@percy+cli-exec@1.28.0_typescript@5.4.2/node_modules/@percy/cli-exec/dist/exec.js:145:17)
    at ChildProcess.emit (node:events:513:28)
    at maybeClose (node:internal/child_process:1100:16)
    at Socket.<anonymous> (node:internal/child_process:458:11)
    at Socket.emit (node:events:513:28)
    at Pipe.<anonymous> (node:net:301:12)
```

## Environment

- `@percy/cli` version: 1.28.0

## Code to reproduce issue

It is easy enough to notice the issue in the code - in [this line](https://github.com/percy/cli/blob/master/packages/cli-exec/src/exec.js#L145) we try to sendBuildEvents, however `percy` is undefined when `PERCY_ENABLE` is 0 (you can see this assumption in the same file [here](https://github.com/percy/cli/blob/master/packages/cli-exec/src/exec.js#L62)) 

Should be a simple fix.

prklm10:
Hi @SemaLuna, Sorry for the late revert we will fix look into this and the fix will be available in the next stable release.

SemaLuna:
@prklm10 no problem at all - thanks for the fix.

prklm10:
Hi @SemaLuna we have fixed this bug in 1.28.3.

prklm10:
Closing this issue.

PR #1569, opened 2024-04-10
🐛 fix: In sending events in cli-exec

prklm10:
🐛  fix: In sending events in cli-exec
Github Issue -> https://github.com/percy/cli/issues/1560
Details:-
- When percy is disabled ther is no percy object created so `percy.client` is throwing us an error.

### DIFF
--- a/packages/cli-exec/test/exec.test.js
+++ b/packages/cli-exec/test/exec.test.js
@@ -173,6 +173,25 @@ describe('percy exec', () => {
     });
   });
 
+  it('tests process.stderr percy is disabled', async () => {
+    delete process.env.PERCY_CLIENT_ERROR_LOGS;
+    process.env.PERCY_ENABLE = '0';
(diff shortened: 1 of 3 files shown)


shantanuk-browserstack (approved):

packages/cli-exec/test/exec.test.js
```diff
@@ -173,6 +173,25 @@ describe('percy exec', () => {
     });
   });
 
+  it('tests process.stderr percy is disabled', async () => {
+    delete process.env.PERCY_CLIENT_ERROR_LOGS;
+    process.env.PERCY_ENABLE = '0';
```
ninadbstack, line 178: this is not reset at the end of the test

prklm10: There is a after each block that deletes this delete process.env.PERCY_ENABLE;

shantanuk-browserstack (approved):

PR #1579, opened 2024-04-24
Cleanup InjectDeclarativeShadowDOMPolyfill script

pankaj443:
JIRA: https://browserstack.atlassian.net/browse/PER-3132
we have shifted this logic to RCW, removing from here


rishigupta1599 (approved):

packages/dom/src/serialize-frames.js
```diff
@@ -9,7 +9,7 @@ function setBaseURI(dom) {
   let $base = document.createElement('base');
   $base.href = dom.baseURI;
 
-  dom.querySelector('head').prepend($base);
+  dom.querySelector('head')?.prepend($base);
```
ninadbstack, line 12: instead of skipping this we can append on `dom.documentElement` equivalent right ?

pankaj443: Comment above says it has to be added on <head> tag.

ninadbstack: Oh, got it fair enough.

ninadbstack (approved):

PR #1580, opened 2024-04-25
Add missing Discovery option types

roykolak:
The `retry` option was recently added in #1529. 

Noticed that the `userAgent` option was also missing from the types.


roykolak:
Bump on this

PR #1597, opened 2024-05-17
Update github workflow to sign Windows executable

pankaj443:
JIRA: https://browserstack.atlassian.net/browse/PER-3213
 
Updated github workflow to sign Windows executable.


.github/workflows/executable.yml
```diff
@@ -19,12 +19,42 @@ jobs:
           APPLE_ID_KEY: ${{secrets.APPLE_ID_KEY}}
       - name: Verify executable
         run: ./percy --version
+      - name: Upload win artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: win-exe
+          path: percy.exe
```
prklm10, line 26: it should be opposite correct? win-exe is the path and percy.exe is the name

pankaj443: no, path is the path of created exe from the workflow which we have renamed to percy.exe
name is the name we want to give to the artifact while uploading.

ninadbstack (approved):

PR #1598, opened 2024-05-20
POA Playwright changes

rishigupta1599:
Changes: 
1. Made a common method captureScreenshot in WebdriverUtils which based on framework will decide to use the provider class to use for further execution.
2. Shifted few common methods like percyScreenshotBegin, percyScreenshotEnd, browserstackExecutor in genericProvider as it is being used as base class.
3. Also in constructor of these classes instead of taking each argument separately taking it as hash and assigning it to instance of object so that we can directly pass from child class without need to add every arguments from every child class.
4. For Playwright Provider we needed following things to be passed for execution 
     1. Page Guid
     2. MainFrame Guid
     3. browserstack sessionId to make percy call.

### DIFF
--- a/packages/core/src/api.js
+++ b/packages/core/src/api.js
@@ -136,14 +136,9 @@ export function createPercyServer(percy, port) {
     }))
     .route('post', '/percy/automateScreenshot', async (req, res) => {
       let data;
-      let comparisonData;
       percyAutomateRequestHandler(req, percy);
-      if(req.body.framework == 'playwright') {
-        comparisonData = await WebdriverUtils.playwrightScreenshot(req.body);
-      } else {
-        comparisonData = await WebdriverUtils.automateScreenshot(req.body);
-      }
-      
+      let comparisonData = await WebdriverUtils.captureScreenshot(req.body);
+
       if (percy.syncMode(comparisonData)) {
         const snapshotPromise = new Promise((resolve, reject) => percy.upload(comparisonData, { resolve, reject }, 'automate'));
         data = await handleSyncJob(snapshotPromise, percy, 'comparison');
--- a/packages/core/src/api.js
+++ b/packages/core/src/api.js
@@ -136,8 +136,14 @@ export function createPercyServer(percy, port) {
     }))
     .route('post', '/percy/automateScreenshot', async (req, res) => {
       let data;
+      let comparisonData;
       percyAutomateRequestHandler(req, percy);
-      let comparisonData = await WebdriverUtils.automateScreenshot(req.body);
+      if(req.body.framework == 'playwright') {
+        comparisonData = await WebdriverUtils.playwrightScreenshot(req.body);
+      } else {
+        comparisonData = await WebdriverUtils.automateScreenshot(req.body);
+      }
+      
       if (percy.syncMode(comparisonData)) {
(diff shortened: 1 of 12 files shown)


packages/core/src/api.js
```diff
@@ -136,8 +136,14 @@ export function createPercyServer(percy, port) {
     }))
     .route('post', '/percy/automateScreenshot', async (req, res) => {
       let data;
+      let comparisonData;
       percyAutomateRequestHandler(req, percy);
-      let comparisonData = await WebdriverUtils.automateScreenshot(req.body);
+      if(req.body.framework == 'playwright') {
```
Amit3200, line 141:
```suggestion
      if(req.body.framework === 'playwright') {
```

Amit3200, line 141: convert to `switch` case

Amit3200 (commented):
Overall Lgtm 🍍

Amit3200 (approved):

chinmay-browserstack (approved):

shantanuk-browserstack (approved):

packages/webdriver-utils/src/index.js
```diff
@@ -1,27 +1,40 @@
...
     try {
       const startTime = Date.now();
-      log.info(`[${snapshotName}] : Starting automate screenshot ...`);
-      const automate = ProviderResolver.resolve(sessionId, commandExecutorUrl, capabilities, sessionCapabilites, clientInfo, environmentInfo, options, buildInfo);
-      log.debug(`[${snapshotName}] : Resolved provider ...`);
-      await automate.createDriver();
+      log.info(`[${snapshotName}] : Starting automate screenshot capture ...`);
+
+      let provider;
+      switch (framework) {
```
Amit3200, line 26: eh one thing I missed we need to ensure that this is in lowercase.

packages/webdriver-utils/src/index.js
```diff
@@ -1,27 +1,40 @@
...
-  static async automateScreenshot({
+  static async captureScreenshot({
     sessionId,
     commandExecutorUrl,
     capabilities,
-    sessionCapabilites,
+    sessionCapabilities,
+    frameGuid,
+    pageGuid,
+    framework,
```
prklm10, line 13: Can we log frameGuid and pageGuid in debug logs. so that it is easier to debug.

packages/webdriver-utils/src/providers/providerResolver.js
```diff
@@ -2,9 +2,30 @@ import GenericProvider from './genericProvider.js';
...
+    sessionCapabilities,
+    clientInfo,
+    environmentInfo,
+    options,
+    buildInfo
+  ) {
     // We can safely do [0] because GenericProvider is catch all
-    const Klass = [AutomateProvider, GenericProvider].filter(x => x.supports(commandExecutorUrl))[0];
-    return new Klass(sessionId, commandExecutorUrl, capabilities, sessionCapabilities, clientInfo, environmentInfo, options, buildInfo);
+    const Klass = [AutomateProvider, GenericProvider].filter((x) =>
```
prklm10, line 16: `x` use a better variable like `provider`.

prklm10 (approved):

PR #1602, opened 2024-05-24
Handle error when styleSheets property is not defined

bkhanale:
Some websites restrict the use of properties on shadow dom by throwing an error, which causes the builds to break. We're handling this with an error handler for one such cases where `styleSheets` property was causing Percy to fail.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.28.7-beta.0",
+  "version": "1.28.7",
   "packages": [
     "packages/*"
   ],
--- a/packages/cli-app/package.json
+++ b/packages/cli-app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@percy/cli-app",
-  "version": "1.28.7-beta.0",
+  "version": "1.28.7",
   "license": "MIT",
   "repository": {
     "type": "git",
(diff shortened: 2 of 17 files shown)


packages/dom/src/serialize-cssom.js
```diff
@@ -27,36 +27,45 @@ function createStyleResource(styleSheet) {
...
-      if (styleSheetsMatch(styleSheet, styleSheetFromNode(cloneOwnerNode))) continue;
-      let style = document.createElement('style');
+  let styleSheets = null;
+  // catch error in case styleSheets property is not available (overwritten to throw error)
+  try {
+    styleSheets = dom.styleSheets;
+  } catch {
+    warnings.add('Skipping `styleSheets` as it is not supported.');
+  }
+  if (styleSheets) {
```
ninadbstack, line 37: Instead of adding this `if` and adding indentation to the code we can just add a return after line 35

ninadbstack (approved):
LGTM!

packages/dom/src/serialize-cssom.js
bkhanale: We also have a check for `dom.adoptedStyleSheets` below this, which we can still check for, hence not returning it from here.

ninadbstack: Oh, my bad, i thought function ends here. For future we should split both into separate functions.

PR #1610, opened 2024-06-05
supporting PERCY_DO_N0T_CAPTURE_RESPONSIVE_ASSETS env variable to disable fetching of responsive assets

prklm10:
supporting PERCY_DO_N0T_CAPTURE_RESPONSIVE_ASSETS env variable to disable fetching of responsive assets

### DIFF
--- a/packages/core/src/percy.js
+++ b/packages/core/src/percy.js
@@ -164,7 +164,11 @@ export class Percy {
       if (!this.skipDiscovery) yield this.#discovery.start();
       // start a local API server for SDK communication
       if (this.server) yield this.server.listen();
-      if (this.projectType === 'web') this.deviceDetails = yield this.client.getDeviceDetails(this.build?.id);
+      if (this.projectType === 'web') {
+        if (!process.env.PERCY_DISABLE_DEVICE_DETAIL_FETCH || process.env.PERCY_DISABLE_DEVICE_DETAIL_FETCH !== 'true') {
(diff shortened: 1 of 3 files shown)


packages/core/src/percy.js
```diff
@@ -164,7 +164,11 @@ export class Percy {
       if (!this.skipDiscovery) yield this.#discovery.start();
       // start a local API server for SDK communication
       if (this.server) yield this.server.listen();
-      if (this.projectType === 'web') this.deviceDetails = yield this.client.getDeviceDetails(this.build?.id);
+      if (this.projectType === 'web') {
+        if (!process.env.PERCY_DISABLE_DEVICE_DETAIL_FETCH || process.env.PERCY_DISABLE_DEVICE_DETAIL_FETCH !== 'true') {
```
ninadbstack, line 168: Not a very easy to understand flag for someone who doesnt know percy internals. Why should someone enable/disable it, what does it do ? 

use a better flag name, that allows user to toggle a functionality and not an input to a functionality

prklm10: renamed to `PERCY_DON0T_CAPTURE_RESPONSIVE_ASSETS`

ninadbstack: Let's make it DO_NOT as those are separate words.

ninadbstack (approved):

packages/core/src/percy.js
prklm10: updated

ninadbstack (approved):

PR #1622, opened 2024-06-18
✨  Analyze Error Logs and Suggest Fixes for Issues

this-is-shivamsingh:
## What is this?
- Sending error messages/logs to Percy-API, which will analyze and suggest some fixes for the issue that occurred.
- The SuggestionResponse will be an array where each suggestion item will contain 3 thing
      - Failure: Failure title
      - FailureReason: string
      - Suggestion/summary: string
      - Reference Doc Links: Array of string
- Errors are categorised in 2 cases
     - Build level error
     - Snapshot level error [ the format of response changes a little for this ]

Cases covered:
- Snapshot command not called
- Error occurred at starting Percy server
- Launching browser
- Error navigating to Page
- Page load failed
- Network idle timed-out
- Error occurred at asset discovery, taking Snapshot, uploading snapshot stage.
- Error occurred at uploading snapshot resource [ Pending: thinking of where to add this, as it doesn't throw any error only logs it ].
- Error verifying comparison tile failure.

Reference pic:
Screenshot 2024-06-21 at 1 06 05 PM

For snapshot level:
image
[ Faillure Reason is not there on the API side yet, so it is coming empty string 😓 ]

Testing scenario:
1. 1st and foremost priority is nothing should break on the existing flow. So an e2e
   - Web, automate, App Percy, ( using CLI commands like `npx Percy snapshot`, `npx percy exec:start` etc. )
2. Start the Percy server and terminate it. It should show [ No snapshot command was called ].
3. Port in use [ start a local server at 5338 ] and then start Percy server, it should hit Api for analysis.
4. Trying setting up `throw new Errors` at different places
    - Inside discover.js, snapshot.js
    - In browser.js [ under launch() function ]
    and check if the Send for log analysis is getting called or not.
5. You may encounter that for some errors, there will be no suggestions as they are not added in our rule engine. Most common errors as handled, and should show suggestions for.

Related PR: https://github.com/percy/percy-api/pull/3678

### DIFF
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -538,9 +538,9 @@ export class PercyClient {
       let errMsg = 'Uploading comparison tile failed';
 
       // Detecting error and logging fix for the same
-      this.log.error(errMsg);
-      await this.formatAndSendForAnalysis(errMsg);
-      return false;
+      // We are throwing this error as the comparison will be failed
+      // even if 1 tile gets failed
+      throw new Error(errMsg);
     }
     return true;
   }
@@ -612,7 +612,8 @@ export class PercyClient {
     });
   }
 
-  async formatAndSendForAnalysis(errorLogs) {
+  // private function
+  #formatLogErrors(errorLogs) {
     let errors = [];
     if (typeof errorLogs === 'string') {
       errors.push({
@@ -629,10 +630,11 @@ export class PercyClient {
       });
     }
 
-    return await this.sendErrorForAnalysis({ logs: errors });
+    return { logs: errors };
   }
 
-  async sendErrorForAnalysis(errorLogs) {
+  async getErrorAnalysis(errors) {
+    const errorLogs = this.#formatLogErrors(errors);
     this.log.debug('Sending error logs for analysis');
 
     return this.post('suggestions/from_logs', {
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -539,7 +539,7 @@ export class PercyClient {
 
       // Detecting error and printing Fix for the same
       this.log.error(errMsg);
-      await formatAndSendErrorLog(errMsg);
+      await this.formatAndSendErrorLog(errMsg);
       return false;
     }
     return true;
@@ -613,8 +613,8 @@ export class PercyClient {
   }
 
   async formatAndSendErrorLog(errorMsg, _options) {
-    let errors = []
-    if(typeof errorMsg == 'string') {
+    let errors = [];
+    if (typeof errorMsg === 'string') {
       errors.push(errorMsg);
     } else if (Array.isArray(errorMsg)) {
       errors = errorMsg;
@@ -622,14 +622,15 @@ export class PercyClient {
       errors.push(JSON.stringify(errorMsg));
     }
 
-    await this.sendErrorLogForSuggestion(errors)
+    await this.sendErrorLogForSuggestion(errors);
   }
 
   async sendErrorLogForSuggestion(errorLogs) {
-    this.log.debug('Sending error logs for analysis')
+    this.log.debug('Sending error logs for analysis');
+
     return this.post('suggestions/from_logs', {
       data: errorLogs
-    })
+    });
   }
 
   mayBeLogUploadSize(contentSize) {
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -535,7 +535,11 @@ export class PercyClient {
     while (retries > 0 && !success);
 
     if (!success) {
-      this.log.error('Uploading comparison tile failed');
+      let errMsg = 'Uploading comparison tile failed';
+
+      // Detecting error and printing Fix for the same
+      this.log.error(errMsg);
+      await formatAndSendErrorLog(errMsg);
       return false;
     }
     return true;
@@ -606,6 +610,26 @@ export class PercyClient {
     return this.post('logs', {
       data: body
     });
+  }
+
+  async formatAndSendErrorLog(errorMsg, _options) {
+    let errors = []
+    if(typeof errorMsg == 'string') {
+      errors.push(errorMsg);
+    } else if (Array.isArray(errorMsg)) {
+      errors = errorMsg;
+    } else {
+      errors.push(JSON.stringify(errorMsg));
+    }
+
+    await this.sendErrorLogForSuggestion(errors)
+  }
+
+  async sendErrorLogForSuggestion(errorLogs) {
+    this.log.debug('Sending error logs for analysis')
+    return this.post('suggestions/from_logs', {
+      data: body
+    })
   }
 
   mayBeLogUploadSize(contentSize) {
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -539,7 +539,7 @@ export class PercyClient {
 
       // Detecting error and printing Fix for the same
       this.log.error(errMsg);
-      await this.formatAndSendErrorLog(errMsg);
+      await this.formatAndSendForAnalysis(errMsg);
       return false;
     }
     return true;
@@ -612,20 +612,27 @@ export class PercyClient {
     });
   }
 
-  async formatAndSendErrorLog(errorMsg, _options) {
+  async formatAndSendForAnalysis(errorLogs) {
     let errors = [];
-    if (typeof errorMsg === 'string') {
-      errors.push(errorMsg);
-    } else if (Array.isArray(errorMsg)) {
-      errors = errorMsg;
+    if (typeof errorLogs === 'string') {
+      errors.push({
+        message: errorLogs
+      });
+    } else if (Array.isArray(errorLogs)) {
+      errors = errorLogs;
     } else {
-      errors.push(JSON.stringify(errorMsg));
+      errors.push({
+        message: errorLogs
+      });
+      errors.push({
+        message: errorLogs?.message || ''
+      });
     }
 
-    await this.sendErrorLogForSuggestion(errors);
+    return await this.sendErrorForAnalysis({ logs: errors });
   }
 
-  async sendErrorLogForSuggestion(errorLogs) {
+  async sendErrorForAnalysis(errorLogs) {
     this.log.debug('Sending error logs for analysis');
 
     return this.post('suggestions/from_logs', {
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -628,7 +628,7 @@ export class PercyClient {
   async sendErrorLogForSuggestion(errorLogs) {
     this.log.debug('Sending error logs for analysis')
     return this.post('suggestions/from_logs', {
-      data: body
+      data: errorLogs
     })
   }
 
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -13,7 +13,8 @@ import {
   getPackageJSON,
   waitForTimeout,
   validateTiles,
-  formatLogErrors
+  formatLogErrors,
+  tagsList
 } from './utils.js';
 
 // Default client API URL can be set with an env var for API development
@@ -55,10 +56,11 @@ export class PercyClient {
     clientInfo,
     environmentInfo,
     config,
+    labels,
     // versioned api url
     apiUrl = PERCY_CLIENT_API_URL
   } = {}) {
-    Object.assign(this, { token, config: config || {}, apiUrl });
+    Object.assign(this, { token, config: config || {}, apiUrl, labels: labels });
     this.addClientInfo(clientInfo);
     this.addEnvironmentInfo(environmentInfo);
     this.buildType = null;
@@ -135,6 +137,8 @@ export class PercyClient {
   async createBuild({ resources = [], projectType } = {}) {
     this.log.debug('Creating a new build...');
 
+    let tagsArr = tagsList(this.labels);
+
     return this.post('builds', {
       data: {
         type: 'builds',
@@ -153,7 +157,8 @@ export class PercyClient {
           'pull-request-number': this.env.pullRequest,
           'parallel-nonce': this.env.parallel.nonce,
           'parallel-total-shards': this.env.parallel.total,
-          partial: this.env.partial
+          partial: this.env.partial,
+          tags: tagsArr
         },
         relationships: {
           resources: {
@@ -368,6 +373,7 @@ export class PercyClient {
     environmentInfo,
     sync,
     testCase,
+    labels,
     thTestCaseExecutionId,
     resources = []
   } = {}) {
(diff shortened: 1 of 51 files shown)


packages/core/src/percy.js
```diff
@@ -423,12 +431,72 @@ export class Percy {
...
+        });
+      }
+    } catch (e) {
+      if (e.statusCode === 403) {
+        // Request Forbidden
+        // This can be due to proxy issue
+        this.log.error('percy.io might not be reachable, check network connection, proxy and ensure that percy.io is whitelisted. View link for details.');
+        this.log.error('When using a proxy, please configure the following environment variables: HTTP_PROXY, HTTPS_PROXY, and NO_PROXY.');
+      }
+      this.log.error('Unable to analyzing error logs');
```
ninadbstack, line 488:
```suggestion
      this.log.error('Unable to analyze error logs');
```

packages/client/src/client.js
```diff
@@ -608,6 +612,34 @@ export class PercyClient {
     });
   }
 
+  async formatAndSendForAnalysis(errorLogs) {
```
ninadbstack, line 615: format separate in utils - api call will remain in client

packages/client/src/client.js
```diff
@@ -535,7 +535,11 @@ export class PercyClient {
     while (retries > 0 && !success);
 
     if (!success) {
-      this.log.error('Uploading comparison tile failed');
```
ninadbstack, line 538: delete

ninadbstack: throw error here

packages/client/src/proxy.js
```diff
@@ -169,6 +169,7 @@ export class ProxyHttpsAgent extends https.Agent {
     let handleError = err => {
       socket.destroy(err);
       logger('client:proxy').error(`Proxying request failed: ${err}`);
+      logger('client:proxy').error('When using a proxy, please configure the following environment variables: HTTP_PROXY, HTTPS_PROXY, and NO_PROXY.');
```
ninadbstack, line 172: - Proxy auth creds are correct - 401/403 
- Proxy is correctly set and reachable 
- The domains percy.io/github/percy-enterprise.browserstack/storage.google.com are whitelisted on proxy

packages/core/src/snapshot.js
```diff
@@ -391,7 +392,7 @@ export function createSnapshotsQueue(percy) {
       return { ...snapshot, response };
     })
   // handle possible build errors returned by the API
-    .handle('error', (snapshot, error) => {
+    .handle('error', async (snapshot, error) => {
```
ninadbstack, line 395: In here we need some errors where we cannot connect percy - so we cannot call logs analysis either. 
it will be mostly found while create build or download of chromium and handle either here or in percy.js

packages/core/src/snapshot.js
```diff
@@ -413,13 +414,23 @@ export function createSnapshotsQueue(percy) {
       let duplicate = errors?.length > 1 && errors[1].detail.includes('must be unique');
       if (duplicate) {
         if (process.env.PERCY_IGNORE_DUPLICATES !== 'true') {
-          percy.log.warn(`Ignored duplicate snapshot. ${errors[1].detail}`);
+          let errMsg = `Ignored duplicate snapshot. ${errors[1].detail}`;
+          percy.log.warn(errMsg);
+
+          await percy.suggestionsForFix(errMsg, { snapshotLevel: true, snapshotName: name });
```
ninadbstack, line 420: send meta here.

packages/core/src/percy.js
```diff
@@ -185,8 +185,11 @@ export class Percy {
 
       // throw an easier-to-understand error when the port is in use
       if (error.code === 'EADDRINUSE') {
-        throw new Error('Percy is already running or the port is in use');
```
ninadbstack, line 188: Port number

packages/core/src/percy.js
```diff
@@ -423,12 +431,72 @@ export class Percy {
     return syncMode;
   }
 
+  async checkForNoSnapshotCommandError() {
+    // Max log line to check, ideally we get only 12 pecy:core
```
ninadbstack, line 435: logger.query(log => log.message.includes('Received Snapshot: '`).length === 0

packages/core/src/percy.js
```diff
@@ -423,12 +431,72 @@ export class Percy {
...
+            this.log.warn('Refer below Doc Links for the same');
+
+            referenceDocLinks?.forEach(_docLink => {
+              this.log.warn(`* ${_docLink}`);
+            });
+          }
+        });
+      }
+    } catch (e) {
+      if (e.statusCode === 403) {
```
ninadbstack, line 482: ECONNECT error handle here

packages/client/src/client.js
this-is-shivamsingh: Resolved.
- Now we are throwing an error here, which will be captured in snapshot.js under snapshotQueue queue(task), -> that will further get caught in error handler -> send for analysis and response back with the possible suggestion.

packages/client/src/client.js
```diff
@@ -608,6 +612,34 @@ export class PercyClient {
...
+      });
+      errors.push({
+        message: errorLogs?.message || ''
+      });
+    }
+
+    return await this.sendErrorForAnalysis({ logs: errors });
+  }
+
+  async sendErrorForAnalysis(errorLogs) {
```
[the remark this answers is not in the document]
this-is-shivamsingh: Resolved

packages/core/src/percy.js
this-is-shivamsingh: Resolved, added port number as well.

packages/core/src/percy.js
this-is-shivamsingh: Resolved 🧹

packages/core/src/percy.js
this-is-shivamsingh: Resolved 🧹

packages/core/src/snapshot.js
this-is-shivamsingh: Now, we are sending meta object with an error log message, I am not putting the `meta` object here as `discovery` and `snapshot` meta objects contain different values. I only want the snapshot name to display a different format in the CLI log with the snapshotName.

packages/core/src/percy.js
this-is-shivamsingh: Resolved. Correct

packages/core/src/snapshot.js
this-is-shivamsingh: Resolved. Handled in `percy.js`

packages/client/src/proxy.js
this-is-shivamsingh: Resolved.
- Added warning message for the same in `proxy.js`

packages/client/src/client.js
```diff
@@ -608,6 +612,36 @@ export class PercyClient {
     });
   }
 
+  // private function
+  #formatLogErrors(errorLogs) {
```
rishigupta1599, line 616: Shift it to utils instead can be used at other places as well

packages/core/src/discovery.js
```diff
@@ -359,15 +362,29 @@ export function createDiscoveryQueue(percy) {
         });
       });
     })
-    .handle('error', ({ name, meta }, error) => {
+    .handle('error', async ({ name, meta }, error) => {
```
rishigupta1599, line 365: Any impact due to async 🤔

ninadbstack (approved):

packages/client/src/proxy.js
```diff
@@ -169,6 +169,14 @@ export class ProxyHttpsAgent extends https.Agent {
...
       socket.destroy(err);
       logger('client:proxy').error(`Proxying request failed: ${err}`);
+
+      // We don't get statusCode here, relying on checking error message only
+      if (!!err.message && (err.message?.includes('ECONNREFUSED') || err.message?.includes('EHOSTUNREACH'))) {
+        logger('client:proxy').warn('If needed, Please verify if your proxy credentials are correct');
+        logger('client:proxy').warn('Please check if your proxy is set correctly and reachable');
+      }
+
+      logger('client:proxy').warn('Please check network connection, proxy and ensure that following domains are whitelisted: github.com, percy.io, storage.google.com. In case you are an enterprise customer make sure to whitelist "percy-enterprise.browserstack.com" as well.');
```
ninadbstack, line 179:
```suggestion
      logger('client:proxy').warn('Please check network connection, proxy and ensure that following domains are whitelisted: github.com, percy.io, storage.googleapis.com. In case you are an enterprise customer make sure to whitelist "percy-enterprise.browserstack.com" as well.');
```

packages/core/src/percy.js
```diff
@@ -423,12 +432,89 @@ export class Percy {
     return syncMode;
   }
 
+  // This specific error will be hard coded
+  async checkForNoSnapshotCommandError() {
```
rishigupta1599, line 436: Why it needs to be hardcoded? 
Even if snapshot command is not called build is still created right?

Only in case of defer-uploads it will not happen, in that case we can hard code right?

Do we really need logs for this 🤔

packages/client/src/proxy.js
```diff
@@ -169,6 +169,14 @@ export class ProxyHttpsAgent extends https.Agent {
     let handleError = err => {
       socket.destroy(err);
       logger('client:proxy').error(`Proxying request failed: ${err}`);
+
+      // We don't get statusCode here, relying on checking error message only
+      if (!!err.message && (err.message?.includes('ECONNREFUSED') || err.message?.includes('EHOSTUNREACH'))) {
+        logger('client:proxy').warn('If needed, Please verify if your proxy credentials are correct');
+        logger('client:proxy').warn('Please check if your proxy is set correctly and reachable');
```
ninadbstack, line 176:
```suggestion
        logger('client:proxy').warn('Please check if your proxy is set correctly and reachable [ currently set as <show proxy here redacting creds>');
```

packages/core/src/percy.js
```diff
@@ -423,12 +432,89 @@ export class Percy {
...
+
+        referenceDocLinks?.forEach(_docLink => {
+          this.log.warn(`* ${_docLink}`);
+        });
+      }
+    });
+  }
+
+  #proxyEnabled() {
+    return !!(process.env.HTTP_PROXY || process.env.HTTPS_PROXY);
```
ninadbstack, line 489: Dont duplicate code - please extract proxyUrl to a function in `getProxy` from client package and use it here

packages/core/src/percy.js
```diff
@@ -423,12 +432,89 @@ export class Percy {
...
+      const suggestionResponse = await this.client.getErrorAnalysis(errors);
+      this.#displaySuggestionLogs(suggestionResponse, options);
+    } catch (e) {
+      // Common error code for Proxy issues
+      const PROXY_CODES = ['ECONNREFUSED', 'ECONNRESET', 'EHOSTUNREACH'];
+      if (!!e.code && PROXY_CODES.includes(e.code)) {
+        // This can be due to proxy issue
+        this.log.error('percy.io might not be reachable, check network connection, proxy and ensure that percy.io is whitelisted.');
+        if (!this.#proxyEnabled()) {
+          this.log.error('If inside a proxied envirnment, please configure the following environment variables: HTTPS_PROXY, HTTP_PROXY. Refer to our documentation for more details');
```
ninadbstack, line 503:
```suggestion
          this.log.error('If inside a proxied envirnment, please configure the following environment variables: HTTP_PROXY, [ and optionally HTTPS_PROXY if you need it ]. Refer to our documentation for more details');
```
Most of the times people dont need second - and setting it would cause problems

packages/core/src/discovery.js
rishigupta1599: Ignore this, we do await in generatePromise so this should be fine, just run few builds to confirm everything is working fine

this-is-shivamsingh: - Initially, I thought of making it without async, the problem here is:
1. The message is not displayed at the proper location.
2. In some cases, even if the script exists, the request is yet not resolved.

Putting here async will not cause an issue, as we already await similarly at sendComparison, sendSnapshot, createSnapshot etc, in case of network request fails it is caught in `suggestForFix` only, so impacts here.

packages/core/src/percy.js
this-is-shivamsingh: This suggestion is not related to the build failure. Moreover, this is related if the snapshot command was called or not, which means the user is aware of the snapshot command.
- We handed it here as to find the condition for this issue is quite simple, and we will be saving 1 network request as well. Another benefit was, that to detect the no snapshot command called, we need to have the info logs. So we have to send 16 - 20 log lines to API to process and provide this error, which can be found here.
- Mostly we have 3 cases for for snapshot upload
1. through script
2. through CLI ( npx percy snapshot <yaml>
3. though cli-upload
All this will be covered here.

packages/core/src/percy.js
this-is-shivamsingh: Resolved 🧹

packages/client/src/client.js
this-is-shivamsingh: Shifted..., but it's a very specific function for formatting error logs. Don't know if it will work for other cases, as the payload may be different.

packages/client/src/client.js
```diff
@@ -617,6 +622,15 @@ export class PercyClient {
     });
   }
 
+  async getErrorAnalysis(errors) {
```
rishigupta1599, line 625: This can also be shift to utils right?

packages/client/src/proxy.js
```diff
@@ -169,6 +169,14 @@ export class ProxyHttpsAgent extends https.Agent {
...
+      }
+
+      logger('client:proxy').warn('Please check network connection, proxy and ensure that following domains are whitelisted: github.com, percy.io, storage.googleapis.com. In case you are an enterprise customer make sure to whitelist "percy-enterprise.browserstack.com" as well.');
```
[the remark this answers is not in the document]
this-is-shivamsingh: No, the user will reach at this point, if he already used `HTTP_PROXY, HTTPS_PROXY` env variables.

packages/client/src/client.js
this-is-shivamsingh: No, why?
It's used to hit Percy-api. It should be a client only.

packages/client/src/proxy.js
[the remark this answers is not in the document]
rishigupta1599: Then why are saying
`'Please check if your proxy is set correctly and reachable`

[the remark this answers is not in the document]
this-is-shivamsingh: This is regarding the user's proxy server that the user has given in `HTTPS_PROXY` is reachable or not. Whether that proxy is running or not.

rishigupta1599 (approved):

PR #1624, opened 2024-06-19
SDK logs support

ninadbstack:
Changes:
- Refactored logs collection to make it similar to any other logging in cli
- Added a POST `/percy/log` endpoint  for sdks to send extra debug data - currently we are missing all of that data and we cannot provide support without it.


this-is-shivamsingh (approved):
Lgtm 🌮 , Nice use of logger.query

pankaj443 (approved):
LGTM 🥊

PR #1625, opened 2024-06-19
tags support

shantanuk-browserstack:
We are now adding support of associating labels (tags) to builds & snapshots when they are getting created.

Labels can be associated by including `labels: dev,prod`
This can be done for both builds & snapshots.

To apply labels for snapshots, we need to include this in snapshot options.
To apply labels for builds, we need to include this in percy options.

we've also introduced a new flag `--labels` that can be included in the commands and then these labels will be associated to the build (ex: `--labels=dev,prod`)


packages/env/src/utils.js
```diff
@@ -74,3 +74,14 @@ export function stripQuotesAndSpaces(line) {
     return strippedLine;
   } else { return null; };
 }
+
+// convert tags comma-separated-names to array of objects for POST request
+export function tagsList(tags) {
```
ninadbstack, line 79: Not a correct place to keep this function - this page is for environment related functions - you can move it to client package as we only need it there - but later we can move to a `@percy/common` like package

packages/client/src/client.js
```diff
@@ -54,10 +55,11 @@ export class PercyClient {
     clientInfo,
     environmentInfo,
     config,
+    labels,
     // versioned api url
     apiUrl = PERCY_CLIENT_API_URL
   } = {}) {
-    Object.assign(this, { token, config: config || {}, apiUrl });
+    Object.assign(this, { token, config: config || {}, apiUrl, labels: labels });
```
ninadbstack, line 62:
```suggestion
    Object.assign(this, { token, config: config || {}, apiUrl, buildLabels: labels });
```

ninadbstack: As the usage is slightly confusing now - one using this.labels and other using labels

ninadbstack (approved):

Issue #1590, opened 2024-05-06
Add `percy exec` flag to avoid failure for Percy-related issues


geoffharcourt:
## The problem

Sometimes if we have some infrastructure failures that result in certain kinds of retries we end up with an issue where we get one of these errors:

```
[percy] Error: Can only finalize pending builds, current state: finished
```

We tried the approach of closing our builds with `percy build:finalize` and it seemed like it created more problems in our builds than before.

Could there be a flag for `percy exec` so that if there's a Percy-related issue (like an already finalized build) that the wrapped process would continue? We don't want our whole CI pipeline to fail over an issue related to snapshots (but we'd still like the error output so that we could look into the problem).

I was envisioning something like:

```shell
percy exec --proceed-on-failure -- bundle exec rake specs
```

This way we'd see the errors/warnings about issues in the build, but issues with retried parallel jobs (or situations where Percy's API is out of service) would be less likely to cause the spec portion of our builds to fail.

## Environment

- Node version: 20
- `@percy/cli` version: @percy/cli 1.28.5

## Details

If necessary, describe the problem you have been experiencing in more detail.

ninadbstack:
We will be taking this as a feature request. Thanks for raising this.

prklm10:
@geoffharcourt We have released cli version v1.28.9-beta.1 . You can set env variable `PERCY_EXIT_WITH_ZERO_ON_ERROR` = true. That will exit with code 0 on any percy related errors. The error logs will be present.
Also if your command that is passed in `exec --` fails than that will not be exit with code 0.

geoffharcourt:
Following up: this was much appreciated, thank you!

PR #1647, opened 2024-07-01
✨  Adding env variable support so that if set to true we exit with status code 0 when any percy related error occurs.

prklm10:
Adding env variable support so that if set to true we exit with status code 0 when any percy related error occurs.
`PERCY_EXIT_WITH_ZERO_ON_ERROR` is the env variable.
It won't override when customer command throws any error.

This will solve https://github.com/percy/cli/issues/1590 issue.
Ticket -> https://browserstack.atlassian.net/browse/PER-3235

### DIFF
--- a/packages/cli-command/src/command.js
+++ b/packages/cli-command/src/command.js
@@ -7,6 +7,8 @@ import * as builtInFlags from './flags.js';
 import formatHelp from './help.js';
 import parse from './parse.js';
 
+const percyExitWithZeroOnError = process.env.PERCY_EXIT_WITH_ZERO_ON_ERROR === 'true';
+
 // Copies a command definition and adds built-in flags and config options.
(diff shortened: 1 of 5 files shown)


packages/cli-command/src/command.js
```diff
@@ -155,7 +159,10 @@ export function command(name, definition, callback) {
         err.message ||= `EEXIT: ${err.exitCode}`;
 
         if (definition.exitOnError) {
-          process.exit(err.exitCode);
+          let shouldOverrideExitCode = err.shouldOverrideExitCode !== false;
+          let percyExitWithZeroOnError = process.env.PERCY_EXIT_WITH_ZERO_ON_ERROR === 'true';
```
pankaj443, line 163: possible to remove duplication?

pankaj443 (approved):

PR #1656, opened 2024-07-10
Auto capture and insert cookie in asset discovery browser

chinmay-browserstack:
- Capture cookies using `document.cookie` in serialize-dom so can be used in asset discovery.
- Use this cookie and insert in asset discovery browser.
- Disable this behaviour using PERCY_DO_NOT_USE_CAPTURED_COOKIES env variable
- Prioritise cookie passed via percy.yml over auto captured cookie.

Note: httponly cookie can't be captured so that needs to be passed using percy.yml

### DIFF
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -157,9 +157,9 @@ async function* captureSnapshotResources(page, snapshot, options) {
   const log = logger('core:discovery');
   let { discovery, additionalSnapshots = [], ...baseSnapshot } = snapshot;
   let { capture, captureWidths, deviceScaleFactor, mobile, captureForDevices } = options;
-  let cookies = snapshot?.domSnapshot?.cookies;
-  if (process.env.PERCY_DO_NOT_USE_CAPTURED_COOKIES === 'true') {
-    cookies = null;
+  let cookies;
+  if (process.env.PERCY_DO_NOT_USE_CAPTURED_COOKIES !== 'true') {
+    cookies = snapshot?.domSnapshot?.cookies;
   }
   if (typeof cookies === 'string') {
     cookies = cookies.split('; ').map(c => c.split('='));
@@ -185,7 +185,7 @@ async function* captureSnapshotResources(page, snapshot, options) {
 
   // navigate to the url
   yield resizePage(snapshot.widths[0]);
-  yield page.goto(snapshot.url, cookies);
+  yield page.goto(snapshot.url, { cookies });
 
   if (snapshot.execute) {
     // when any execute options are provided, inject snapshot options
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -158,7 +158,7 @@ async function* captureSnapshotResources(page, snapshot, options) {
   let { discovery, additionalSnapshots = [], ...baseSnapshot } = snapshot;
   let { capture, captureWidths, deviceScaleFactor, mobile, captureForDevices } = options;
   let cookies = snapshot?.domSnapshot?.cookies;
-  if (!process.env.PERCY_DO_NOT_USE_CAPTURED_COOKIES || process.env.PERCY_DO_NOT_USE_CAPTURED_COOKIES !== 'true') {
+  if (process.env.PERCY_DO_NOT_USE_CAPTURED_COOKIES || process.env.PERCY_DO_NOT_USE_CAPTURED_COOKIES === 'true') {
     cookies = null;
   }
   if (typeof cookies === 'string') {
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -161,7 +161,7 @@ async function* captureSnapshotResources(page, snapshot, options) {
   if (process.env.PERCY_DO_NOT_USE_CAPTURED_COOKIES !== 'true') {
     cookies = snapshot?.domSnapshot?.cookies;
   }
-  if (typeof cookies === 'string') {
+  if (typeof cookies === 'string' && cookies !== '') {
     cookies = cookies.split('; ').map(c => c.split('='));
     cookies = cookies.map(([key, value]) => { return { name: key, value: value }; });
   }
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -157,6 +157,14 @@ async function* captureSnapshotResources(page, snapshot, options) {
   const log = logger('core:discovery');
   let { discovery, additionalSnapshots = [], ...baseSnapshot } = snapshot;
   let { capture, captureWidths, deviceScaleFactor, mobile, captureForDevices } = options;
+  let cookies = snapshot?.domSnapshot?.cookies;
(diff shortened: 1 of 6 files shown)


packages/core/src/discovery.js
```diff
@@ -157,6 +157,14 @@ async function* captureSnapshotResources(page, snapshot, options) {
   const log = logger('core:discovery');
   let { discovery, additionalSnapshots = [], ...baseSnapshot } = snapshot;
   let { capture, captureWidths, deviceScaleFactor, mobile, captureForDevices } = options;
+  let cookies = snapshot?.domSnapshot?.cookies;
+  if (process.env.PERCY_DO_NOT_USE_CAPTURED_COOKIES || process.env.PERCY_DO_NOT_USE_CAPTURED_COOKIES === 'true') {
```
prklm10, line 161:
```suggestion
  if (process.env.PERCY_DO_NOT_USE_CAPTURED_COOKIES === 'true') {
```
This is sufficient.

packages/core/src/discovery.js
```diff
@@ -157,6 +157,14 @@ async function* captureSnapshotResources(page, snapshot, options) {
   const log = logger('core:discovery');
   let { discovery, additionalSnapshots = [], ...baseSnapshot } = snapshot;
   let { capture, captureWidths, deviceScaleFactor, mobile, captureForDevices } = options;
+  let cookies = snapshot?.domSnapshot?.cookies;
```
rishigupta1599, line 160:
```suggestion
  let cookies;
  if (process.env.PERCY_DO_NOT_USE_CAPTURED_COOKIES !== 'true') {
    cookies = snapshot?.domSnapshot?.cookies;
  }
```

packages/core/src/discovery.js
```diff
@@ -177,7 +185,7 @@ async function* captureSnapshotResources(page, snapshot, options) {
 
   // navigate to the url
   yield resizePage(snapshot.widths[0]);
-  yield page.goto(snapshot.url);
+  yield page.goto(snapshot.url, cookies);
```
rishigupta1599, line 188: Why passing as separate argument, can't we pass as an option?

packages/core/src/page.js
```diff
@@ -49,17 +49,18 @@ export class Page {
   }
 
   // Go to a URL and wait for navigation to occur
-  async goto(url, { waitUntil = 'load' } = {}) {
+  async goto(url, cookies, { waitUntil = 'load' } = {}) {
```
rishigupta1599, line 52: Can't we merge in options as waitUntil?

packages/core/src/page.js
```diff
@@ -49,17 +49,18 @@ export class Page {
...
     this.log.debug(`Navigate to: ${url}`, this.meta);
 
     let navigate = async () => {
       // set cookies before navigation so we can default the domain to this hostname
-      if (this.session.browser.cookies.length) {
+      if (this.session.browser.cookies.length || cookies) {
         let defaultDomain = hostname(url);
 
+        cookies = this.session.browser.cookies.length > 0 ? this.session.browser.cookies : cookies;
```
rishigupta1599, line 60: Here we are either taking this.session.browser.cookies.length or auto capture cookies, ideally we should merge both cookies map and set.

packages/core/src/discovery.js
rishigupta1599: This will break lot of flows where we use page.goto as it will now require cookies to be passed explicitly.

packages/core/src/page.js
```diff
@@ -49,17 +49,39 @@ export class Page {
   }
 
   // Go to a URL and wait for navigation to occur
-  async goto(url, { waitUntil = 'load' } = {}) {
+  async goto(url, { waitUntil = 'load', cookies } = {}) {
     this.log.debug(`Navigate to: ${url}`, this.meta);
 
     let navigate = async () => {
+      const mergeCookies = (userPassedCookie, autoCapturedCookie) => {
```
rishigupta1599, line 56: Instead of keeping arrow function inside goto itslef let's create separate function outside this method as it makes this code clumsy.

packages/core/src/page.js
```diff
@@ -49,17 +49,39 @@ export class Page {
...
+        return uniqueCookies;
+      };
+
+      const userPassedCookie = this.session.browser.cookies;
       // set cookies before navigation so we can default the domain to this hostname
-      if (this.session.browser.cookies.length) {
+      if (userPassedCookie.length || cookies) {
         let defaultDomain = hostname(url);
 
+        if (userPassedCookie.length > 0 && cookies) {
```
rishigupta1599, line 76: Pass both userPassedCookie and cookies in the method suggested above and handle all conditions there, so all cookie related stuff will be in that method and easy to debug in future.

rishigupta1599 (approved):
LGTM

packages/core/src/page.js
```diff
@@ -48,18 +48,38 @@ export class Page {
...
+    const mergedCookies = [...userPassedCookie, ...autoCapturedCookie];
+    const uniqueCookies = [];
+    const names = new Set();
+
+    for (const cookie of mergedCookies) {
+      if (!names.has(cookie.name)) {
+        uniqueCookies.push(cookie);
+        names.add(cookie.name);
+      }
+    }
```
prklm10, line 64: write a comment that user passed cookie will override similar auto captured cookies.

chinmay-browserstack: Done

prklm10 (approved):

davidrunger:
@chinmay-browserstack @prklm10 I _think_ that this PR is causing me to see the following warning in my logs:

```
[percy] Invalid snapshot options:
```

There is no additional logging about what the specific invalid snapshot options are; only that log line is printed.

Example here: https://github.com/davidrunger/david_runger/actions/runs/10013824950/job/27682261695#step:10:231

Part of the reason that I believe that this PR is the cause of that warning being logged is because if I open this file -- `node_modules/.pnpm/node_modules/@percy/config/dist/validate.js` -- and if I add the following to the top of the `validate` function that is defined therein on L186, then I stop seeing the log line about `[percy] Invalid snapshot options:` when I run my test:

```js
delete data.domSnapshot.cookies;
```

I tried this because when (before making the just-mentioned addition) I logged `ajv.errors`, part of the output included this:

```js
    // ...
    keyword: 'unevaluatedProperties',
    params: { unevaluatedProperty: 'cookies' },
    message: 'must NOT have unevaluated properties',
    // ...
```

davidrunger:
Also, downgrading from `@percy/cli` `1.29.0` to `1.28.9` fixes the issue (i.e. causes `[percy] Invalid snapshot options:` not to be logged anymore).

chinmay-browserstack:
@davidrunger Thanks for pointing this out. i have checked and found cookies field is missing from config as you mentioned.

Created a fix PR for it https://github.com/percy/cli/pull/1669. it will be fixed in next release.
Meanwhile you can continue using @percy/cli 1.29.0 as it just prints warning and has no impact on any functionality of percy cli.

PR #1658, opened 2024-07-14
Add log for verfying if capture responsive assets is on

chinmay-browserstack:
- Added log when capture responsive assets are on
- Move capture responsive assets logic after capturing dom snapshot (serialize-dom) earlier it is capturing dom in mobile widths when mobile browsers are enabled

### DIFF
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -158,11 +158,28 @@ async function* captureSnapshotResources(page, snapshot, options) {
   let { discovery, additionalSnapshots = [], ...baseSnapshot } = snapshot;
   let { capture, captureWidths, deviceScaleFactor, mobile, captureForDevices, captureResponsiveAssetsEnabled } = options;
 
+  // iterate over device to trigger reqeusts and capture other dpr width
+  async function* captureResponsiveAssets() {
+    if (!captureForDevices) return;
+
+    for (const device of captureForDevices) {
+      // We are not adding these widths and pixels ratios in loop below because we want to explicitly reload the page after resize which we dont do below
+      yield* captureSnapshotResources(page, { ...snapshot, widths: [device.width] }, {
+        captureResponsiveAssetsEnabled,
+        deviceScaleFactor: device.deviceScaleFactor,
+        mobile: true
+      });
+      yield waitForDiscoveryNetworkIdle(page, discovery, captureResponsiveAssetsEnabled);
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


packages/core/src/discovery.js
```diff
@@ -76,11 +76,11 @@ function debugSnapshotOptions(snapshot) {
 }
 
 // Wait for a page's asset discovery network to idle
-function waitForDiscoveryNetworkIdle(page, options) {
+function waitForDiscoveryNetworkIdle(page, options, captureResponsiveAssetsEnabled) {
```
ninadbstack, line 79: isnt that already part of `options` ?  if not as discovery is passed you can just add it under discovery options. - it should already be there regardless

packages/core/src/page.js
```diff
@@ -174,7 +175,7 @@ export class Page {
     }
 
     // wait for any final network activity before capturing the dom snapshot
-    await this.network.idle();
+    await this.network.idle(undefined, undefined, captureResponsiveAssetsEnabled);
```
ninadbstack, line 178: wait why are we doing this ? its basically not doing anything at all and not waiting for page idle

packages/core/src/network.js
```diff
@@ -101,7 +101,7 @@ export class Network {
     }).catch(error => {
       if (error.message.startsWith('Timeout')) {
         this._throwTimeoutError((
-          'Timed out waiting for network requests to idle.'
+          `Timed out waiting for network requests to idle.\nCapturing responsive asset: ${captureResponsiveAssetsEnabled}`
```
ninadbstack, line 104:
```suggestion
          `Timed out waiting for network requests to idle.\nCapturing responsive assets: ${captureResponsiveAssetsEnabled}`
```

packages/core/src/discovery.js
```diff
@@ -346,7 +351,8 @@ export function createDiscoveryQueue(percy) {
             yield* captureSnapshotResources(page, snapshot, {
               captureWidths: !snapshot.domSnapshot && percy.deferUploads,
               capture: callback,
-              captureForDevices: percy.deviceDetails || []
+              captureForDevices: percy.deviceDetails || [],
+              captureResponsiveAssetsEnabled: !!percy.deviceDetails
```
ninadbstack, line 355: Not sure why this is required. On line 166 wouldnt it be always true ? if array is empty it will never go in loop

ninadbstack (approved):

PR #1663, opened 2024-07-18
Adding logs for debugging missing-resources

pankaj443:


packages/client/src/client.js
```diff
@@ -334,7 +334,7 @@ export class PercyClient {
     }
     let encodedContent = base64encode(content);
 
-    this.log.debug(`Uploading ${formatBytes(encodedContent.length)} resource: ${url}...`);
+    this.log.debug(`Uploading ${formatBytes(encodedContent.length)} resource for snapshotId: ${snapshotId}: ${url}...`);
```
ninadbstack, line 337:
```suggestion
    this.log.debug(`Uploading ${formatBytes(encodedContent.length)} resource for snapshotId ${snapshotId} : ${url}...`);
```
avoiding double `:`

ninadbstack: Better option is to put build id and snapshot id in meta though. That way it will be automatically extracted and downloaded in snapshot logs

ninadbstack: Ideally update all logs in client to use meta which contains useful info like { build id, snapshot id, comparison id, url if applicable etc }

PR #1664, opened 2024-07-18
feat: added invalid URL detection for snapshot & network urls

this-is-shivamsingh:


packages/core/src/network.js
```diff
@@ -194,6 +194,21 @@ export class Network {
     // do not handle data urls
     if (request.url.startsWith('data:')) return;
 
+    // Network requests are handled by the browser, which is lenient about URL encoding.
+    // Invalid characters in the URL won't cause snapshot/build failures.
+    // Therefore, this code only checks for invalid characters in the network request URL.
+    try {
```
rishigupta1599, line 200: I thought we were doing only for root URL and not other, here it will encode all the URLs

packages/core/src/snapshot.js
```diff
@@ -86,8 +88,22 @@ function mapSnapshotOptions(snapshots, context) {
     // transform snapshot URL shorthand into an object
     if (typeof snapshot === 'string') snapshot = { url: snapshot };
 
-    // normalize the snapshot url and use it for the default name
+    try {
+      // encoding snapshot url, if contians invalid URI characters/syntax
+      let modifiedURL = modifyURL(snapshot.url);
```
rishigupta1599, line 93: Above code and this code is exactly same except the URL. Don't duplicate the code

packages/core/src/utils.js
```diff
@@ -373,6 +373,22 @@ export function base64encode(content) {
...
+// Also, if a character is a part of valid URI syntax, those characters
+// are not encoded
+// Eg: [abc] -> gets encoded to %5Babc%5D
+// ab c -> ab%20c
+export function modifyURL(url) {
+  // In case the url is partially encoded, then directly using encodeURI()
+  // will encode those characters again. Therefore decodeURI once helps is decoding
+  // partially encoded URL and then after encoding it again, full URL get encoded
+  // correctly.
+  let decodeURL = decodeURI(url); // This can throw error, so handle it will trycatch
```
rishigupta1599, line 387: Handle try catch here only and return unmodified URL in case of any error, why parent function should handle that

packages/core/src/network.js
```diff
@@ -1,7 +1,7 @@
 import { request as makeRequest } from '@percy/client/utils';
 import logger from '@percy/logger';
 import mime from 'mime-types';
-import { DefaultMap, createResource, hostnameMatches, normalizeURL, waitFor } from './utils.js';
+import { DefaultMap, createResource, hostnameMatches, normalizeURL, waitFor, modifyURL } from './utils.js';
```
ninadbstack, line 4:
```suggestion
import { DefaultMap, createResource, hostnameMatches, normalizeURL, waitFor, decodeEncodeUrl } from './utils.js';
```

packages/core/src/snapshot.js
```diff
@@ -86,8 +88,22 @@ function mapSnapshotOptions(snapshots, context) {
...
+      if (modifiedURL !== snapshot.url) {
+        log.info(`Snapshot URL modified to: ${modifiedURL}`);
```
ninadbstack, line 95:
```suggestion
        log.debug(`Snapshot URL modified to: ${modifiedURL}`);
```

packages/core/src/snapshot.js
```diff
@@ -86,8 +88,22 @@ function mapSnapshotOptions(snapshots, context) {
...
+        snapshot.url = modifiedURL;
+      }
+    } catch (error) {
+      log.debug(error);
+      if (error.name === 'URIError') {
+        log.warn(`Invalid URL detected for url: ${snapshot.url}`);
```
ninadbstack, line 101:
```suggestion
        log.warn(`Invalid URL detected for url: ${snapshot.url} - the snapshot may fail on percy. Please confirm that your website URL valid. `);
```

packages/core/src/snapshot.js
```diff
@@ -86,8 +88,22 @@ function mapSnapshotOptions(snapshots, context) {
     // transform snapshot URL shorthand into an object
     if (typeof snapshot === 'string') snapshot = { url: snapshot };
 
-    // normalize the snapshot url and use it for the default name
+    try {
```
ninadbstack, line 91: Move to `validateAndFixSnapshotUrl`

packages/core/src/network.js
```diff
@@ -194,6 +194,21 @@ export class Network {
     // do not handle data urls
     if (request.url.startsWith('data:')) return;
 
+    // Network requests are handled by the browser, which is lenient about URL encoding.
+    // Invalid characters in the URL won't cause snapshot/build failures.
+    // Therefore, this code only checks for invalid characters in the network request URL.
+    try {
+      let modifiedURL = modifyURL(request.url);
+      if (modifiedURL !== request.url) {
+        this.log.debug(`Inconsistent network URL detected for url: ${request.url}.`);
```
ninadbstack, line 203: Check if this will fail on jackproxy - if not then dont show error - if yes then show. 

And also move this to utils and pass logger

packages/core/src/network.js
this-is-shivamsingh: - We are only modifying for root URLs, but only detects for network URLs as well.

packages/core/src/snapshot.js
this-is-shivamsingh: - The reason to handle try catch separately, is because if we log the error in utils.js, we will not know from where we are logging this error either from snapshot or network etc. So Because of that parent function needs to handle it.

However, to avoid here, I am also passing `logger` now, from the parent function to log the error. This will ensure, from where it is getting logged.

packages/core/src/utils.js
this-is-shivamsingh: ^same as above.

packages/core/src/network.js
this-is-shivamsingh: Addressed. Named it `decodeAndEncodeURLWithLogging`

packages/core/src/snapshot.js
this-is-shivamsingh: Added metadata for this as well.

packages/core/src/snapshot.js
this-is-shivamsingh: Addressed.

packages/core/src/network.js
this-is-shivamsingh: - Tried with different invalid characters eg. `?? @ !`. But for some of the special characters `%`, `%11 `, or leading spaces at the end of URL. ie. `http//localhost:8000/test% .png`, for these cases it's passed in the browser but our jackproxy breaks. These cases can be detected by `decodeAndEncodeURLWithLogging` functions.

this-is-shivamsingh: Here, I test with this url:` http://localhost:3000/avif_edge_support/testfile%_5232.png`, that load fine in browser, but causes error on render side. `Proxy fails to start!`.

packages/core/src/snapshot.js
this-is-shivamsingh: Addressed.

ninadbstack (approved):

packages/core/src/network.js
```diff
@@ -194,6 +194,14 @@ export class Network {
     // do not handle data urls
     if (request.url.startsWith('data:')) return;
 
+    // Browsers handle URL encoding leniently, but invalid characters can break tools like Jackproxy.
+    // This code checks for issues such as `%` and leading spaces and warns the user accordingly.
+    decodeAndEncodeURLWithLogging(request.url, this.log, {
+      meta: { ...this.meta, url: request.url },
+      shouldLogWarning: true,
+      warningMessage: `An invalid URL was detected for url: ${request.url} - the snapshot may fail on Percy. Please verify that your asset URL is valid.`
```
rishigupta1599, line 202: Is this warning required if we have fixed this?

packages/core/src/snapshot.js
```diff
@@ -24,6 +25,21 @@ function validURL(url, base) {
   }
 }
 
+function validateAndFixSnapshotUrl(snapshot) {
+  let log = logger('core:snapshot');
+  // encoding snapshot url, if contians invalid URI characters/syntax
+  let modifiedURL = decodeAndEncodeURLWithLogging(snapshot.url, log, {
+    meta: { snapshot: { name: snapshot.name || snapshot.url } },
+    shouldLogWarning: true,
+    warningMessage: `Invalid URL detected for url: ${snapshot.url} - the snapshot may fail on Percy. Please confirm that your website URL is valid.`
```
rishigupta1599, line 34: Same question as as previous usage

packages/core/src/snapshot.js
```diff
@@ -86,8 +102,10 @@ function mapSnapshotOptions(snapshots, context) {
     // transform snapshot URL shorthand into an object
     if (typeof snapshot === 'string') snapshot = { url: snapshot };
 
-    // normalize the snapshot url and use it for the default name
+    validateAndFixSnapshotUrl(snapshot);
```
rishigupta1599, line 105: Why not calling this method in `validURL` method, single place for all URL validation

packages/core/src/network.js
```diff
@@ -194,6 +194,14 @@ export class Network {
     // do not handle data urls
     if (request.url.startsWith('data:')) return;
 
+    // Browsers handle URL encoding leniently, but invalid characters can break tools like Jackproxy.
+    // This code checks for issues such as `%` and leading spaces and warns the user accordingly.
+    decodeAndEncodeURLWithLogging(request.url, this.log, {
+      meta: { ...this.meta, url: request.url },
+      shouldLogWarning: true,
```
rishigupta1599, line 201: `shouldLogWarning: true`, this is always true so do we even need this as an argument?

packages/core/src/snapshot.js
this-is-shivamsingh: - valid URL is used to create URL obj, and has been used in other places as well in snapshot.js, like getting allowedHostname, so do not want to modify any such URL at any place apart from `snapshot.url`, so created a separate function for that.

packages/core/src/network.js
this-is-shivamsingh: - `shouldLogWarning: true`, in our case Yes, but this function can be used in other cases as well in future to only get modified url and don't want to show the warning, so made it a little generic to use.

packages/core/src/snapshot.js
rishigupta1599: All the URL validation should have encoding ideally, and should return URL object with proper encoding applied, if I need to use validURL and want encoding as well I will need to call two methods. Either take it as argument if encoding is required or not so that in future I can just pass encode option in this method and it will return correct URL

this-is-shivamsingh: - Will be testing `npx percy snapshot command` + defer uploads as well.

PR #1669, opened 2024-07-22
Add cookies in config for domSnapshot

chinmay-browserstack:
Adding `cookies` field in config of domSnapshot

### DIFF
--- a/packages/core/test/unit/config.test.js
+++ b/packages/core/test/unit/config.test.js
@@ -85,6 +85,21 @@ describe('SnapshotSchema', () => {
     expect(snapshotSchema.$defs.common.properties).toEqual(jasmine.objectContaining({ domTransformation: jasmine.anything() }));
   });
 
+  it('should contain valid dom serialization values', () => {
+    const dom = {
+      url: 'some_url',
+      width: 1000,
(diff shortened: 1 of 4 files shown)


rishigupta1599 (approved):

rishigupta1599 (changes requested):
Add testcase with config validation for DOM serialization response so that we don't miss it in future config changes as well

packages/core/test/unit/config.test.js
```diff
@@ -85,6 +85,21 @@ describe('SnapshotSchema', () => {
     expect(snapshotSchema.$defs.common.properties).toEqual(jasmine.objectContaining({ domTransformation: jasmine.anything() }));
   });
 
+  it('should contain valid dom serialization values', () => {
+    const dom = {
```
rishigupta1599, line 89: This you have hardcoded so it will always pass until you remove it from config.
Expectation for test case should be PercyDOM.serialize -> response -> PercyConfig.validate(response) -> No errors

rishigupta1599 (approved):

pankaj443 (approved):
LGTM

Issue #1674, opened 2024-07-22
Percy reports "Invalid snapshot options:" even when there is zero problems


navrkald:
## The problem

Percy reports "Invalid snapshot options:" even when there is zero problems.

## Environment

- Node version: 20
- `@percy/cli` version: 1.29.0
- Version of Percy SDK you’re using:
- If needed, a build or snapshot ID:
- OS version: Mac
- Type of shell command-line [interface]:

## Debug logs

tsc && PERCY_BRANCH=staging npx percy exec -c censored.js -- node censored.js
[percy] Notice: Percy collects CI logs for service improvement, stored for 30 days. Opt-out anytime with export PERCY_CLIENT_ERROR_LOGS=false
[percy] Percy has started!
[percy] Running "node censored.js"
[percy] Invalid snapshot options:
[percy] Snapshot taken: censored
[percy] Finalized build censored
[percy] Build's CLI and CI logs sent successfully. Please share this log ID with Percy team in case of any issues - 35435194_build_35435194_cli_5bdfdd0532339ace71a37854a9c60702339fe9f6
[percy] Command "node censored.js" exited with status: 0

davidrunger:
@navrkald : I believe that you are seeing this because of what I mentioned here: https://github.com/percy/cli/pull/1656#issuecomment-2240277498 . A fix to address the reason that error started emerging is here: https://github.com/percy/cli/pull/1669 .

That being said, from the PR that you have put together (https://github.com/percy/cli/pull/1675), I notice that another issue related to this has come to your attention, which is that `[percy] Invalid snapshot options:` will be logged even if `errors` is an empty array, since an empty array is truthy in JavaScript. Prior to the issue that I mentioned above, `errors` was `undefined`, which is why nothing was logged, but now `errors` is an empty array, which is why `[percy] Invalid snapshot options:` is logged, but then no further details about what is invalid are logged.

[...]

PR #1675, opened 2024-07-22
Fix of issue id #1674, reporting false positive invalid config warning.

navrkald:
Validate presence of errors in config validator.

Fix: https://github.com/percy/cli/issues/1674


sarah-gelt:
This false warning error is really annoying and easily fixed. Could it get a review please? 🙏

chinmay-browserstack (approved):

PR #1677, opened 2024-07-25
adding request url to proxying request error log

prklm10:
adding request url to proxying request error log

### DIFF
--- a/packages/client/src/proxy.js
+++ b/packages/client/src/proxy.js
@@ -168,7 +168,7 @@ export class ProxyHttpsAgent extends https.Agent {
 
     let handleError = err => {
       socket.destroy(err);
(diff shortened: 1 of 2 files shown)


packages/client/src/proxy.js
```diff
@@ -168,7 +168,7 @@ export class ProxyHttpsAgent extends https.Agent {
 
     let handleError = err => {
       socket.destroy(err);
-      logger('client:proxy').error(`Proxying request failed: ${err}`);
+      logger('client:proxy').error(`Proxying request ${href(options)} failed: ${err}`);
```
prklm10, line 171: don't think this is needed as line 153 logs the proxying request.
Added because it would be better to understand error logs.

ninadbstack (approved):

PR #1686, opened 2024-08-06
PPLT_3233: Added CLI Start Time

this-is-shivamsingh:
Jira: https://browserstack.atlassian.net/browse/PPLT-3233

Changes:
- Added cliStartTime when percy server get started first time.

### DIFF
--- a/packages/cli-build/src/finalize.js
+++ b/packages/cli-build/src/finalize.js
@@ -19,7 +19,7 @@ export const finalize = command('finalize', {
   log.info('Finalizing parallel build...');
 
   // rely on the parallel nonce to cause the API to return the current running build for the nonce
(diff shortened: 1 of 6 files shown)


Amit3200 (approved):

chinmay-browserstack (approved):

pankaj443 (approved):
LGTM

prklm10 (approved):

this-is-shivamsingh:
- Small Change
- Instead of sending Date.now() ( ie. an epoch time), we are now sending the exact UTC date, because epoch time in JS is in millisecond while percy-api consider it in seconds. Therefore there was a date mismatch happening. So we are now sending the exact date only

prklm10 (approved):

PR #1711, opened 2024-09-05
Feat(DOM-Snapshot) Serialization support for External StyleSheets (configurable)

sharad-sharma:


ninadbstack:
@sharad-sharma hey, so I see that you still need to update this and its not ready. But just need to know why do you need to serialize external stylesheets ? 

We already handle external stylesheets as part of resource discovery step further in the core package. Which captures all external calls - css or any other. Are you facing any issue with that ?

sharad-sharma:
> @sharad-sharma hey, so I see that you still need to update this and its not ready. But just need to know why do you need to serialize external stylesheets ?
> 
> We already handle external stylesheets as part of resource discovery step further in the core package. Which captures all external calls - css or any other. Are you facing any issue with that ?

hey @ninadbstack, we wanted to use `@percy/dom` as a standalone library without the core package as we need to capture the snapshot via browser extension.

I see this test `does not override previous guids when reserializing`, is it a requirement for `percy-cli`to mark the original dom with guids?

ninadbstack:
@sharad-sharma I see, but then only external stylesheets wont help you here. There are many resources that needs to be captured and serialized. 

That said we dont recommend doing this in dom package [ at least as of now ] as any code added here runs in sync in user test code blocking user test execution. + More items we capture more we need to return via selenium/any other framework on the network -  which makes calls further slower.

As for guid marking its info that we are planning on using for some future features - allowing us to know that snapshots were taken on common page and we could use it to optimize discovery. But if you dont want it - you can add an optional flag to disable it.

PR #1715, opened 2024-09-10
Adding waitForTimeout/waitForSelector for assets discovery browser when JS is enabled 

amandeepsingh333:
- added waitForTimeout and waitForSelector at the time of discovery when JS is enabled 

**so now waitForTimeout and waitForSelector are at 2 places:**
- snapshot level (snapshot.waitForTimeout)
- discovery level (snapshot.discovery.waitForTimeout)

**usage of both:**

 **when it is passed at snapshot level** => 
these options are used before a DOM snapshot is taken. The primary goal is to ensure that the page has fully loaded and any necessary DOM elements are present before capturing the state of the page. This mode is currently only supported when used with percy snapshot cli command.

 **when it is passed at discovery level and JS is enabled** => in this phase, these options are used to wait for specific assets or network requests to be discovered and captured. The asset-discovery process often occurs during or after initial page load, and some resources may not be immediately available. 

**how client should use:**

snapshot level => You should use these options at the snapshot level when you want to ensure the readiness of a page before it is captured. This ensures that no partial or incomplete content is included in the snapshot.

discovery level =>  You should use these options in the discovery level when you are facing issues with missing assets or incomplete network requests. By waiting for the selector or a set timeout, you give enough time for assets to load

- used both at the time of discovery
- added test cases for the same

### DIFF
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -1,5 +1,6 @@
 import logger from '@percy/logger';
 import Queue from './queue.js';
+import Page from './page.js';
 import {
   normalizeURL,
   hostnameMatches,
@@ -9,7 +10,9 @@ import {
   createLogResource,
   yieldAll,
   snapshotLogName,
-  withRetries
+  waitForTimeout,
+  withRetries,
+  waitForSelectorInsideBrowser
 } from './utils.js';
 import {
   sha256hash
@@ -202,6 +205,18 @@ async function* captureSnapshotResources(page, snapshot, options) {
   // navigate to the url
   yield resizePage(snapshot.widths[0]);
   yield page.goto(snapshot.url, { cookies });
+
+  // wait for any specified timeout
+  if (snapshot.discovery.waitForTimeout && page.enableJavaScript) {
+    log.debug(`Wait for ${snapshot.discovery.waitForTimeout}ms timeout`);
+    await waitForTimeout(snapshot.discovery.waitForTimeout);
+  }
+
+  // wait for any specified selector
+  if (snapshot.discovery.waitForSelector && page.enableJavaScript) {
+    log.debug(`Wait for selector: ${snapshot.discovery.waitForSelector}`);
+    await waitForSelectorInsideBrowser(page, snapshot.discovery.waitForSelector, Page.TIMEOUT);
+  }
 
   if (snapshot.execute) {
     // when any execute options are provided, inject snapshot options
--- a/packages/core/test/discovery.test.js
+++ b/packages/core/test/discovery.test.js
@@ -2430,6 +2430,77 @@ describe('Discovery', () => {
     });
   });
 
+  describe('waitForSelector/waitForTimeout at the time of discovery when Js is enabled =>', () => {
+    it('calls waitForTimeout, waitForSelector and page.eval when their respective arguments are given', async () => {
+      const page = await percy.browser.page();
+      spyOn(percy.browser, 'page').and.returnValue(page);
+      spyOn(page, 'eval').and.callThrough();
+      percy.loglevel('debug');
+
+      await percy.snapshot({
+        name: 'test discovery',
+        url: 'http://localhost:8000',
+        enableJavaScript: true,
+        discovery: {
+          waitForTimeout: 100,
+          waitForSelector: 'body'
+        }
+      });
+      await percy.idle();
+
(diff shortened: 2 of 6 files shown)


packages/core/src/discovery.js
```diff
@@ -203,6 +204,20 @@ async function* captureSnapshotResources(page, snapshot, options) {
   yield resizePage(snapshot.widths[0]);
   yield page.goto(snapshot.url, { cookies });
 
+  // wait for any specified timeout
+  if (snapshot.discovery.waitForTimeout && snapshot.enableJavaScript) {
```
this-is-shivamsingh, line 208: we are checking for js enabled assets discovery browser, check for `page.enableJavascript` that will have condition checked if the assets discovery browser is opened in js enabled or not.

this-is-shivamsingh:
1. Please Update the PR description and PR title as well

packages/core/src/discovery.js
```diff
@@ -9,6 +9,7 @@ import {
   createLogResource,
   yieldAll,
   snapshotLogName,
+  waitForTimeout as sleep,
```
this-is-shivamsingh, line 12: is waitForTimeout name already taken ?

packages/core/src/discovery.js
```diff
@@ -203,6 +204,20 @@ async function* captureSnapshotResources(page, snapshot, options) {
   yield resizePage(snapshot.widths[0]);
   yield page.goto(snapshot.url, { cookies });
 
+  // wait for any specified timeout
+  if (snapshot.discovery.waitForTimeout && page.enableJavaScript) {
+    log.debug(`Wait for ${snapshot.discovery.waitForTimeout}ms timeout`);
+    await sleep(snapshot.discovery.waitForTimeout);
+  }
+
+  let timeout = parseInt(process.env.PERCY_PAGE_LOAD_TIMEOUT) || 30000;
```
this-is-shivamsingh, line 213: use Page.TIMEOUT instead

packages/core/test/discovery.test.js
```diff
@@ -2430,6 +2430,29 @@ describe('Discovery', () => {
     });
   });
 
+  describe('discovery parameters =>', () => {
+    it('checks for waitForTimeout and waitForSelector at discovery', async () => {
```
this-is-shivamsingh, line 2434: `describe('waitForSelector/waitForTimeout')`

   describe('when Js is enabled')
     it('waits for Timeout')
     it('waits for selector')
 
  describe('when JS is disabled')
    it('doesn't waits for Timeout')

- Add this test case as well
- Also spyOn(page), and validate that page.eval was called when JS in enabled

packages/core/src/discovery.js
```diff
@@ -203,6 +205,18 @@ async function* captureSnapshotResources(page, snapshot, options) {
...
+  // wait for any specified timeout
+  if (snapshot.waitForTimeout && page.enableJavaScript) {
+    log.debug(`Wait for ${snapshot.waitForTimeout}ms timeout`);
+    await waitForTimeout(snapshot.waitForTimeout);
+  }
+
+  // wait for any specified selector
+  if (snapshot.waitForSelector && page.enableJavaScript) {
+    log.debug(`Wait for selector: ${snapshot.waitForSelector}`);
+    await page.eval(`await waitForSelector(${JSON.stringify(snapshot.waitForSelector)}, ${Page.TIMEOUT})`);
```
ninadbstack, line 217: Create a common function like waitForTimeout that you have included above

ninadbstack (approved):

packages/core/test/discovery.test.js
```diff
@@ -2430,6 +2431,93 @@ describe('Discovery', () => {
...
+      });
+      await percy.idle();
+
+      expect(page.eval).toHaveBeenCalledWith('await waitForSelector("body", 30000)');
+      expect(logger.stderr).toEqual(jasmine.arrayContaining([
+        '[percy:core:discovery] Wait for selector: body',
+        '[percy:core:discovery] Wait for 100ms timeout'
+      ]));
+    });
+    it('waitForTimeout and waitForSelector are not called Js is disabled and domSnapshot is not present', async () => {
```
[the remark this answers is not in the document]
this-is-shivamsingh: Rest everything looks fine to me

this-is-shivamsingh (approved):
Lgtm 🌮 , Just update the tests please.

PR #1718, opened 2024-09-11
Removed snapshopt logs from being uploaded as a resource

revathySV:
https://browserstack.atlassian.net/browse/PER-3804


prklm10:
Closing this as we need this data for fetching network logs. 
It is going to be reopened once we find good way to parse existing build logs.

PR #1723, opened 2024-09-16
Load blank before url for each discovery

revathySV:


ninadbstack:
Closed - handled in another PR

PR #1727, opened 2024-09-16
Fixed cookies parsing bug in auto captured cookies

ninadbstack:
Context:
- Cookie parsing logic failed when cookies were encoded as base64 and had `=` at the end. 
- When servers used strict base64 cookie check this ended up with error response causing `ERR_HTTP2_PROTOCOL_ERROR` in discovery chromium. 

This PR fixes the logic and refactors the function. 
Also few unrelated logging improvements.

Secure cookie ->
- Sometime we are getting `Protocol error (Network.setCookies): Invalid cookie fields` because the percy captured cookie did not capture the secure flag
- Adding support for secure flag that can be set by setting this env variable `PERCY_SECURE_COOKIES` to true.

### DIFF
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -78,6 +78,20 @@ function debugSnapshotOptions(snapshot) {
   }
 }
 
+// parse browser cookies in correct format if flag is enabled
+// it assumes that cookiesStr is string returned by document.cookie
+function parseCookies(cookiesStr) {
+  if (
+    process.env.PERCY_DO_NOT_USE_CAPTURED_COOKIES === 'true' ||
+    !(typeof cookiesStr === 'string' && cookiesStr !== '')
+  ) return null;
+
+  return cookiesStr.split('; ').map(c => {
+    const eqIdx = c.indexOf('=');
+    return { name: c.substring(0, eqIdx), value: c.substring(eqIdx + 1) };
+  });
+}
+
 // Wait for a page's asset discovery network to idle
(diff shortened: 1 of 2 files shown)


prklm10 (approved):

pankaj443 (approved):

packages/core/src/utils.js
```diff
@@ -461,6 +461,27 @@ export function snapshotLogName(name, meta) {
...
+  ) return null;
+
+  return cookiesStr.split('; ').map(c => {
+    const eqIdx = c.indexOf('=');
+    const cookieObj = {
+      name: c.substring(0, eqIdx),
+      value: c.substring(eqIdx + 1)
+    };
+    if (process.env.PERCY_SECURE_COOKIES === 'true') {
+      cookieObj.secure = true;
```
ninadbstack, line 479: Why is this required? also we cant capture secure cookies, how can we set those here?

prklm10: It is because for one of the customer setting  captured cookie was throwing error. I debugged it and found this. setting the cookie to secure solves the issue.
I have added support for this through env variable so it does not affect other cookie.

ninadbstack: Wouldnt it affect all the cookies ? All will become secure. And secure cookies. Ideally we should set these to same as in test browser. 

Please add a comment on this that new cookie capture process should update this code.

prklm10: reverted my changes. And i will be sending it in separate pr.

PR #1728, opened 2024-09-18
[Defer Uploads v2] Added support for handling multiple root resource in single asset discovery phase

chinmay-browserstack:
- Added new endpoints which will return widths from deviceDetails if present and from config settings.
- Added new config `multiDOM` which when passed we capture DOM in all widths in which it will be rendered.
- Changed structure of `domSnapshot`. now domSnapshot can also be a list of objects with following strucuture `{ 'domSnapshot': <DOM>, 'width': <int>}` 
- Resources map will now contain list of root resources. if multiple root resources are present.
- `getResource` will return root resource based on width on which discovery browser is opened. in case root resource not present for that width we will fallback to first dom in list.
- re load page in case of multiDOM. since root resource will only get updated in we re load the page.
- disable `captureSrcset` if multiDOM is enabled.


packages/core/src/api.js
```diff
@@ -78,6 +78,13 @@ export function createPercyServer(percy, port) {
     .route('get', '/percy/dom.js', (req, res) => {
       return res.file(200, PERCY_DOM);
     })
+    // returns widths used in config and of devices enabled
+    .route('get', '/percy/widths', (req, res) => res.json(200, {
```
ninadbstack, line 82: This is an extra call - all SDKs already make healthcheck call, if we just return it as part of that we can just save it in class variables as this will be only required once and will remain same for all snapshots.

packages/core/src/discovery.js
```diff
@@ -88,21 +88,42 @@ function waitForDiscoveryNetworkIdle(page, options) {
...
 function parseDomResources({ url, domSnapshot }) {
-  if (!domSnapshot) return new Map();
-  let isHTML = typeof domSnapshot === 'string';
-  let { html, resources = [] } = isHTML ? { html: domSnapshot } : domSnapshot;
-  let rootResource = createRootResource(url, html);
+  const map = new Map();
+  if (!domSnapshot) return map;
+  let allResources = new Set();
+
+  if (Array.isArray(domSnapshot)) {
```
ninadbstack, line 95: Any specific reason to bifurcate this ? wouldnt converting a single snapshot entry to an array work here ? Its basically doing same thing per snapshot

packages/core/src/discovery.js
```diff
@@ -88,21 +88,42 @@ function waitForDiscoveryNetworkIdle(page, options) {
...
+    let isHTML = typeof domSnapshot === 'string';
+    let { html, resources = [] } = isHTML ? { html: domSnapshot } : domSnapshot;
+    allResources = resources;
+    let rootResource = createRootResource(url, html);
+    map.set(rootResource.url, rootResource);
+  }
 
   // reduce the array of resources into a keyed map
-  return resources.reduce((map, { url, content, mimetype }) => {
+  return allResources.reduce((map, { url, content, mimetype }) => {
```
ninadbstack, line 118: this is accurate ? can we have same resource url across multiple widths ?

packages/core/src/discovery.js
```diff
@@ -114,11 +135,17 @@ function processSnapshotResources({ domSnapshot, resources, ...snapshot }) {
   let rootContent = domSnapshot?.html ?? domSnapshot;
   let root = resources.find(r => r.content === rootContent);
 
+  let roots = root ? [root] : [];
```
ninadbstack, line 138: same here, this can always be an array if above is handled

packages/core/src/discovery.js
```diff
@@ -129,16 +156,21 @@ function processSnapshotResources({ domSnapshot, resources, ...snapshot }) {
       log.warn('DOM elements found outside </body>, percyCSS might not work');
     }
 
-    let css = createPercyCSSResource(root.url, snapshot.percyCSS);
```
ninadbstack, line 132: Lets refactor this code out, its getting longer.

packages/core/src/discovery.js
```diff
@@ -228,7 +260,7 @@ async function* captureSnapshotResources(page, snapshot, options) {
   // Running before page idle since this will trigger many network calls
   // so need to run as early as possible. plus it is just reading urls from dom srcset
   // which will be already loaded after navigation complete
-  if (discovery.captureSrcset) {
+  if (!snapshot.multiDOM && discovery.captureSrcset) {
```
ninadbstack, line 263: Please update comment above why this is not required for multi dom

packages/core/src/discovery.js
```diff
@@ -247,6 +279,7 @@ async function* captureSnapshotResources(page, snapshot, options) {
         yield page.evaluate(execute?.beforeResize);
         yield waitForDiscoveryNetworkIdle(page, discovery);
         yield resizePage(width = widths[i + 1]);
+        if (snapshot.multiDOM) { yield page.goto(snapshot.url, { cookies }); }
```
ninadbstack, line 282: This creates a problem when url has a `#id` in it. This is an existing problem with mobile: true as well. Please update both codes to first load `about:blank` and then load the actual url.

packages/core/src/page.js
```diff
@@ -18,6 +18,7 @@ export class Page {
     this.session = session;
     this.browser = session.browser;
     this.enableJavaScript = options.enableJavaScript ?? true;
+    this.width = null;
```
ninadbstack, line 21: dont pass this around everywhere - just update it in `intercept` like `currentWidth` and set it when you resize the page.

packages/core/src/discovery.js
chinmay-browserstack: So instead of rootResource if we capture any other resource with a url. we assume it will have same content. 
this is same as current behaviour of deferUploads where we don't override a url if it's already present in resources map. refer: https://github.com/percy/cli/blob/676420ece5f33bfa1e3d8be223abdd8093facb77/packages/core/src/snapshot.js#L303
Even if we start storing resources for different widths our rendering can only pick one resource for a given url.

ninadbstack: What about the custom resources like CSSOM that we gather ? do we uniquely name it ? and would correct width get added to correct root resource ?

packages/core/src/discovery.js
```diff
@@ -88,21 +88,52 @@ function waitForDiscoveryNetworkIdle(page, options) {
...
-  let isHTML = typeof domSnapshot === 'string';
-  let { html, resources = [] } = isHTML ? { html: domSnapshot } : domSnapshot;
-  let rootResource = createRootResource(url, html);
+  const map = new Map();
+  if (!domSnapshot) return map;
+  let allRootResources = new Set();
+  let allResources = new Set();
+
+  if (!Array.isArray(domSnapshot)) {
+    domSnapshot = [{ domSnapshot }];
```
ninadbstack, line 97: should this be an object array with domSnapshot key i it ? that doesnt seem to match your schema

packages/core/src/discovery.js
```diff
@@ -195,16 +223,21 @@ async function* captureSnapshotResources(page, snapshot, options) {
...
+      width
+    });
+  };
 
   // navigate to the url
   yield resizePage(snapshot.widths[0]);
-  yield page.goto(snapshot.url, { cookies });
+  let forceReload = false;
+  if (discovery.captureResponsiveAssetsEnabled) { forceReload = true; }
+  yield page.goto(snapshot.url, { cookies, forceReload });
```
ninadbstack, line 240: Optional
```suggestion
  yield page.goto(snapshot.url, { cookies, forceReload: discovery.captureResponsiveAssetsEnabled });
```

packages/core/src/page.js
```diff
@@ -68,9 +68,14 @@ export class Page {
   }
 
   // Go to a URL and wait for navigation to occur
-  async goto(url, { waitUntil = 'load', cookies } = {}) {
+  async goto(url, { waitUntil = 'load', cookies, forceReload } = {}) {
     this.log.debug(`Navigate to: ${url}`, this.meta);
 
+    if (forceReload) {
+      this.log.debug('Navigating to blank page', this.meta);
+      await this.session.send('Page.navigate', { url: 'about:blank' });
```
ninadbstack, line 76: Do we not need to wait for navigation here ? does it work regardless ?

packages/core/src/discovery.js
chinmay-browserstack: Screenshot 2024-09-19 at 11 06 58 PM
This is correct. check schema

packages/core/src/discovery.js
chinmay-browserstack: Checked for all captured resources(in serializeDOM). we always add a uid (random 36 character string). so it should always be unique url.
serialized canvas - uid (data-percy-element-id)
serialized videos - uid (data-percy-element-id)
stylesheets - uid

packages/core/src/config.js
```diff
@@ -291,6 +295,7 @@ export const snapshotSchema = {
         domTransformation: { $ref: '/config/snapshot#/properties/domTransformation' },
         enableLayout: { $ref: '/config/snapshot#/properties/enableLayout' },
         sync: { $ref: '/config/snapshot#/properties/sync' },
+        multiDOM: { $ref: '/config/snapshot#/properties/multiDOM' },
```
Amit3200, line 298: is `multiDOM` confirmed?

packages/core/src/discovery.js
```diff
@@ -195,16 +223,21 @@ async function* captureSnapshotResources(page, snapshot, options) {
...
-    width
-  });
+  let resizePage = width => {
+    page.network.currentWidth = width;
+    return page.resize({
+      height: snapshot.minHeight,
+      deviceScaleFactor,
+      mobile,
+      width
+    });
```
Amit3200, line 233: Generally if we are having too many CDP commands shouldn't we consider creating a wrapper class which will receive parameters and do the things using that?

packages/core/src/discovery.js
Amit3200: @chinmay-browserstack  also account for the additional cookies which we are planning to capture via SDKs, how they are going to be passed here?

packages/core/src/discovery.js
```diff
@@ -365,8 +400,15 @@ export function createDiscoveryQueue(percy) {
               disableCache: snapshot.discovery.disableCache,
               allowedHostnames: snapshot.discovery.allowedHostnames,
               disallowedHostnames: snapshot.discovery.disallowedHostnames,
-              getResource: u => snapshot.resources.get(u) || cache.get(u),
-              saveResource: r => { snapshot.resources.set(r.url, r); if (!r.root) { cache.set(r.url, r); } }
+              getResource: (u, width = null) => {
```
Amit3200, line 403: is this responsible for returning the resource for that width?
If width is not present what are we planning to do, I am seeing we are returning first index are we sure its works?

packages/core/src/network.js
```diff
@@ -354,7 +354,7 @@ async function sendResponseResource(network, request, session) {
   let send = (method, params) => network.send(session, method, params);
 
   try {
-    let resource = network.intercept.getResource(url);
+    let resource = network.intercept.getResource(url, network.currentWidth);
```
Amit3200, line 357: apart from `network.currentWidth` is there any way to get any width this looks little weird.

packages/core/src/page.js
chinmay-browserstack: about:blank is an empty page and does not load any external resources so i don't think we need to wait for it to load. it is instantaneous

packages/core/src/discovery.js
chinmay-browserstack: we can pass cookies is same fashion as we currently do inside domSnapshot object. we will override this value

packages/core/src/discovery.js
chinmay-browserstack: Actually page class is a wrapper only for running different cdp commands on browser.

packages/core/src/discovery.js
chinmay-browserstack: Yes it works. since only root resource is an array and in case if multiDOM is disabled we are storing root resource as a array only with no widths. that case is working fine. for other resources no change in logic

packages/core/src/network.js
chinmay-browserstack: we can use cdp to get current window width but that will be slower operation as compared to this.

packages/core/src/discovery.js
```diff
@@ -195,16 +223,19 @@ async function* captureSnapshotResources(page, snapshot, options) {
...
 
   // used to resize the using capture options
-  let resizePage = width => page.resize({
-    height: snapshot.minHeight,
-    deviceScaleFactor,
-    mobile,
-    width
-  });
+  let resizePage = width => {
+    page.network.currentWidth = width;
```
ninadbstack, line 227: Network is not directly related to current width. Dont update in network. Update in intercept instead as intercept needs current width to save/retrieve resources. Network just passes it to intercept in the end

packages/core/src/discovery.js
ninadbstack: Any specific reason to put width and domSnapshot separately ? shouldnt width be part of it ? 

Especially considering you are defining it in schema separately but not passing it here on line 97 ?

packages/core/src/config.js
ninadbstack: Doesnt sound very straightforward for end user. `captureMultipleWidths` seems better - it does exactly what it says.

packages/core/src/network.js
ninadbstack: I have commented to make it better

packages/core/src/page.js
ninadbstack: have we added tests for this ? or ^ is from chromium documentation ?

packages/core/src/api.js
```diff
@@ -62,6 +62,12 @@ export function createPercyServer(percy, port) {
       build: percy.testing?.build ?? percy.build,
       loglevel: percy.loglevel(),
       config: percy.config,
+      widths: {
+        // This is always needed even if width is passed
+        mobile: percy.deviceDetails ? percy.deviceDetails.map((d) => d.width) : [],
```
ninadbstack, line 67: Shouldnt we send dpr as well ?

chinmay-browserstack: doesn't make sense as we are just resizing page and not reloading page so changing dpr wouldn't help

packages/core/test/discovery.test.js
```diff
@@ -2885,4 +2885,124 @@ describe('Discovery', () => {
       }));
     });
   });
+
+  describe('Handles multiple root resources', () => {
+    it('gathers resources for a snapshot', async () => {
+      let DOM1 = testDOM.replace('Percy!', 'Percy! at 370');
+      let DOM2 = testDOM.replace('Percy!', 'Percy! at 765');
+      const capturedResource = {
```
ninadbstack, line 2893: Please update this test such that there is 1 resource in DOM1 which is not there in DOM2 or testDOM and similar for DOM2 

So that your assertions will contain those 2 unique resources as well.

packages/core/src/discovery.js
```diff
@@ -69,6 +69,11 @@ function debugSnapshotOptions(snapshot) {
   debugProp(snapshot, 'clientInfo');
   debugProp(snapshot, 'environmentInfo');
   debugProp(snapshot, 'domSnapshot', Boolean);
+  if (snapshot.responsiveSnapshotCapture) {
+    debugProp(snapshot, 'domSnapshot.0.userAgent');
```
ninadbstack, line 73: this can fail if user has set `responsiveSnapshotCapture` in cli yml but is using older sdk which wont use this. 
Update condition to check if domSnapshot is a list or not

ninadbstack: Also I am assuming that you are not using this value yet and it will be a different pr. but we need to use this in default requestHeaders

chinmay-browserstack: yes that will be in separate PR as this PR already has lots of things

Amit3200 (approved):
lgtm

ninadbstack (approved):

PR #1730, opened 2024-09-18
adding secure flag for cookies that starts with __Secure

prklm10:
adding secure flag for cookies that starts with __Secure because any cookie that starts with __Secure. The browser search for secure flag.

Docs -> https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie


pankaj443 (approved):
LGTM

PR #1734, opened 2024-09-19
Throttle Logs Endpoint from CLI

amandeepsingh333:
**problem** : 
 if large number(500) of snapshots are failing or have some issues for the customer then we are causing Rate limiting to get triggered on `api/v1/suggestions/from_logs` as we fetch suggestions for all snapshots. This triggers alert on our side.

**approach:**
To tackle this we have CLI in a way that we dont ask for more than 10 suggestions per build. We are basically limiting the maximum suggestions we can suggest per build. The idea is that if all snapshots are getting some suggestions in build they would be probably same.

**implementation:**
we have stored a counter of suggestionsCalls and will stop the call to end the count is > 10.


packages/core/src/percy.js
```diff
@@ -507,8 +511,17 @@ export class Percy {
...
+      if (this.suggestionsCallCounter > MAX_SUGGESTION_CALLS) {
+        if (this.suggestionsCallCounter === MAX_SUGGESTION_CALLS + 1) {
+          this.log.debug('Rate limit exceeded for Maximum allowed suggestions per build.');
+        } else {
+          return;
+        }
+      } else {
+        const suggestionResponse = await this.client.getErrorAnalysis(errors);
+        this.#displaySuggestionLogs(suggestionResponse, options);
+      }
```
chinmay-browserstack, line 524:
```suggestion
      this.suggestionsCallCounter++;
      if (this.suggestionsCallCounter > MAX_SUGGESTION_CALLS) {
        if (this.suggestionsCallCounter === MAX_SUGGESTION_CALLS + 1) {
          this.log.debug('Rate limit exceeded for Maximum allowed suggestions per build.');
        }
        return;
      } 
      const suggestionResponse = await this.client.getErrorAnalysis(errors);
      this.#displaySuggestionLogs(suggestionResponse, options);
```

packages/core/test/percy.test.js
```diff
@@ -1301,6 +1301,40 @@ describe('Percy', () => {
...
+          await expectAsync(percy.suggestionsForFix('some_error')).toBeResolved();
+        }
+
+        await expectAsync(percy.suggestionsForFix('some_error')).toBeResolved();
+        expect(logger.stderr).toEqual(jasmine.arrayContaining([
+          '[percy:core] Rate limit exceeded for Maximum allowed suggestions per build.'
+        ]));
+      });
+
+      it('should not call getErrorAnalysis after exceeding the rate limit', async () => {
```
chinmay-browserstack, line 1323: can you update this case for checking multiple logs are not getting printed if we cross threshold

chinmay-browserstack (approved):

ninadbstack (approved):

prklm10 (approved):

PR #1764, opened 2024-10-23
Added cookies support in direct requests

ninadbstack:
Context:
- We missed using cookies in the direct request done for woff2 font files
- The request object coming with cdp does not contain cookie header - mostly its added later
- So now we fetch correct cookies for url using cdp and use those cookies in direct request

### DIFF
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -89,7 +89,7 @@ function parseCookies(cookies) {
   // If cookies is collected via SDK
   if (Array.isArray(cookies) && cookies.every(item => typeof item === 'object' && 'name' in item && 'value' in item)) {
     // omit other fields reason sometimes expiry comes as actual date where we expect it to be double
-    return cookies.map(c => ({ name: c.name, value: c.value, secure: c.secure }));
+    return cookies.map(c => ({ name: c.name, value: c.value, secure: c.secure, domain: c.domain }));
   }
 
   if (!(typeof cookies === 'string' && cookies !== '')) return null;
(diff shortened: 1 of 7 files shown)


packages/core/src/discovery.js
```diff
@@ -89,7 +89,7 @@ function parseCookies(cookies) {
   // If cookies is collected via SDK
   if (Array.isArray(cookies) && cookies.every(item => typeof item === 'object' && 'name' in item && 'value' in item)) {
     // omit other fields reason sometimes expiry comes as actual date where we expect it to be double
-    return cookies.map(c => ({ name: c.name, value: c.value, secure: c.secure }));
+    return cookies.map(c => ({ name: c.name, value: c.value, secure: c.secure, domain: c.domain }));
```
chinmay-browserstack, line 92: domain will get overridden in page.js
https://github.com/percy/cli/blob/d86b39a87fdacc10398e78d1202781fdbcb1e17b/packages/core/src/page.js#L89

packages/core/src/discovery.js
```diff
@@ -218,7 +218,7 @@ async function* captureSnapshotResources(page, snapshot, options) {
   const log = logger('core:discovery');
   let { discovery, additionalSnapshots = [], ...baseSnapshot } = snapshot;
   let { capture, captureWidths, deviceScaleFactor, mobile, captureForDevices } = options;
-  let cookies = snapshot?.domSnapshot?.cookies || snapshot?.domSnapshot?.[0]?.cookies;
+  let cookies = snapshot.domSnapshot?.cookies || snapshot.domSnapshot?.[0]?.cookies;
```
prklm10, line 221:
```suggestion
  let cookies = snapshot.domSnapshot?.cookies || (Array.isArray(snapshot.domSnapshot) && snapshot.domSnapshot[0]?.cookies);
```

ninadbstack: Are you recommending it for change or style ?

prklm10: no looks good

prklm10 (approved):

PR #1765, opened 2024-10-24
Added default headers for chromium in direct requests

ninadbstack:
Context:
- It seems like some security softwares / or cdns have started requiring these headers - else they return 403s 
- I am assuming that addition of extra headers would not affect where its already working


prklm10 (approved):
approved. As this request is made only for font. So this should not affect other calls.

PR #1778, opened 2024-11-12
Adding visual scanner support

rishigupta1599:


ninadbstack (approved):

ninadbstack:
Also update schedular code to exclude any projects with testhub id so that you can test in prod with few projects without touching other projects

PR #1779, opened 2024-11-12
Fix import in ternary statement

khushhalm:
It will fix the bundling issue of Percy CLI lib when trying to bundle Percy inside LCNC Electron app

### DIFF
--- a/packages/client/src/utils.js
+++ b/packages/client/src/utils.js
@@ -135,7 +135,7 @@ export async function request(url, options = {}, callback) {
   let { protocol, hostname, port, pathname, search, hash } = new URL(url);
 
   // reference the default export so tests can mock it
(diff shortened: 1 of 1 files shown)


ninadbstack (approved):

packages/client/src/utils.js
```diff
@@ -135,7 +135,9 @@ export async function request(url, options = {}, callback) {
   let { protocol, hostname, port, pathname, search, hash } = new URL(url);
 
   // reference the default export so tests can mock it
-  let { default: http } = await import(protocol === 'https:' ? 'https' : 'http');
+  // bundling cli insdie electron (LCNC) fails if we import it like
```
ninadbstack, line 138:
```suggestion
  // bundling cli inside electron (LCNC) fails if we import it like 
```

prklm10 (approved):

PR #1784, opened 2024-11-19
PAC file support via PERCY_PAC_FILE_URL env var

khushhalm:
- Added support for PAC file based Proxy

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,13 +31,13 @@
     "@babel/eslint-parser": "^7.14.7",
     "@babel/preset-env": "^7.14.7",
     "@babel/register": "^7.17.7",
-    "babel-plugin-transform-import-meta": "^2.2.1",
     "@rollup/plugin-alias": "^5.1.1",
     "@rollup/plugin-babel": "^6.0.0",
     "@rollup/plugin-commonjs": "^21.0.0",
     "@rollup/plugin-node-resolve": "^15.0.0",
     "babel-plugin-istanbul": "^6.0.0",
     "babel-plugin-module-resolver": "^4.1.0",
+    "babel-plugin-transform-import-meta": "^2.2.1",
     "cross-env": "^7.0.2",
     "eslint": "^7.30.0",
     "eslint-config-standard": "^16.0.2",
@@ -59,5 +59,8 @@
     "nyc": "^15.1.0",
     "rollup": "^2.53.2",
     "tsd": "^0.31.2"
+  },
+  "dependencies": {
+    "pac-proxy-agent": "^7.0.2"
   }
 }
--- a/package.json
+++ b/package.json
@@ -59,8 +59,5 @@
     "nyc": "^15.1.0",
     "rollup": "^2.53.2",
     "tsd": "^0.31.2"
-  },
-  "dependencies": {
-    "pac-proxy-agent": "^7.0.2"
   }
 }
--- a/packages/client/src/proxy.js
+++ b/packages/client/src/proxy.js
@@ -4,9 +4,27 @@ import http from 'http';
 import https from 'https';
 import logger from '@percy/logger';
 import { stripQuotesAndSpaces } from '@percy/env/utils';
+import { PacProxyAgent } from 'pac-proxy-agent';
 
 const CRLF = '\r\n';
 const STATUS_REG = /^HTTP\/1.[01] (\d*)/;
+
+// function to create PAC proxy agent
+function createPacAgent(pacUrl, options = {}) {
+  try {
+    const agent = new PacProxyAgent(pacUrl, {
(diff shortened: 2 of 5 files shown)


packages/client/src/proxy.js
```diff
@@ -4,10 +4,28 @@ import http from 'http';
 import https from 'https';
 import logger from '@percy/logger';
 import { stripQuotesAndSpaces } from '@percy/env/utils';
+import { PacProxyAgent } from 'pac-proxy-agent';
```
ninadbstack, line 7: missed adding this dependency ?

packages/client/src/proxy.js
```diff
@@ -219,11 +237,31 @@ export function proxyAgentFor(url, options) {
...
 
-  return cache.get(cachekey);
+  try {
+    let agent;
+    const pacUrl = process.env.PERCY_PAC_FILE_PATH;
+
+    // If PAC URL is provided, use PAC proxy
+    if (pacUrl) {
+      logger('client:proxy').info(`Using PAC file from: ${pacUrl}`);
+      agent = createPacAgent(stripQuotesAndSpaces(pacUrl), options);
```
ninadbstack, line 252:
```suggestion
      agent = createPacAgent(pacUrl, options);
```
Move that inside the `createPacAgent` function

packages/client/src/proxy.js
```diff
@@ -4,10 +4,28 @@ import http from 'http';
...
 const CRLF = '\r\n';
 const STATUS_REG = /^HTTP\/1.[01] (\d*)/;
 
+// function to create PAC proxy agent
+function createPacAgent(pacUrl, options = {}) {
+  try {
+    const agent = new PacProxyAgent(pacUrl, {
+      keepAlive: true,
+      ...options,
+      rejectUnauthorized: false
```
ninadbstack, line 18: That is by default setting behaviour of `NODE_TLS_REJECT_UNAUTHORIZED` - ideally should be governed by this env var

packages/client/src/proxy.js
```diff
@@ -219,11 +237,31 @@ export function proxyAgentFor(url, options) {
...
-      : new ProxyHttpAgent(options));
+  // If we already have a cached agent, return it
+  if (cache.has(cachekey)) {
+    return cache.get(cachekey);
   }
 
-  return cache.get(cachekey);
+  try {
+    let agent;
+    const pacUrl = process.env.PERCY_PAC_FILE_PATH;
```
ninadbstack, line 247: Is this expecting a URL or File path ? or either is fine ? 

considering the most common way would be to use url - we should call this `PERCY_PAC_FILE_URL`

packages/client/src/proxy.js
ninadbstack, line 1: Missing specs

packages/client/src/proxy.js
khushhalm: yes, added now. Please check that it is at the right place. (It is in the base package json)

packages/client/src/proxy.js
khushhalm: It can be either.

> considering the most common way would be to use url - we should call this PERCY_PAC_FILE_URL

Makes sense, naming it `PERCY_PAC_FILE_URL `

package.json
```diff
@@ -59,5 +59,8 @@
     "nyc": "^15.1.0",
     "rollup": "^2.53.2",
     "tsd": "^0.31.2"
+  },
+  "dependencies": {
+    "pac-proxy-agent": "^7.0.2"
```
ninadbstack, line 64: This should be in package.json of core package

packages/client/src/proxy.js
khushhalm: Can I do this later? Can create a Jira so that this is not missed.
Need to share the custom build to GM by today to test PAC support.

Or if it blocker can I take help from Percy Ops for this? Since need to test couple of other things.

packages/client/src/proxy.js
khushhalm: Will handle this on LCNC side since this is needed in 1 case.

package.json
khushhalm: If we include it inside module itself, then it should be inside client module, right?

khushhalm: @ninadbstack for now I've added in client one, let me know if that is incorrect or something

packages/client/src/proxy.js
ninadbstack: No, we dont merge anything without specs on CLI - I can release alpha version from branch without specs if you want - so that its not merged to master

package.json
ninadbstack: Correct, my bad, it should be in client.

ninadbstack (approved):

PR #1786, opened 2024-11-20
tests for pac file support

surya91521:


packages/client/test/unit/proxy.test.js
```diff
@@ -0,0 +1,80 @@
+import { ProxyHttpAgent, ProxyHttpsAgent, createPacAgent, getProxy, proxyAgentFor } from '../../src/proxy.js';
+import { PacProxyAgent } from 'pac-proxy-agent';
+
+describe('proxy', () => {
+
+  describe('getProxy', () => {
+    it('should return proxy object if proxy is set', () => {
+      process.env.http_proxy = 'http://proxy.com:8080';
```
ninadbstack, line 8: update env vars in before and after, so that its not skipped in case test fails.

packages/client/test/unit/proxy.test.js
```diff
@@ -0,0 +1,80 @@
...
+    it('should create and cache new HTTPS agent if not available', () => {
+      const url = 'https://example.com';
+      const options = {};
+      const agent = proxyAgentFor(url, options);
+      expect(agent).toBeInstanceOf(ProxyHttpsAgent);
+      expect(proxyAgentFor.cache.get('https://example.com')).toBe(agent);
+    });
+
+    it('should create PAC proxy agent if PAC URL is provided', () => {
+      process.env.PERCY_PAC_FILE_URL = 'http://example.com/proxy.pac';
```
ninadbstack, line 69: same here

packages/client/test/unit/proxy.test.js
surya91521: Changed

packages/client/test/unit/proxy.test.js
surya91521: Changed

PR #1793, opened 2024-11-25
Fixing Race Condition incase of Screenshot Cmd

Amit3200:
- Fixing Race Condition incase of Screenshot Cmd
- Currently when calling `percyScreenshot` command the tasks were getting deleted as we used to find them using snapshot name.
- Since in screenshot command we upload comparisons they are allowed to have same snapshot name so now deletion cannot work, so we need to add find with tag name.

### DIFF
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -512,3 +512,19 @@ export class DefaultMap extends Map {
     return super.get(key);
   };
 };
+
+export function compareObjectTypes(obj1, obj2) {
+  if (obj1 === obj2) return true; // Handles primitives
+  if (typeof obj1 !== 'object' || typeof obj2 !== 'object' || obj1 === null || obj2 === null) return false;
+
+  const keys1 = Object.keys(obj1);
+  const keys2 = Object.keys(obj2);
+
+  if (keys1.length !== keys2.length) return false;
+
+  for (const key of keys1) {
+    if (!keys2.includes(key) || !compareObjectTypes(obj1[key], obj2[key])) return false;
+  }
+
+  return true;
+}
--- a/packages/core/src/snapshot.js
+++ b/packages/core/src/snapshot.js
@@ -8,7 +8,8 @@ import {
   hostnameMatches,
   yieldTo,
   snapshotLogName,
-  decodeAndEncodeURLWithLogging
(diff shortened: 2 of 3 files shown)


chinmay-browserstack (approved):

packages/core/src/utils.js
```diff
@@ -512,3 +512,19 @@ export class DefaultMap extends Map {
     return super.get(key);
   };
 };
+
+export function compareObjectTypes(obj1, obj2) {
+  if (obj1 === obj2) return true; // Handles primitives
```
this-is-shivamsingh, line 517: - Do we need this, as this only check obj1 and obj2 reference and not the actual value.
- Below we are already doing this check `key in key2`

Amit3200: above is early exit for the primitive check as we don't want to spend time in converting data type and checking. Check lodash implementation may be.

packages/core/src/utils.js
```diff
@@ -512,3 +512,19 @@ export class DefaultMap extends Map {
...
+  if (typeof obj1 !== 'object' || typeof obj2 !== 'object' || obj1 === null || obj2 === null) return false;
+
+  const keys1 = Object.keys(obj1);
+  const keys2 = Object.keys(obj2);
+
+  if (keys1.length !== keys2.length) return false;
+
+  for (const key of keys1) {
+    if (!keys2.includes(key) || !compareObjectTypes(obj1[key], obj2[key])) return false;
+  }
```
this-is-shivamsingh, line 527:
```
  name: tag.name || null,
                width: tag.width || null,
                height: tag.height || null,
                'os-name': tag.osName || null,
                'os-version': tag.osVersion || null,
                orientation: tag.orientation || null,
                'browser-name': tag.browserName || null,
                'browser-version': tag.browserVersion || null,
                resolution: tag.resolution || null
```
These are the only tag values, does it needes recursive checking ?

Amit3200: Tomorrow the tag can change and evolve, we need to ensure that cli is fail proof so that we don't miss cases.

this-is-shivamsingh: - Resolved it

pankaj443 (approved):

PR #1803, opened 2024-11-28
Accounting base64 encoding for resources size validation

rishigupta1599:
Resources which were greater than 25 MB in size we want to skip them in resource array as they might fail to upload, but when we upload we do base64 encoding which increases the size but while validation we were not accounting for increase in size due to encoding.
Base64 has 33-37% increase in content size so we are reducing max size check by that much factor

### DIFF
--- a/packages/client/src/client.js
+++ b/packages/client/src/client.js
@@ -2,7 +2,6 @@ import fs from 'fs';
 import PercyEnv from '@percy/env';
 import { git } from '@percy/env/utils';
 import logger from '@percy/logger';
-import Pako from 'pako';
 
 import {
   pool,
@@ -14,7 +13,8 @@ import {
   waitForTimeout,
   validateTiles,
   formatLogErrors,
-  tagsList
+  tagsList,
+  getEncodedContent
 } from './utils.js';
 
 // Default client API URL can be set with an env var for API development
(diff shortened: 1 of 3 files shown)


Amit3200 (approved):

packages/client/src/client.js
```diff
@@ -342,11 +342,8 @@ export class PercyClient {
     validateId('build', buildId);
     if (filepath) {
       content = await fs.promises.readFile(filepath);
-      if (process.env.PERCY_GZIP) {
-        content = Pako.gzip(content);
-      }
     }
-    let encodedContent = base64encode(content);
+    let encodedContent = getEncodedContent(content);
```
ninadbstack, line 346: There is already percy gzip check inside discovery. So this will cause double gzip

rishigupta1599: In network file we are not setting it to encoded content we are just using it for size calculation

rishigupta1599: client code is existing one just moved it to method

rishigupta1599: Should I remove then?

ninadbstack: No its not :P you missed a condition

ninadbstack (approved):

Amit3200 (approved):

PR #1805, opened 2024-11-29
Adding support for IGNORE_TIMEOUT_ERROR env

Amit3200:
- Adding support for PERCY_IGNORE_TIMEOUT_ERROR env
- This will ensure that if above env is set to true, we would regardless upload snapshots instead of failing them.

### DIFF
--- a/packages/core/src/network.js
+++ b/packages/core/src/network.js
@@ -154,7 +154,7 @@ export class Network {
       msg += this.logNetworkRequests(filter);
     }
 
-    if (process.env.IGNORE_TIMEOUT_ERROR === 'true') {
+    if (process.env.PERCY_IGNORE_TIMEOUT_ERROR === 'true') {
       let warnMsg = 'Ignoring network timeout failures.';
       warnMsg += this.logNetworkRequests(filter);
       this.log.warn(warnMsg);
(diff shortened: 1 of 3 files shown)


rishigupta1599 (approved):

packages/core/src/network.js
```diff
@@ -141,11 +141,24 @@ export class Network {
...
+
   // Throw a better network timeout error
   _throwTimeoutError(msg, filter = () => true) {
     if (this.log.shouldLog('debug')) {
-      let reqs = Array.from(this.#requests.values()).filter(filter).map(r => r.url);
-      msg += `\n\n  ${['Active requests:', ...reqs].join('\n  - ')}\n`;
+      msg += this.logNetworkRequests(filter);
+    }
+
+    if (process.env.PERCY_IGNORE_TIMEOUT_ERROR === 'true') {
```
ninadbstack, line 157: This change only ignores network idle timeouts. Not page load. That you need to handle in page.ja

Amit3200: `page.js` calls the function `_throwTimeoutError` to throw the error so if its set that is also handled.

prklm10 (approved):

PR #1814, opened 2024-12-05
Cs snapshot queue

revathySV:
Adding a different concurrency count for snapshot processing

### DIFF
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -259,6 +259,10 @@ export const configSchema = {
         type: 'integer',
         minimum: 1
       },
+      snapshotConcurrency: {
+        type: 'integer',
+        minimum: 1
+      },
       retry: {
         type: 'boolean',
         default: false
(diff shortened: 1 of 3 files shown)


Amit3200 (approved):

Amit3200 (approved):

packages/client/src/client.js
```diff
@@ -367,6 +368,7 @@ export class PercyClient {
     validateId('build', buildId);
     this.log.debug(`Uploading resources for ${buildId}...`, meta);
 
+    const uploadConcurrency = parseInt(process.env.PERCY_RESOURCE_UPLOAD_CONCURRENCY) || 2;
```
pankaj443, line 371: is this new env var? Should it be added to readme/docs if new

chinmay-browserstack: This we have added only for central scanner. i don't think some user needs to update this config.

pankaj443: Ack

pankaj443 (approved):

PR #1820, opened 2024-12-11
Fix[ISSUE_1819]: Support for serialising ShadowRoot Element for Chrome 131 onwards

this-is-shivamsingh:
Changes:
1. From chrome 131 onward, there is a restriction on serialising shadowroot elements
2. To serialise the the shadow element perfectly, we need to explicitly pass the shadowRoot elements with `getHTML`
     Refer here: https://developer.mozilla.org/en-US/docs/Web/API/Element/getHTML#shadowroots

#### Testing:
1. Tested with CLI flow ( that contains Chrome 123 )
2. Tested with firefox as test browser from versions in between (126 - 132)
3. Tested with Chrome 131 ( latest )

### DIFF
--- a/packages/dom/src/clone-dom.js
+++ b/packages/dom/src/clone-dom.js
@@ -84,20 +84,20 @@ export function cloneNodeAndShadow(ctx) {
 /**
  * Use `getInnerHTML()` to serialize shadow dom as <template> tags. `innerHTML` and `outerHTML` don't do this. Buzzword: "declarative shadow dom"
  */
-export function getOuterHTML(docElement) {
-  // All major browsers in latest versions supports getHTML API to get serialized DOM
-  // https://developer.mozilla.org/en-US/docs/Web/API/Element/getHTML
-  // old firefox doesn't serialize shadow DOM, we're awaiting API's by firefox to become ready and are not polyfilling it.
-  // new firefox from 128 onwards serializes it using getHTML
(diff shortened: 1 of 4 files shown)


packages/dom/src/clone-dom.js
```diff
@@ -84,20 +84,20 @@ export function cloneNodeAndShadow(ctx) {
 /**
  * Use `getInnerHTML()` to serialize shadow dom as <template> tags. `innerHTML` and `outerHTML` don't do this. Buzzword: "declarative shadow dom"
  */
-export function getOuterHTML(docElement) {
-  // All major browsers in latest versions supports getHTML API to get serialized DOM
-  // https://developer.mozilla.org/en-US/docs/Web/API/Element/getHTML
-  // old firefox doesn't serialize shadow DOM, we're awaiting API's by firefox to become ready and are not polyfilling it.
-  // new firefox from 128 onwards serializes it using getHTML
-  /* istanbul ignore if: Only triggered in firefox <= 128 and tests runs on latest */
-  if (!docElement.getHTML) { return docElement.outerHTML; }
```
this-is-shivamsingh, line 93: -> Need to remove this, as this was blocking snapshot CLI flow, as our discovery is Chrome 123, and `getHTML` is only supported from Chrome 125 onwards

Amit3200 (approved):

pankaj443 (approved):
LGTM 🎸

PR #1829, opened 2024-12-19
Pac file test

bhutianimukul:
Did the fixes for test case failing and added tests for test coverage
Reference PR
https://github.com/percy/cli/pull/1787


shantanuk-browserstack:
LGTM!

packages/client/src/client.js
```diff
@@ -83,6 +83,8 @@ export class PercyClient {
 
   // Stringifies client and environment info.
   userAgent() {
+    if (this.env.forcedPkgValue) pkg = this.env.forcedPkgValue;
```
prklm10, line 86: Add that this is added by LCNC team and why?

packages/core/src/findRevision.sh
```diff
@@ -0,0 +1,37 @@
...
+
+if (($# < 1)); then
+  printf "usage: \n"
+  printf "  ./get_chromuim [-r] rev    - will get chromium by revision\n"
+  exit 1
+fi
+
+export REVISION=$1
+
+for os in "Linux" "Mac" "Mac_Arm" "Win" "Win_x64";
```
prklm10, line 33: from where this  bash file is getting called?

prklm10: is this something lcnc calling internally?

ninadbstack: Oh no, thats something probably was committed from my local - we can remove it for now. I wrote it to find version numbers for discovery chromium update

bhutianimukul: Should I remove the file and push, as this came because master pull

shantanuk-browserstack (approved):

prklm10 (approved):

PR #1848, opened 2025-01-22
Cache base64 resources

ninadbstack:
Changes:
- Now we cache data src resources - is helpful in some cases where we were getting hundreds of repeated resources as those were on different dom nodes
- We now set src on cloned node using `data-percy-serialized-attribute-src` - which is later replaced with src. This allows us to avoid network call that browser makes when we set src on any cloned node.

### DIFF
--- a/packages/dom/src/serialize-base64.js
+++ b/packages/dom/src/serialize-base64.js
@@ -12,7 +12,7 @@ function getBase64Substring(src) {
   return src.substring(base64Index);
 }
 
-export function serializeBase64(node, resources) {
+export function serializeBase64(node, resources, cache) {
   let src = node.src;
   let isHrefUsed = false;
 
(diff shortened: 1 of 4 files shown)


rishigupta1599 (approved):
LGTM

packages/dom/src/serialize-base64.js
```diff
@@ -27,15 +27,23 @@ export function serializeBase64(node, resources) {
...
+    node.href.baseVal = cache.get(base64String);
   } else {
-    node.src = resource.url;
+    // we use data-percy-serialized-attribute-src here instead of `src`.
+    // As soon as src is used the browser will try to load the resource,
+    // thus making a network call which would fail as this is a
+    // dynamic cached resource and not a resource that backend can serve.
+    // we later post converting domtree to html replace this with src
+    node.removeAttribute('src');
+    node.setAttribute('data-percy-serialized-attribute-src', cache.get(base64String));
```
pankaj443, line 46: @ninadbstack where are we resetting it back to src?

ninadbstack: check serializeHTML function

pankaj443: got it.

pankaj443 (approved):
LGTM

PR #1860, opened 2025-02-13
skipping unsafe key

Shivanshu-07:
Description: 
1. Sanitising keys for config file. 
2. Skipping unsafe keys.


packages/config/src/utils/normalize.js
```diff
@@ -66,4 +68,28 @@ export function normalize(object, options) {
   });
 }
 
+// Utility function to prevent prototype pollution
+export function isSafeKey(key) {
+  const unsafeKeys = ['__proto__', 'constructor', 'prototype', 'toString', 'valueOf',
```
shantanuk-browserstack, line 73: move this const outside method

shantanuk-browserstack (changes requested):

ninadbstack:
@Shivanshu-07 please add why are we doing this

ninadbstack (approved):

shantanuk-browserstack (approved):

PR #1865, opened 2025-02-19
Added normalization function to make options case-insensitive

bs-shobhitkumar:
Have added a normalization function in validateSnaphotOptions to convert any type of casing of options to camecase making them case-insensitive.

Example: 

If a user provides options as:
```
  options = {
          nAmE: 'Snapshot 1',
          url: 'http://localhost:8000',
          WiDtHs: [375, 1280],
          sCoPe: '#app',
          sCoPeOpTiOnS: { scroll: true },
          mInHeIgHt: 1024,
          eNaBlEJaVaScRiPt: true,
          eNaBlELaYoUt: false,
          cLiEnTInFo: 'client-info',
          eNvIrOnMeNtInFo: 'env-info',
          sYnC: true,
          tEsTcAsE: 'testCase',
          lAbElS: 'label1',
          tHtEsTcAsEeXeCuTiOnId: '12345'
        }
```

These will be converted to:
```
 options = {
        clientInfo: 'client-info',
        environmentInfo: 'env-info',
        name: 'Snapshot 1',
        url: 'http://localhost:8000',
        widths: [375, 1280],
        scope: '#app',
        scopeOptions: { scroll: true },
        minHeight: 1024,
        enableJavaScript: true,
        enableLayout: false,
        sync: true,
        testCase: 'testCase',
        labels: 'label1',
        thTestCaseExecutionId: '12345'
      }
```

### DIFF
--- a/packages/core/src/snapshot.js
+++ b/packages/core/src/snapshot.js
@@ -165,7 +165,7 @@ function getSnapshotOptions(options, { config, meta }) {
     }
   });
 }
-const optionMappings = {
+const OPTION_MAPPINGS = {
   name: 'name',
   widths: 'widths',
   scope: 'scope',
@@ -189,7 +189,7 @@ function normalizeOptions(options) {
 
   for (const key in options) {
     const lowerCaseKey = key.toLowerCase().replace(/[-_]/g, '');
-    const normalizedKey = optionMappings[lowerCaseKey] ? optionMappings[lowerCaseKey] : key;
+    const normalizedKey = OPTION_MAPPINGS[lowerCaseKey] ? OPTION_MAPPINGS[lowerCaseKey] : key;
     normalizedOptions[normalizedKey] = options[key];
   }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/snapshot.js
```diff
@@ -165,13 +165,41 @@ function getSnapshotOptions(options, { config, meta }) {
     }
   });
 }
+const optionMappings = {
```
shantanuk-browserstack, line 168: always keep constants in upper case, refer other places,

packages/core/src/snapshot.js
```diff
@@ -165,13 +165,41 @@ function getSnapshotOptions(options, { config, meta }) {
...
+  sync: 'sync',
+  testcase: 'testCase',
+  labels: 'labels',
+  thtestcaseexecutionid: 'thTestCaseExecutionId',
+  resources: 'resources',
+  meta: 'meta',
+  snapshot: 'snapshot'
+};
+
+function normalizeOptions(options) {
```
shantanuk-browserstack, line 187: move this to utils

shantanuk-browserstack (approved):

PR #1866, opened 2025-02-19
fixed content length NaN

Shivanshu-07:
Description 

Fixed content length NaN
Added check and debug log for NaN content-length
Added check to skip body length > 25MB. 
Added test case for NaN content-length


packages/core/src/network.js
```diff
@@ -471,6 +471,9 @@ async function saveResponseResource(network, request, session) {
   // maxPayload option of websocket defaulted to 100MB.
   // If content-length is more than our allowed 25MB, no need to process that resouce we can return log.
   let contentLength = parseInt(response.headers['Content-Length']);
+  if (isNaN(contentLength)) {
+    log.debug('- Content length is NaN for the resource', meta);
+  }
```
prklm10, line 476: we don't need it here. it can be done in line 497. As if condition in line 497 is executed the content length will be NaN.

Shivanshu-07: If we add this condition in 497, there may come some condition where body.length > 25MB. Only checking isNaN might break things in future.

prklm10: we don't want to check of NaN. we just log that above condition was not able to fetch content length.

Shivanshu-07: ack. Fixed

prklm10 (approved):

packages/core/test/discovery.test.js
```diff
@@ -535,6 +535,74 @@ describe('Discovery', () => {
...
+        })
+      })
+    ]);
+    expect(logger.stderr).toContain(
+      '[percy:core:discovery] - Content length is NaN for the resource'
+    );
+  });
+
+  it('does not capture remote files with content-length NAN greater than 25MB', async () => {
+    server.reply('/large.css', () => [200, 'text/css', 'A'.repeat(30_000_000)], { contentLengthNaN: true });
```
ninadbstack, line 573: make it headers

packages/core/src/network.js
```diff
@@ -470,7 +470,8 @@ async function saveResponseResource(network, request, session) {
   // Checing for content length more than 100MB, to prevent websocket error which is governed by
   // maxPayload option of websocket defaulted to 100MB.
   // If content-length is more than our allowed 25MB, no need to process that resouce we can return log.
-  let contentLength = parseInt(response.headers['Content-Length']);
+  let contentLength = response.headers['content-length'] || response.headers['Content-Length'];
```
ninadbstack, line 473: not a good way, please lowercase all and check - CONTENT-LENGTH is also probably a valid combination

packages/core/test/helpers/server.js
```diff
@@ -1,18 +1,22 @@
...
+export function createTestServer({ default: defaultReply, ...replies }, port = 8000, options = {}) {
   let server = new Server();
 
   // alternate route handling
-  let handleReply = reply => async (req, res) => {
+  let handleReply = (reply, options = {}) => async (req, res) => {
     let [status, headers, body] = typeof reply === 'function' ? await reply(req) : reply;
     if (!Buffer.isBuffer(body) && typeof body !== 'string') body = JSON.stringify(body);
+
+    if (options.noHeader) {
```
ninadbstack, line 12: shouldn't this be `noHeaders` ?

packages/core/test/discovery.test.js
```diff
@@ -535,6 +535,74 @@ describe('Discovery', () => {
...
+        })
+      })
+    ]);
+    expect(logger.stderr).toContain(
+      '[percy:core:discovery] - Missing headers for the requested resource.'
+    );
+  });
+
+  it('does not capture remote files with content-length NAN greater than 25MB', async () => {
+    server.reply('/large.css', () => [200, 'text/css', 'A'.repeat(30_000_000)], { contentLengthNaN: true });
```
ninadbstack, line 573: contentLengthNaN is not being used. You need to add appropriate header.

packages/core/src/network.js
```diff
@@ -470,7 +470,8 @@ async function saveResponseResource(network, request, session) {
   // Checing for content length more than 100MB, to prevent websocket error which is governed by
   // maxPayload option of websocket defaulted to 100MB.
   // If content-length is more than our allowed 25MB, no need to process that resouce we can return log.
-  let contentLength = parseInt(response.headers['Content-Length']);
+  let contentLength = response.headers?.[Object.keys(response.headers).find(key => key.toLowerCase() === 'content-length')];
```
ninadbstack, line 473: You dont have any tests that actually confirms that code works regardless of casing of header.

packages/core/src/network.js
Shivanshu-07: fixed

packages/core/src/network.js
```diff
@@ -491,6 +492,9 @@ async function saveResponseResource(network, request, session) {
         return log.debug('- Skipping remote resource', meta);
       } else if (!body.length) {
         return log.debug('- Skipping empty response', meta);
+      } else if (body.length > MAX_RESOURCE_SIZE) {
+        log.debug('- Content length is NaN for the resource', meta);
```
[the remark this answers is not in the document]
Shivanshu-07: done

packages/core/src/network.js
Shivanshu-07: Done

packages/core/test/discovery.test.js
Shivanshu-07: Done

packages/core/test/helpers/server.js
```diff
@@ -1,18 +1,25 @@
...
-  let handleReply = reply => async (req, res) => {
+  let handleReply = (reply, options = {}) => async (req, res) => {
     let [status, headers, body] = typeof reply === 'function' ? await reply(req) : reply;
     if (!Buffer.isBuffer(body) && typeof body !== 'string') body = JSON.stringify(body);
+
+    if (options.noHeaders) {
+      return res.writeHead(status).end(body);
+    }
+    if (options.contentLength) {
+      headers = { ...headers, ...options.contentLength };
```
ninadbstack, line 16: why should this be contentLength, thats not generic header

ninadbstack (approved):

PR #1881, opened 2025-03-18
Passing unique build uuid in create build

bhokaremoin:
Description: 
1. Passing unique build uuid from env to create builds.


shantanuk-browserstack (approved):

shantanuk-browserstack:
Discussed on thread regarding the ENVs.
LGTM

PR #1882, opened 2025-03-19
fixed style ordering

Shivanshu-07:
Description: 

1. Changed the order in which styles are captured.  
  a. If no style links are present then it is prepended.
  b. If style links are present the new once's are added in the order in which they are received.

### DIFF
--- a/packages/dom/src/serialize-cssom.js
+++ b/packages/dom/src/serialize-cssom.js
@@ -99,14 +99,27 @@ export function serializeCSSOM(ctx) {
         cache.set(sheet, resource.url);
       }
       styleLink.setAttribute('data-percy-adopted-stylesheets-serialized', 'true');
+      styleLink.setAttribute('data-percy-adopted-stylesheets-shivanshu-test', 'true');
       styleLink.setAttribute('data-percy-serialized-attribute-href', cache.get(sheet));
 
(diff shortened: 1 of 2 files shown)


packages/dom/src/serialize-cssom.js
```diff
@@ -104,9 +104,21 @@ export function serializeCSSOM(ctx) {
...
       if (clone.constructor.name === 'HTMLDocument' || clone.constructor.name === 'DocumentFragment') {
         // handle document and iframe
-        clone.body.prepend(styleLink);
+        // We are checking if we have multiple stylesheets present for the same clone or clone.body then we add
+        // them in the same order in which we receive them.
+        const lastLink = [...clone.body.querySelectorAll('link[data-percy-adopted-stylesheets-serialized]')].pop();
+        if (lastLink) {
+          lastLink.after(styleLink);
+        } else {
+          clone.body.prepend(styleLink);
```
pankaj443, line 113: why can we just use append directly? that should also fix this right?

pankaj443: querySelector('link[data-percy-adopted-stylesheets-serialized]:last-of-type');
can you check if this works

pankaj443 (approved):

PR #1888, opened 2025-03-26
✨  Adding support for algorithm and algorithmConfiguration in root level

prklm10:
adding support for algorithm and algorithmConfiguration in root level

### DIFF
--- a/packages/config/src/validate.js
+++ b/packages/config/src/validate.js
@@ -295,17 +295,20 @@ export function validate(data, key = '/config') {
   if (data.algorithmConfiguration) {
     const pathStr = 'algorithmConfiguration';
     const algorithmType = data.algorithm;
-    if (algorithmType) {
-      if (algorithmType === 'layout') {
-        errors.set(pathStr, {
-          path: pathStr,
-          message: `algorithmConfiguration is not applicable for '${algorithmType}' algorithm`
-        });
-      }
-    } else {
+
+    if (!algorithmType) {
       errors.set(pathStr, {
         path: pathStr,
         message: 'algorithmConfiguration needs algorigthm to be passed'
+      });
+    }
+
+    const nonAlgoConfigTypes = ['layout'];
+
+    if (nonAlgoConfigTypes.includes(algorithmType)) {
+      errors.set(pathStr, {
+        path: pathStr,
+        message: `algorithmConfiguration is not applicable for '${algorithmType}' algorithm`
       });
     }
   }
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -31,7 +31,7 @@ export const configSchema = {
         type: 'object',
         additionalProperties: false,
         properties: {
-          diffSensitivity: { type: 'integer', minimum: 0, maximum: 5 },
+          diffSensitivity: { type: 'integer', minimum: 0, maximum: 4 },
           imageIgnoreThreshold: { type: 'number', minimum: 0, maximum: 1 },
           carouselsEnabled: { type: 'boolean' },
           bannersEnabled: { type: 'boolean' },


packages/core/src/config.js
```diff
@@ -222,6 +222,21 @@ export const configSchema = {
           },
           required: ['algorithm']
         }
+      },
+      algorithm: {
+        type: 'string',
+        enum: ['standard', 'layout', 'intelliignore']
```
this-is-shivamsingh, line 228: ignore, will also be there

packages/core/src/config.js
```diff
@@ -222,6 +222,21 @@ export const configSchema = {
...
+      },
+      algorithmConfiguration: {
```
this-is-shivamsingh, line 230:
```suggestion
      algorithmParams: {
```

packages/core/src/config.js
```diff
@@ -222,6 +222,21 @@ export const configSchema = {
           },
           required: ['algorithm']
         }
+      },
+      algorithm: {
```
ninadbstack, line 226: dont duplicate this, use references

packages/core/src/config.js
```diff
@@ -222,6 +222,21 @@ export const configSchema = {
...
+        type: 'string',
+        enum: ['standard', 'layout', 'intelliignore']
+      },
+      algorithmConfiguration: {
+        type: 'object',
+        additionalProperties: false,
+        properties: {
+          diffSensitivity: { type: 'integer', minimum: 0 },
```
this-is-shivamsingh, line 234:
```suggestion
          diffSensitivity: { type: 'integer', minimum: 0 , maximum: 5}, we don't consider any value above this
```

packages/core/src/config.js
prklm10: No. We should not have ignore in fullpage.

packages/core/src/config.js
prklm10: It has been discussed.

packages/config/src/validate.js
```diff
@@ -292,6 +292,24 @@ export function validate(data, key = '/config') {
...
+        errors.set(pathStr, {
+          path: pathStr,
+          message: `algorithmConfiguration is not applicable for '${algorithmType}' algorithm`
+        });
+      }
+    } else {
+      errors.set(pathStr, {
+        path: pathStr,
+        message: 'algorithmConfiguration needs algorigthm to be passed'
+      });
```
this-is-shivamsingh, line 309:
```suggestion
   if (!algorithmType) errors.set()
   
   const non_algo_config_types = ['layout']
   if (non_algo_config_types.includes(algorithmType)) error.set(pathStr, {})
```

packages/core/src/config.js
this-is-shivamsingh, line 230: Ques:`algorithmConfiguration`, will it be a required field, or there will be some default value present for diffThreshold, diffSensivity value

prklm10: No it will be optional.

packages/core/src/config.js
prklm10: refactored.

packages/core/src/config.js
prklm10: updated.

packages/config/src/validate.js
prklm10: refactored.

packages/core/src/config.js
this-is-shivamsingh: @pankaj443, please review this part only

this-is-shivamsingh (approved):

packages/core/src/config.js
pankaj443: its handled in API

pankaj443 (approved):

PR #1892, opened 2025-04-02
Recreating custom element and not cloning them directly as it causes attributeCallback function to get triggered

prklm10:
Recreating custom element and not cloning them directly as it causes attributeCallback function to get triggered

Issue ->
- When were were doing cloneNode on a custom element. If attributeCallback is defined on that custom element then it was getting triggered, which caused spike in cpu usage.
- We are now recreating the node and not cloning them.

### DIFF
--- a/packages/dom/src/clone-dom.js
+++ b/packages/dom/src/clone-dom.js
@@ -15,7 +15,12 @@ import { handleErrors } from './utils';
 
 const ignoreTags = ['NOSCRIPT'];
 
-function cloneCustomElementWithoutAttributeChanged(element) {
+/**
+ * if a custom element has attribute callback then cloneNode calls a callback that can
+ * increase CPU load or some other change.
+ * So we want to make sure that it is not called when doing serialization.
+*/
+function cloneElementWithoutLifecycle(element) {
   if (!(element.attributeChangedCallback) || !element.tagName.includes('-')) {
     return element.cloneNode(); // Standard clone for non-custom elements
   }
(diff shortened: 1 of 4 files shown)


packages/dom/src/clone-dom.js
```diff
@@ -15,6 +15,26 @@ import { handleErrors } from './utils';
 
 const ignoreTags = ['NOSCRIPT'];
 
+function cloneCustomElementWithoutAttributeChanged(element) {
```
ninadbstack, line 18: Add a comment on why we need this for future dev

ninadbstack: Name it correctly

packages/dom/src/clone-dom.js
```diff
@@ -63,7 +89,8 @@ export function cloneNodeAndShadow(ctx) {
 
       // shallow clone should not contain children
       if (clone.children) {
-        Array.from(clone.children).forEach(child => clone.removeChild(child));
+        /* istanbul ignore next */
```
prklm10, line 92: some how due to this like functional test coverage is not at 100%. I see that this line is called 1k time.

ninadbstack (approved):

packages/dom/src/clone-dom.js
```diff
@@ -32,7 +57,8 @@ export function cloneNodeAndShadow(ctx) {
       // mark the node before cloning
       markElement(node, disableShadowDOM);
 
-      let clone = node.cloneNode();
+      let clone = cloneElementWithoutLifecycle(node);
+      // let clone = node.cloneNode();
```
ninadbstack, line 61: rempve tjos

PR #1898, opened 2025-04-11
limited resources to max limit

amandeepsingh333:
limited resources to max limit

### DIFF
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -133,15 +133,22 @@ function parseDomResources({ url, domSnapshot }) {
   if (!domSnapshot) return map;
   let allRootResources = new Set();
   let allResources = new Set();
+  let limitResources = process.env.LIMIT_SNAPSHOT_RESOURCES || false;
+  const MAX_RESOURCES = 750 - 2;
 
   if (!Array.isArray(domSnapshot)) {
     domSnapshot = [domSnapshot];
   }
 
   for (let dom of domSnapshot) {
+    if (limitResources && allResources.size + allRootResources.size >= MAX_RESOURCES) break;
     let isHTML = typeof dom === 'string';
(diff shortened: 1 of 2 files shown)


packages/core/src/discovery.js
```diff
@@ -133,15 +133,22 @@ function parseDomResources({ url, domSnapshot }) {
   if (!domSnapshot) return map;
   let allRootResources = new Set();
   let allResources = new Set();
+  let limitResources = process.env.LIMIT_SNAPSHOT_RESOURCES || false;
+  const MAX_RESOURCES = 750 - 2;
```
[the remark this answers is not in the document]
amandeepsingh333: this was to indicate here max resources limit in api is 750 but i noticed 2 extra resources, html file and percy log file
is also added 
should i change it to 748

packages/core/src/discovery.js
```diff
@@ -133,15 +133,22 @@ function parseDomResources({ url, domSnapshot }) {
   if (!domSnapshot) return map;
   let allRootResources = new Set();
   let allResources = new Set();
+  let limitResources = process.env.LIMIT_SNAPSHOT_RESOURCES || false;
+  const MAX_RESOURCES = 748; // 750 limit - 2 additional resources
```
Amit3200, line 137: Make in env variable

packages/core/src/discovery.js
```diff
@@ -451,6 +451,12 @@ export function createDiscoveryQueue(percy) {
                 return resource;
               },
               saveResource: r => {
+                const limitResources = process.env.LIMIT_SNAPSHOT_RESOURCES || false;
```
pankaj443, line 454: why its default false?

amandeepsingh333: we will be enabling this functionality for CS as of now so they will be passing these 2 env variables - so kept as false

pankaj443 (approved):

PR #1904, opened 2025-04-18
added logging for snapshot capture

Shivanshu-07:
Added logging to check the requests that are send and the resources that are saved.

### DIFF
--- a/packages/core/src/discovery.js
+++ b/packages/core/src/discovery.js
@@ -180,7 +180,12 @@ function createAndApplyPercyCSS({ percyCSS, roots }) {
 function processSnapshotResources({ domSnapshot, resources, ...snapshot }) {
   let log = logger('core:snapshot');
   resources = [...(resources?.values() ?? [])];
-
+  // Log all resources for the snapshot
+  log.debug(`Processing resources for snapshot: ${snapshot.name}`, snapshot.meta);
+  log.debug(`Total resources: ${resources.length}`);
+  resources.forEach(resource => {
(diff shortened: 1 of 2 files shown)


packages/core/src/discovery.js
```diff
@@ -180,7 +180,12 @@ function createAndApplyPercyCSS({ percyCSS, roots }) {
 function processSnapshotResources({ domSnapshot, resources, ...snapshot }) {
   let log = logger('core:snapshot');
   resources = [...(resources?.values() ?? [])];
-
+  // Log all resources for the snapshot
+  log.debug(`Processing resources for snapshot: ${snapshot.name}`, snapshot.meta);
+  log.debug(`Total resources: ${resources.length}`);
+  resources.forEach(resource => {
+    log.debug(`Resource URL for ${snapshot.name}: ${resource.url}, Size: ${resource.content?.length || 0}, MIME: ${resource.mimetype}`);
+  });
```
ninadbstack, line 188: meta missing

packages/core/src/network.js
```diff
@@ -467,6 +468,7 @@ async function saveResponseResource(network, request, session) {
     url,
     responseStatus: response?.status
   };
+  log.debug(`Response resource url: ${url}`, meta);
```
ninadbstack, line 471: why is this needed ? isnt it already part of meta ?

ninadbstack:
This wouldnt help debug storybook first load right ? at that point is this network attached and used ?

PR #1911, opened 2025-05-06
✨ Snapshot Level Browser Enablement

this-is-shivamsingh:
### Context:
- We are adding a new feature where user can passing browsers at snapshot level. Then triggered snapshot will be rendered on all given browsers
- Allowed Browsers: [chrome, firefox, safari, edge, safari_on_iphone, chrome_on_android]
- These browser are relevant with Percy and should be given in this format only
- The feature is only applicable for percy-web projects (not Automate)

### Supported with:
- `percySnapshot()`
- `npx percy snapshot <snapshot_file(json/yml)>`
- `.percy.yml` under snapshot

#### Note
- Not supported with `npx percy upload `<> command

### Changes
- Added Config validation on config for `browsers` params
    - This will be an array of string
    - Only applicable to web base projects

#### Testing
- Ran a percySnapshot command with and without browsers params
- Ran `npx percy snapshot ` command with and with browsers params
- Ran with `.percy.yml` commands


prklm10 (approved):
Just a question what will happen when use passes incorrect browsers. it will fail to create snapshot right?
Can we add that test as well.

PR #1919, opened 2025-05-28
Fix SVG element serialization and prevent network requests to serialized paths

bs-shobhitkumar:
Added `node.setAttribute('data-percy-serialized-attribute-xlink:href', cache.get(base64String));` in `serialize-base64.js `to prevent browsers from making network requests to original base64 URLs when scanning websites. This addresses an issue causing high server loads and 5xx errors on target sites like browserstack.com by removing the original attribute after setting the serialized version, similar to how we already handle 'src' attributes.

Updated tests to verify proper SVG handling.

### DIFF
--- a/packages/dom/src/serialize-base64.js
+++ b/packages/dom/src/serialize-base64.js
@@ -39,7 +39,8 @@ export function serializeBase64(node, resources, cache) {
       node.removeAttribute('xlink:href');
       node.setAttribute('data-percy-serialized-attribute-xlink:href', cache.get(base64String));
     } else {
-      node.href.baseVal = cache.get(base64String);
+      node.removeAttribute('href');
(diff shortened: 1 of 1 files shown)


packages/dom/src/serialize-base64.js
```diff
@@ -35,7 +35,12 @@ export function serializeBase64(node, resources, cache) {
   }
 
   if (isHrefUsed === true) {
-    node.href.baseVal = cache.get(base64String);
+    if (node.hasAttribute('xlink:href')) {
+      node.removeAttribute('xlink:href');
+      node.setAttribute('data-percy-serialized-attribute-xlink:href', cache.get(base64String));
+    } else {
+      node.href.baseVal = cache.get(base64String);
```
ninadbstack, line 42: this is not correct right ? it would cause the network call

PR #1959, opened 2025-08-15
chore: Typos

pieterocp:
Spotted one in the readme, so had a quick gander for a few more. Generally looking great though.

The only one which has me a bit concerned is the change to `packages/core/src/api.js` but as far as I can tell this should be fine?


ninadbstack:
@pieterocp Thanks for your contributions in fixing typos! I have started tests on it and will get it merged as it pass!

PR #1967, opened 2025-08-21
SDK Version Check

ayushatstack:
Ticket - https://browserstack.atlassian.net/browse/PPLT-4314
Screenshot 2025-08-19 at 8 30 24 PM

### DIFF
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -1,5 +1,5 @@
 import EventEmitter from 'events';
-import { sha256hash, request } from '@percy/client/utils';
+import { sha256hash } from '@percy/client/utils';
 import { camelcase, merge } from '@percy/config/utils';
 import YAML from 'yaml';
 import path from 'path';
@@ -630,6 +630,7 @@ export async function checkSDKVersion(clientInfo) {
     }
 
     // Fetch latest version from GitHub releases
+    const { request } = await import('@percy/client/utils');
     const githubData = await request(`https://api.github.com/repos/percy/${repoName}/releases`, {
       headers: { 'User-Agent': '@percy/cli' },
       retries: 0
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -584,3 +584,27 @@ export async function* maybeScrollToBottom(page, discovery) {
     yield page.eval('await scrollToBottom()');
   }
 }
+
+// Utility function to check SDK version updates
+export async function checkSDKVersion(clientInfo) {
+  const log = logger('core:sdk-version');
+  
+  try {
+    const parts = clientInfo.split('/');
+    const packageName = `${parts[0]}/${parts[1]}`;
+    const currentVersion = parts[2];
+    
+    // Fetch latest version from npm registry
+    const { request } = await import('@percy/client/utils');
+    const npmData = await request(`https://registry.npmjs.org/${packageName}`, { retries: 0 });
+    const latestVersion = npmData['dist-tags'].latest;
+    
+    log.debug(`[SDK Version Check] Current: ${currentVersion}, Latest: ${latestVersion}`);
+    
+    if (currentVersion !== latestVersion) {
+      log.warn(`[SDK Update Available] ${packageName}: ${currentVersion} -> ${latestVersion}`);
+    }
+  } catch (error) {
+    log.debug('Could not check SDK version');
+  }
+}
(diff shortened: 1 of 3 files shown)


packages/core/src/utils.js
```diff
@@ -584,3 +584,72 @@ export async function* maybeScrollToBottom(page, discovery) {
...
+      log.debug(`No repo mapping found for package: ${packageName}`);
+      return;
+    }
+
+    // Fetch latest version from GitHub releases
+    const { request } = await import('@percy/client/utils');
+    const githubData = await request(`https://api.github.com/repos/percy/${repoName}/releases`, {
+      headers: { 'User-Agent': '@percy/cli' },
+      retries: 0
+    });
```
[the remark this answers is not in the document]
ayushatstack: I dont think we need it for this feature since its a just a nudge for the user not the developer

packages/core/src/utils.js
```diff
@@ -584,3 +584,72 @@ export async function* maybeScrollToBottom(page, discovery) {
...
+    if (!latestRelease) {
+      log.debug('No stable release found');
+      return;
+    }
+
+    const latestVersion = latestRelease.tag_name.replace(/^v/, ''); // Remove 'v' prefix if present
+
+    log.debug(`[SDK Version Check] Current: ${currentVersion}, Latest: ${latestVersion}`);
+
+    if (currentVersion !== latestVersion) {
```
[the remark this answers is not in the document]
ayushatstack: We dont need less than or more than here just if it is equal to the latest version or not

packages/core/src/utils.js
```diff
@@ -584,3 +584,71 @@ export async function* maybeScrollToBottom(page, discovery) {
...
+    if (!repoName) {
+      log.debug(`No repo mapping found for package: ${packageName}`);
+      return;
+    }
+
+    // Fetch latest version from GitHub releases
+    const githubData = await request(`https://api.github.com/repos/percy/${repoName}/releases`, {
+      headers: { 'User-Agent': '@percy/cli' },
+      retries: 0
+    });
```
prklm10, line 636:
```suggestion
    const githubData = await request(`https://api.github.com/repos/percy/${repoName}/releases?page=1`, {
      headers: { 'User-Agent': '@percy/cli' },
      retries: 0
    });
```
By default it only returns first page data. But since we are using it, we should not rely on third party defaults.

packages/core/src/utils.js
```diff
@@ -584,3 +584,71 @@ export async function* maybeScrollToBottom(page, discovery) {
...
+
+    const latestRelease = githubData.find(r => !r.prerelease);
+    if (!latestRelease) {
+      log.debug('No stable release found');
```
prklm10, line 640: Is there a need to add this log?
Customer is using some kind of stable version right? We should not log this as it will contradict.

packages/core/src/utils.js
ayushatstack: made the changes

packages/core/src/utils.js
```diff
@@ -584,3 +584,70 @@ export async function* maybeScrollToBottom(page, discovery) {
...
+  '@percy/nightwatch': 'percy-nightwatch',
+  'percy-java-selenium': 'percy-selenium-java',
+  'percy-playwright-java': 'percy-playwright-java',
+  'percy-appium-dotnet': 'percy-appium-dotnet',
+  'percy-selenium-dotnet': 'percy-selenium-dotnet',
+  'percy-playwright-dotnet': 'percy-playwright-dotnet',
+  'percy-playwright-python': 'percy-playwright-python',
+  'percy-selenium-python': 'percy-selenium-python',
+  'percy-selenium-ruby': 'percy-selenium-ruby',
+  'percy-capybara': 'percy-capybara'
```
prklm10, line 607: i don't think this is all  appium java, appium ruby etc are missing.

packages/core/src/utils.js
```diff
@@ -584,3 +584,70 @@ export async function* maybeScrollToBottom(page, discovery) {
...
+      return;
+    }
+
+    const packageName = clientInfo.substring(0, lastSlashIndex);
+    const currentVersion = clientInfo.substring(lastSlashIndex + 1);
+
+    // Get GitHub repo name from mapping
+    const repoName = PACKAGE_TO_REPO[packageName];
+    if (!repoName) {
+      log.debug(`No repo mapping found for package: ${packageName}`);
```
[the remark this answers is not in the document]
ayushatstack: @prklm10  I kept it for debugging purpose sometimes the user-agent can come as empty or there might be a new newer sdk. Will it be a issue if it is shown only in debug logs? Normally the customer won't see it

PR #1970, opened 2025-08-26
Capture user machine specifications PPLT-4317

RakshitMandloi-bs:
added option to capture exact cpu model name and available disk space in the logs for the user machine


pankaj443 (approved):
LGTM

PR #1975, opened 2025-09-05
Fix Serializing CSS Stylesheet Issue - A11Y

pranavz28:
Before:
`styleSheetFromNode()` built its scratch document using `document.cloneNode()`.
This produced a head-less document where appended <style> tags never created a CSSStyleSheet. As a result, the helper often returned null, and later code attempted to access null.cssRules, causing TypeError exceptions on some sites (especially with CSS-in-JS or Safari/Firefox).

After:
`styleSheetFromNode()` now uses `document.implementation.createHTMLDocument()` to create a fully-formed scratch document, appending cloned <style> nodes into its <head>. This reliably initializes CSSOM across browsers. Additional null and SecurityError guards were added around .cssRules access, so unreadable or missing sheets are skipped instead of crashing.

Impact:
- The old `cloneNode()` method worked on some browsers but failed on others because CSSOM initialization requires a real `<head>` context.
- Without this fix, snapshotting could fail entirely on certain sites, making Percy unable to capture consistent DOM states.
- Eliminates Cannot read properties of null (reading 'cssRules') errors.
- Improves robustness of stylesheet serialization across Chrome, Firefox, and Safari.
- Safer handling of CSS-in-JS and CSP-restricted styles.

### DIFF
--- a/packages/dom/src/serialize-cssom.js
+++ b/packages/dom/src/serialize-cssom.js
@@ -7,15 +7,31 @@ function isCSSOM(styleSheet) {
   return !styleSheet.href && styleSheet.cssRules && styleSheet.ownerNode;
 }
 
+function safeRulesLen(sheet) {
+  if (!sheet) return null;
+  try {
+    return sheet.cssRules ? sheet.cssRules.length : 0;
+  } catch {
+    return null;
+  }
+}
+
 // Returns false if any stylesheet rules do not match between two stylesheets
 function styleSheetsMatch(sheetA, sheetB) {
-  for (let i = 0; i < sheetA.cssRules.length; i++) {
-    let ruleA = sheetA.cssRules[i].cssText;
-    let ruleB = sheetB.cssRules[i]?.cssText;
-    if (ruleA !== ruleB) return false;
+  const lenA = safeRulesLen(sheetA);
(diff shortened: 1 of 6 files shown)


packages/dom/src/serialize-cssom.js
```diff
@@ -9,13 +9,21 @@ function isCSSOM(styleSheet) {
...
+    lenB = sheetB?.cssRules?.length;
 
-  return true;
+    if (lenA !== lenB) return false;
+    for (let i = 0; i < lenA; i++) {
+      let ruleA = sheetA.cssRules[i].cssText;
+      let ruleB = sheetB.cssRules[i]?.cssText;
+      if (ruleA !== ruleB) return false;
+    }
+    return true;
```
prklm10, line 23: use early exit we should not rely of try catch. it makes our code slower.

prklm10: try to avoid it as much as possible.

pranavz28: I have removed the try catch blocks and added if statements for the checks.

Shivanshu-07 (commented):
Can you please also add how and why the previous `cloneNode()` approach was working? Add it in desc.

PR #1995, opened 2025-09-24
[PER-5758] fix canvas-to-image width handling

pranavz28:
This pull request updates the way maxWidth is handled when serializing canvas elements to images, ensuring that only canvases with an explicit maxWidth style have that style transferred to the resulting image element. The changes also add thorough test coverage for this behavior.

**Canvas serialization improvements:**

* Updated `createAndInsertImageElement` in `serialize-canvas.js` to only set `img.style.maxWidth` if the original canvas had a `maxWidth` style, rather than always defaulting to `100%`.

**Test enhancements:**

* Added a new canvas with a `max-width` style to the test DOM in `serialize-canvas.test.js` to verify correct style transfer.
* Added logic to draw on the new canvas and cache its data URL for testing.
* Updated existing tests to expect no default `max-width: 100%` style on images unless set on the canvas.
* Added a new test to ensure that a canvas with a `max-width` style results in an image with the same `max-width` style.When a <canvas> element has a CSS width exceeding that of its parent element, the correct behaviour is for the canvas to overflow its parent and be fully visible.

When Percy serialises the <canvas> to an image, it incorrectly resizes that image to fit inside the parent element.

For example, if we have:
```
<div style="width: 200px">
  <canvas style="width: 300px" />
</div>
```

The browser will render the canvas as 300px wide, but Percy will render the canvas as 200px wide.
Detailed issue description: https://github.com/12joan/percy-canvas-repro  

We are doing img.style.maxWidth = img.style.maxWidth || 100%;
We need to check if the canvas max width exists or not. If exists, keep the img max width as canvas width. No need to explicitly mention 100% maxwidth.
```
if (canvas.style.maxWidth) {
    img.style.maxWidth = canvas.style.maxWidth;
  }
```

### DIFF
--- a/packages/dom/test/serialize-canvas.test.js
+++ b/packages/dom/test/serialize-canvas.test.js
@@ -15,6 +15,12 @@ describe('serializeCanvas', () => {
           style="border: 5px solid black;"
         ></canvas>
         <canvas
+          id="canvas-with-maxwidth"
+          width="200px"
+          height="100px"
+          style="border: 2px solid red; max-width: 150px;"
+        ></canvas>
+        <canvas
           id="empty"
           width="0px"
           height="0px"
@@ -37,6 +43,14 @@ describe('serializeCanvas', () => {
         ctx.stroke();
 
         cache[plat].dataURL = canvas.toDataURL();
+
+        // Draw on canvas with maxWidth
+        let canvasWithMaxWidth = dom.getElementById('canvas-with-maxwidth');
+        let ctxWithMaxWidth = canvasWithMaxWidth.getContext('2d');
+        ctxWithMaxWidth.fillStyle = 'blue';
+        ctxWithMaxWidth.fillRect(10, 10, 50, 30);
+
+        cache[plat].dataURLWithMaxWidth = canvasWithMaxWidth.toDataURL();
       });
 
       serialized = serializeDOM();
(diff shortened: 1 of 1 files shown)


shantanuk-browserstack (approved):

packages/dom/src/serialize-canvas.js
```diff
@@ -13,8 +13,10 @@ function createAndInsertImageElement(canvas, clone, percyElementId, imageUrl) {
   img.setAttribute('data-percy-canvas-serialized', '');
   img.setAttribute('data-percy-serialized-attribute-src', imageUrl);
 
-  // set a default max width to account for canvases that might resize with JS
-  img.style.maxWidth = img.style.maxWidth || '100%';
+  // set the maxWidth only if it was set on the canvas
+  if (canvas.style.maxWidth) {
+    img.style.maxWidth = canvas.style.maxWidth;
+  }
```
12joan, line 19: @pranavz28 Thanks for fixing this issue!

By the way, this conditional assignment is redundant, since the full `style` attribute is already copied from the canvas onto the image by [lines 8-10](https://github.com/percy/cli/blob/b9748538e0c46842a502953badeb013d2ea8f04b/packages/dom/src/serialize-canvas.js#L8-L10) above.

PR #1999, opened 2025-09-30
feat: Add opacity serialization for DOM elements

RakshitMandloi-bs:
- Add serializeOpacity function to capture opacity values during DOM serialization
- Only captures opacity for elements with explicit styles or non-default values
- Add comprehensive test coverage for all opacity scenarios
- Preserve existing functionality with no regressions
- Support for both plain and shadow DOM

Resolves: PPLT-4320


packages/dom/src/transform-dom.js
```diff
@@ -19,10 +19,33 @@ export function serializeScrollState(original, clone) {
   }
 }
 
+export function serializeOpacity(original, clone) {
+  if (!original || !clone) return;
+
+  try {
+    // Get computed styles for the original element
+    const styles = window.getComputedStyle(original);
```
rahulXbrowserstack, line 27: do we need to get all styles here

RakshitMandloi-bs: original.style.opacity - Only gets inline styles (styles set directly on the element via style attribute)
window.getComputedStyle(original).opacity - Gets the final computed opacity after all CSS rules are applied (inline + CSS classes + inherited styles)

PR #2095, opened 2026-01-16
fix: handle non-http(s) URL schemes in DOM serialization

pranavz28:
## Problem

Customers migrating from Confluence Server to Cloud with broken macros were unable to take snapshots with Percy. When Confluence pages contain broken macros, Confluence adds `chrome-error://chromewebdata/` URIs to the DOM. Percy's DOM serialization creates `__serialized__` resource URLs using the page's base URL (`document.URL`), and if that base URL uses a non-http(s) scheme, the generated resource URLs inherit that scheme (e.g., `chrome-error://chromewebdata/__serialized__/uid.png`). Percy's upload validation rejects non-http(s) schemes, causing snapshot upload failures.

**Jira Ticket:** [PER-6667](https://browserstack.atlassian.net/browse/PER-6667)

## Solution

Modified `rewriteLocalhostURL()` in `@percy/dom` to detect and rewrite URLs with non-http(s) schemes to `http://render.percy.local`. The function now:

1. Parses the incoming URL
2. Checks if the protocol is NOT `http:` or `https:`
3. If non-http(s) and have chrome-error protocol, reconstructs the URL as a random URL
4. Otherwise, applies the existing localhost rewrite logic

This ensures that serialized resources always use valid http(s) schemes, regardless of the page's base URL scheme.

## Changes

- **Modified:** `packages/dom/src/utils.js`
  - Updated `rewriteLocalhostURL()` to handle non-http(s) schemes
  - Preserves pathname, search parameters, and hash fragments when rewriting
  
- **Modified:** `packages/dom/test/utils.test.js`
  - Added test cases for `chrome-error://` schemes
  - Added test cases verifying pathname/search/hash preservation

## Testing

- ✅ All existing tests pass (213 tests)
- ✅ Linting passes
- ✅ New test coverage for non-http(s) scheme handling
- ✅ Verified locally with reproduction case

## Impact

- **Affected Products:** Percy Visual Testing (all SDKs that use `@percy/dom`)
- **Deployment:** No special deployment order required
- **Breaking Changes:** None
- **Backwards Compatible:** Yes - only affects edge cases with non-http(s) base URLs

[PER-6667]: https://browserstack.atlassian.net/browse/PER-6667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dom/src/utils.js
+++ b/packages/dom/src/utils.js
@@ -45,12 +45,12 @@ export function styleSheetFromNode(node) {
 }
 
 export function rewriteLocalhostURL(url) {
-  // check if URL has non-http(s) scheme and rewrite to render.percy.local
   let parsedURL = new URL(url);
-  if (parsedURL.protocol !== 'http:' && parsedURL.protocol !== 'https:') {
-    url = 'http://render.percy.local' + parsedURL.pathname + parsedURL.search + parsedURL.hash;
+
+  // check if URL has chrome-error scheme and rewrite to a non-existent URL that will 404
+  if (parsedURL.protocol === 'chrome-error:') {
+    url = 'http://we-got-a-chrome-error-url-handled-gracefully.com/' + parsedURL.host + parsedURL.pathname + parsedURL.search + parsedURL.hash;
   }
-  // else ->
   return url.replace(/(http[s]{0,1}:\/\/)(localhost|127.0.0.1)[:\d+]*/, '$1render.percy.local');
 }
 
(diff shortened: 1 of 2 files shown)


packages/dom/src/utils.js
```diff
@@ -45,6 +45,12 @@ export function styleSheetFromNode(node) {
 }
 
 export function rewriteLocalhostURL(url) {
+  let parsedURL = new URL(url);
+
+  // check if URL has chrome-error scheme and rewrite to a non-existent URL that will 404
+  if (parsedURL.protocol === 'chrome-error:') {
+    url = 'http://we-got-a-chrome-error-url-handled-gracefully.com/' + parsedURL.host + parsedURL.pathname + parsedURL.search + parsedURL.hash;
+  }
```
rishigupta1599: Instead of CLI we should bypass the check in API?, in case any future changes or request comes we will be able to support from API itself without any new release?

PR #2172, opened 2026-04-07
feat: Fully Loaded Snapshot Capture — Readiness Gate (PER-7348)

Shivanshu-07:
## Summary

Adds a pre-snapshot readiness gate that waits for the page to be fully loaded and stable before capturing DOM. This reduces false visual diffs caused by skeleton loaders, partially rendered SPAs, and async content.

**Approach:** V2 Capture-Level Readiness — for SDK-provided snapshots, the CLI discards the SDK's domSnapshot and re-captures from the live URL with readiness checks. **Zero SDK changes required.**

### `@percy/dom` (browser-side)
- **New `waitForReady(options)`** async function with 6 composable checks:
  - DOM stability (MutationObserver with layout-affecting filter)
  - Network idle (PerformanceObserver-based)
  - Font readiness (`document.fonts.ready`)
  - Image readiness (above-the-fold viewport check)
  - `readySelectors` — wait for elements to appear
  - `notPresentSelectors` — wait for skeleton loaders to disappear
- 3 named presets: `balanced` (default), `strict`, `fast`, `disabled`

### `@percy/core` (CLI-side)
- **`readiness.js`** — CLI orchestrator that calls `PercyDOM.waitForReady()` via `page.eval()`
- **`page.js`** — Readiness gate inserted between `insertPercyDom()` and `serialize()`. Config priority: per-snapshot > per-page > global (`Page._globalReadinessConfig`)
- **`discovery.js`** — V2 re-capture: when SDK sends `domSnapshot` + readiness enabled, CLI deletes `domSnapshot`, clears cached root resource, forces URL-based capture path
- **`config.js`** — `readiness` schema with presets and per-snapshot overrides
- **`api.js`** — Tags SDK snapshots with `_fromSDK` flag
- **`percy.js`** — Preserves `_fromSDK` through validation, sets `Page._globalReadinessConfig`

### Configuration
```yaml
# .percy.yml
version: 2
snapshot:
  readiness:
    preset: strict
    notPresentSelectors:
      - .skeleton
      - .loading-spinner
```

## POC Results

| Build | Type | Readiness | Result |
|-------|------|-----------|--------|
| [#648](https://percy.io/9560f98d/web/Figma_test_shivanshu-17c329c0/builds/48445550) | Storybook | disabled | Skeletons |
| [#649](https://percy.io/9560f98d/web/Figma_test_shivanshu-17c329c0/builds/48445625) | Storybook | strict | **Fully loaded** (3.6s wait) |
| [#652](https://percy.io/9560f98d/web/Figma_test_shivanshu-17c329c0/builds/48445785) | SDK sim | disabled | Skeletons |
| [#653](https://percy.io/9560f98d/web/Figma_test_shivanshu-17c329c0/builds/48445871) | SDK sim | V2 re-capture | **Fully loaded** (1.9s wait) |

## Test Plan

- [x] `@percy/dom` browser tests — 18 specs covering all 6 checks, presets, kill switch, timeout, mutation filtering
- [x] CLI integration test — URL-based + SDK simulation with readiness
- [x] Storybook POC — `percy storybook` with `.percy.yml` readiness config
- [ ] Existing CLI test suite (CI)

## Known Limitations

- **Post-interaction state:** V2 re-captures from fresh page load — can't reproduce test interactions (button clicks, form fills)
- **Auth-protected pages:** CLI's browser has no test session — configure `discovery.authorization` or `discovery.cookies`
- **Performance:** V2 adds ~2-5s per snapshot for re-navigation
- **Shadow DOM:** MutationObserver doesn't observe inside shadow roots

## Jira

[PER-7348](https://browserstack.atlassian.net/browse/PER-7348)

[PER-7348]: https://browserstack.atlassian.net/browse/PER-7348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/readiness-gate.md
```diff
@@ -0,0 +1,691 @@
+# Fully Loaded Snapshot Capture — Readiness Gate
+
+> **Status:** POC validated, PR open ([percy/cli#2172](https://github.com/percy/cli/pull/2172))
+> **Jira:** [PER-7348](https://browserstack.atlassian.net/browse/PER-7348)
+> **Confluence:** [Fully Loaded Snapshot Capture](https://browserstack.atlassian.net/wiki/spaces/PROD/pages/6068505301/Fully+Loaded+Snapshot+Capture)
```
rishigupta1599: Remove this file

packages/dom/src/readiness.js
```diff
@@ -0,0 +1,331 @@
...
+    }
+    let total = document.querySelectorAll('img').length;
+    let incStart = getIncomplete().length;
+    if (incStart === 0) { resolve({ passed: true, duration_ms: 0, images_checked: total, images_incomplete_at_start: 0 }); return; }
+    let interval = setInterval(() => {
+      if (getIncomplete().length === 0) {
+        clearInterval(interval);
+        resolve({ passed: true, duration_ms: Math.round(performance.now() - start), images_checked: total, images_incomplete_at_start: incStart });
+      }
+    }, 100);
```
amandeepsingh333, line 187: 🔴 **HIGH — Memory/timer leak**: This `setInterval` (and the ones in `checkReadySelectors` at line 196, `checkNotPresentSelectors` at line 207, and `checkNetworkIdle` at line 132) is never cleared when the overall `waitForReady` timeout fires.

When `Promise.race` in `waitForReady` resolves via the timeout branch, these promises are simply abandoned — but their intervals keep firing indefinitely, wasting CPU and potentially throwing errors if the page navigates away.

**Fix:** Pass an `AbortSignal` (or shared `cancelled` flag) into each check function, and clear intervals on abort:
```js
function checkImageReady(config, signal) {
  return new Promise(resolve => {
    const interval = setInterval(() => {
      if (signal?.aborted) { clearInterval(interval); resolve({ passed: false, aborted: true }); return; }
      // existing logic...
    }, 100);
  });
}
```
Then in `waitForReady`, create an `AbortController`, pass `controller.signal` to all checks, and call `controller.abort()` in the timeout branch.

packages/dom/src/readiness.js
```diff
@@ -0,0 +1,331 @@
...
+    requestAnimationFrame(() => {
+      observing = true;
+      idleTimer = setTimeout(confirmIdle, idleWindowMs);
+    });
+  });
+}
+
+// --- Orchestrator ---
+
+async function runAllChecks(config, result) {
```
amandeepsingh333: 🔴 **HIGH — PerformanceObserver + rAF leak in `checkJSIdle`**: If the overall `waitForReady` timeout fires before JS becomes idle, this function is abandoned by `Promise.race` but:
- The `PerformanceObserver` is never disconnected (`cleanup()` is never called)
- The `requestAnimationFrame` and `requestIdleCallback` callbacks can fire after the page context is gone

Same abort mechanism fix applies here — disconnect the observer and cancel callbacks when the parent aborts.

packages/dom/src/readiness.js
```diff
@@ -0,0 +1,331 @@
...
+  let result = { passed: false, timed_out: false, preset: presetName, checks: {} };
+  let settled = false;
+
+  try {
+    await Promise.race([
+      runAllChecks(config, result).then(() => { settled = true; }),
+      new Promise(resolve => setTimeout(() => { if (!settled) result.timed_out = true; resolve(); }, effectiveTimeout))
+    ]);
+  } catch (error) {
+    /* istanbul ignore next: safety net for unexpected errors in readiness checks */
```
amandeepsingh333, line 372: 🔴 **HIGH — Timeout race: partial `result.checks` can silently pass**: When `Promise.race` resolves via timeout, `result.checks` only contains checks that completed before the timeout fired. Checks still running are simply missing from the object.

The problem: `Object.values(result.checks).every(c => c.passed)` returns `true` on an empty object (vacuous truth). So if ALL checks are still running at timeout, `result.checks` is `{}` and `every()` returns `true`.

`result.passed` is saved by `timed_out` being `true` on line 332, but this is fragile. **After timeout, explicitly mark any check that hasn't reported as `{ passed: false, timed_out: true }`** so the checks object is always complete.

packages/core/src/discovery.js
```diff
@@ -431,6 +431,23 @@ export function createDiscoveryQueue(percy) {
...
+          let extractedCookies = snapshot.domSnapshot?.cookies || snapshot.domSnapshot?.[0]?.cookies;
+          if (extractedCookies) snapshot._extractedCookies = extractedCookies;
+          delete snapshot.domSnapshot;
+          // Clear cached root HTML so the browser fetches the live page
+          for (let [key, val] of snapshot.resources) {
+            if (Array.isArray(val) && val[0]?.root) snapshot.resources.delete(key);
+          }
+        }
+
         // expectation explained in tests
```
amandeepsingh333, line 439: 🟡 **MEDIUM — V2 re-capture silently enables JavaScript**: Deleting `domSnapshot` here changes the semantics of downstream code. On line 457, `enableJavaScript` is computed as `!snapshot.domSnapshot` — which now becomes `true` even if the SDK originally sent `enableJavaScript: false`.

This means re-capture always runs with JS enabled, which could change screenshot behavior (animations, hover states, JS-injected content).

**Fix:** Preserve the original `enableJavaScript` value before deleting `domSnapshot`:
```js
const originalEnableJS = snapshot.enableJavaScript;
delete snapshot.domSnapshot;
snapshot.enableJavaScript = originalEnableJS ?? snapshot.enableJavaScript;
```

packages/core/src/percy.js
```diff
@@ -96,6 +97,11 @@ export class Percy {
     this.config = config;
     this.cliStartTime = null;
```
amandeepsingh333, line 99: 🟡 **MEDIUM — Static mutable state not safe for concurrent instances**: `Page._globalReadinessConfig` is a class-level static. If multiple `Percy` instances are created (tests, programmatic usage), the last constructor wins. When any instance calls `stop()` (line 377: `Page._globalReadinessConfig = null`), it nullifies the config for all instances.

**Fix:** Attach readiness config to the Percy instance and pass it through the call chain rather than using a class-level static.

packages/core/src/discovery.js
```diff
@@ -431,6 +431,23 @@ export function createDiscoveryQueue(percy) {
...
+        // V2 Readiness: For SDK-provided snapshots (_fromSDK), discard domSnapshot
+        // and re-capture from the live URL so the readiness gate can run.
+        // Only triggers when readiness is explicitly configured (not by default).
+        let readinessConfig = snapshot.readiness || percy.config?.snapshot?.readiness;
+        let readinessPreset = readinessConfig?.preset;
+        /* istanbul ignore next: V2 re-capture requires SDK snapshot + readiness config + live browser */
+        if (snapshot._fromSDK && snapshot.domSnapshot && snapshot.url && readinessConfig && readinessPreset !== 'disabled') {
+          percy.log.debug('Readiness enabled (SDK snapshot): re-capturing from URL', snapshot.meta);
+          let extractedCookies = snapshot.domSnapshot?.cookies || snapshot.domSnapshot?.[0]?.cookies;
+          if (extractedCookies) snapshot._extractedCookies = extractedCookies;
```
amandeepsingh333: 🟡 **MEDIUM — Readiness config resolved in 3+ places with different fallback chains**:
1. Here (line 443): `snapshot.readiness || percy.config?.snapshot?.readiness`
2. Line 500: `page._readinessConfig = snapshot.readiness || percy.config?.snapshot?.readiness`
3. In `page.js` line 220: `snapshot.readiness || this._readinessConfig || Page._globalReadinessConfig`

The third includes `Page._globalReadinessConfig` while the others don't. This means the V2 re-capture gate (which uses resolution #1) might decide readiness is disabled, while the actual readiness execution (resolution #3) sees it as enabled, or vice versa.

**Fix:** Resolve once early, attach to the snapshot object, and use that resolved value everywhere.

packages/core/src/api.js
```diff
@@ -141,7 +141,7 @@ export function createPercyServer(percy, port) {
     .route('post', '/percy/snapshot', async (req, res) => {
       let data;
       const snapshotPromise = {};
-      const snapshot = percy.snapshot(req.body, snapshotPromise);
+      const snapshot = percy.snapshot({ ...req.body, _fromSDK: true }, snapshotPromise);
```
amandeepsingh333: 🟡 **MEDIUM — ALL SDK snapshots get `_fromSDK: true` unconditionally**: This means when a user enables readiness globally in `.percy.yml`, every SDK snapshot with a `domSnapshot` will trigger V2 re-capture (domSnapshot deleted, fresh URL fetch). This is a significant behavior change — the SDK's pre-serialized DOM is discarded and replaced with a fresh page load.

For users with auth-protected pages, post-interaction states, or SDKs that rely on their own serialization, this will break silently.

Consider: (a) requiring an explicit `readiness.v2Recapture: true` config to opt in, or (b) warning in logs when re-capture discards an SDK snapshot, or (c) allowing SDKs to opt out with a flag like `_skipRecapture`.

packages/core/src/percy.js
```diff
@@ -448,8 +455,12 @@ export class Percy {
         : { url: options };
     }
 
+    // preserve internal flags stripped by validation
+    let fromSDK = options._fromSDK;
+
```
amandeepsingh333: 🟡 **MEDIUM — `_fromSDK` save/restore pattern is fragile**: Saving the flag before validation and re-attaching after is brittle. Any future change to `validateSnapshotOptions` that clones the options object, or any intermediate processing step, will silently drop this flag.

Consider either: (a) having `validateSnapshotOptions` preserve underscore-prefixed internal props, or (b) passing `_fromSDK` as a separate parameter alongside options rather than mixing it into the options object.

packages/dom/src/readiness.js
```diff
@@ -0,0 +1,331 @@
...
+/* istanbul ignore next: network idle polling is browser-timing dependent */
+function checkNetworkIdle(networkIdleWindowMs) {
+  return new Promise(resolve => {
+    let startTime = performance.now();
+    let lastCount = performance.getEntriesByType('resource').length;
+    let timer = null;
+    let interval = setInterval(() => {
+      let count = performance.getEntriesByType('resource').length;
+      /* istanbul ignore next: timer is always set before interval fires */
+      if (count !== lastCount) { lastCount = count; if (timer) clearTimeout(timer); timer = setTimeout(settle, networkIdleWindowMs); }
```
amandeepsingh333, line 141: 🟡 **MEDIUM — `performance.getEntriesByType('resource')` buffer overflow**: This polls `.length` every 50ms. The performance buffer has a default limit (~250 entries for resource timing). Once full, new resource loads aren't recorded, causing a false "network idle" signal on pages with many resources.

Consider using a `PerformanceObserver` with `type: 'resource'` instead, which receives entries as they happen without buffer limits and avoids creating a full array every 50ms.

packages/core/src/page.js
```diff
@@ -213,6 +217,14 @@ export class Page {
 
     await this.insertPercyDom();
 
+    // Run readiness checks before capturing — only when explicitly configured.
+    // Config priority: per-snapshot > per-page > global (from Percy constructor)
+    let readinessDiagnostics = null;
```
amandeepsingh333, line 222: 🟡 **MEDIUM — Percy DOM injection may trigger false DOM instability**: `insertPercyDom()` runs on line 216, injecting a script tag. Then `waitForReadiness` starts here, which includes a `checkDOMStability` MutationObserver. The script injection can trigger layout mutations that are counted as instability, adding false delay.

`checkJSIdle` has a rAF skip for this ("skip first frame to avoid detecting Percy's own setup"), but `checkDOMStability` has no equivalent protection. Consider either: (a) adding a similar initial-frame skip to `checkDOMStability`, or (b) running `waitForReadiness` before `insertPercyDom`.

packages/core/src/readiness.js
```diff
@@ -0,0 +1,89 @@
...
+  if (config.preset === 'disabled') return null;
+
+  log.debug(`Running readiness checks: preset=${config.preset}, not_present=${JSON.stringify(config.not_present_selectors || [])}`);
+
+  await page.insertPercyDom();
+
+  let result;
+  try {
+    /* istanbul ignore next: no instrumenting injected code */
+    /* istanbul ignore next: no instrumenting injected code */
```
amandeepsingh333, line 48: 🟡 **MEDIUM — Double `insertPercyDom` call**: This calls `await page.insertPercyDom()`, but the caller in `page.js` (line 216) already called `insertPercyDom()` before invoking `waitForReadiness`. If `insertPercyDom` is idempotent this is just wasteful; if it's not, it could cause issues. Consider removing this redundant call since the caller handles it.

amandeepsingh333 (commented):
## Code Review: Fully Loaded Snapshot Capture — Readiness Gate

Overall this is a well-structured feature with a solid approach (composable checks, named presets, V2 re-capture). The PR description and POC results are excellent. However, there are several issues that should be addressed before merge — most critically around resource cleanup and the V2 re-capture semantics.

### Key Concerns

| Severity | Count | Summary |
|----------|-------|---------|
| 🔴 High | 3 | Timer/observer leaks on timeout, race condition in check results |
| 🟡 Medium | 8 | V2 JS semantics change, static mutable state, config resolution inconsistency, test gaps |
| 🔵 Low | 4 | Duplicated presets, hardcoded font timeout, missing positive test |

### Top 3 Actionable Items
1. **Add cleanup/abort mechanism** for all readiness checks so timers and observers are cleared when the overall timeout fires
2. **Preserve original `enableJavaScript` value** when V2 re-capture deletes `domSnapshot`
3. **Add a positive integration test** for the V2 re-capture path

packages/dom/src/readiness.js
```diff
@@ -0,0 +1,331 @@
...
+  if (config.stability_window_ms > 0) checks.push(checkDOMStability(config.stability_window_ms).then(r => { result.checks.dom_stability = r; }));
+  if (config.network_idle_window_ms > 0) checks.push(checkNetworkIdle(config.network_idle_window_ms).then(r => { result.checks.network_idle = r; }));
+  if (config.font_ready !== false) checks.push(checkFontReady().then(r => { result.checks.font_ready = r; }));
+  if (config.image_ready !== false) checks.push(checkImageReady().then(r => { result.checks.image_ready = r; }));
+  if (config.js_idle !== false) checks.push(checkJSIdle(config.js_idle_window_ms || config.stability_window_ms).then(r => { result.checks.js_idle = r; }));
+  if (config.ready_selectors?.length) checks.push(checkReadySelectors(config.ready_selectors).then(r => { result.checks.ready_selectors = r; }));
+  if (config.not_present_selectors?.length) checks.push(checkNotPresentSelectors(config.not_present_selectors).then(r => { result.checks.not_present_selectors = r; }));
+  await Promise.all(checks);
+}
+
```
amandeepsingh333, line 345: 🔵 **LOW — `js_idle_window_ms` is undocumented dead code**: This reads `config.js_idle_window_ms` which is never defined in any preset, the config schema, or `resolveReadinessConfig`. It always falls through to `stability_window_ms`. Either add it to the schema/presets or remove the fallback.

packages/core/src/readiness.js
```diff
@@ -0,0 +1,89 @@
...
+};
+
+// Resolve readiness config from preset + per-snapshot overrides.
+// Accepts camelCase (from Percy config normalizer) and outputs snake_case for @percy/dom.
+/* istanbul ignore next: default params and ?? branches tested via unit tests */
+export function resolveReadinessConfig(options = {}) {
+  let readiness = options.readiness || {};
+  let presetName = readiness.preset || 'balanced';
+  if (presetName === 'disabled') return { preset: 'disabled' };
+
```
amandeepsingh333, line 18: 🔵 **LOW — PRESETS duplicated**: Both `packages/core/src/readiness.js` and `packages/dom/src/readiness.js` define identical PRESETS objects. If a preset value is updated in one but not the other, behavior will diverge. Consider a single source of truth — either export from one and import in the other, or have the CLI always pass fully-resolved config so the DOM module doesn't need presets.

packages/core/test/discovery.test.js
```diff
@@ -5020,4 +5022,73 @@ describe('Discovery', () => {
...
+    it('does not re-capture for non-SDK snapshots', async () => {
+      // Stop the outer percy instance before starting a new one with readiness config
+      await percy?.stop(true);
+
+      percy = await Percy.start({
+        token: 'PERCY_TOKEN',
+        snapshot: {
+          widths: [1000],
+          // Use disabled preset to avoid running readiness gate in page.snapshot()
+          readiness: { preset: 'disabled' }
```
amandeepsingh333, line 5103: 🔵 **LOW — Missing positive test for V2 re-capture**: Tests cover "does not re-capture when readiness is disabled" and "does not re-capture for non-SDK snapshots", but there's no test verifying the happy path — `_fromSDK: true` + readiness enabled asserts that `domSnapshot` was deleted and the page was fetched from URL. The positive path should have test coverage.

packages/core/test/percy.test.js
```diff
@@ -2211,4 +2211,47 @@ describe('Percy', () => {
...
+
+    it('preserves _fromSDK flag through snapshot validation', async () => {
+      await percy.start();
+
+      spyOn(percy, 'snapshot').and.callFake(async () => {});
+
+      await percy.snapshot({ url: server.address, name: 'test', domSnapshot: '<html></html>', _fromSDK: true });
+
+      expect(percy.snapshot).toHaveBeenCalledWith(
+        jasmine.objectContaining({ _fromSDK: true })
```
amandeepsingh333: 🔵 **LOW — Test provides false confidence**: This test spies on `percy.snapshot` with `callFake(async () => {})`, then asserts the spy was called with `_fromSDK: true`. But the spy replaces the real implementation — the flag preservation logic (save before validation, restore after) never actually runs. This test only verifies the test's own input, not that `_fromSDK` survives through `validateSnapshotOptions`.

**Fix:** Don't spy on `snapshot` — instead spy on a downstream method (e.g., `#discover`) to verify the flag reaches the actual consumer.

docs/readiness-gate.md
Shivanshu-07: Done — file was already removed in a subsequent commit (1d1b8fd5).

packages/dom/src/readiness.js
Shivanshu-07: Fixed in 80d28e12. Added an abort mechanism: `waitForReady` creates an abort handle, passes it to all checks, and calls `abort()` when the timeout fires. Each check registers cleanup callbacks via `aborted.onAbort()` that clear intervals, disconnect observers, and cancel timers. No more leaked intervals/observers on timeout.

packages/dom/src/readiness.js
Shivanshu-07: Fixed in same commit (80d28e12). `checkJSIdle` now checks `aborted.value` at every callback entry point and registers `aborted.onAbort(() => cleanup())` to disconnect the PerformanceObserver and clear the idle timer when the parent timeout fires.

packages/dom/src/readiness.js
Shivanshu-07: Fixed in 80d28e12. `runAllChecks` now tracks `result._expectedChecks` (list of check names that were started). After timeout, `waitForReady` iterates expected checks and marks any missing ones as `{ passed: false, timed_out: true }`. The checks object is always complete before `every()` runs.

packages/core/src/discovery.js
Shivanshu-07: Fixed in 80d28e12. Now explicitly preserving the original `enableJavaScript` value before deleting `domSnapshot`:
```js
snapshot.enableJavaScript = snapshot.enableJavaScript ?? false;
delete snapshot.domSnapshot;
```
This ensures the downstream `!snapshot.domSnapshot` computation doesn't silently flip JS to enabled.

packages/core/src/percy.js
Shivanshu-07: Valid concern. In practice, Percy CLI creates only one instance per process (the server binds to a single port). Multiple instances are only used in tests, where `afterEach` resets the static. That said, the reviewer is right that this is a code smell. I'll track this as a follow-up refactor to pass readiness config through the instance chain instead of using a class static. For now, the test-level cleanup is sufficient and the risk is low in production.

packages/core/src/discovery.js
Shivanshu-07: Good catch — the three resolution points could diverge. The design intent is:
1. **discovery.js (V2 gate)**: decides whether to re-capture. Uses snapshot + global config.
2. **discovery.js (page._readinessConfig)**: sets per-page config for Storybook path.
3. **page.js**: the actual readiness execution, with full fallback chain.

The `Page._globalReadinessConfig` in #3 is specifically for Storybook, which doesn't pass readiness in snapshot options — it relies on the global. Consolidating to a single resolution point would require threading the resolved config through the Storybook capture path, which is a larger refactor. Will track as follow-up.

packages/core/src/api.js
Shivanshu-07: By design — V2 re-capture only triggers when ALL of these are true: `_fromSDK && domSnapshot && url && readinessConfig && preset !== 'disabled'`. The key gate is `readinessConfig` — it only triggers when the user has **explicitly configured** readiness in `.percy.yml`. Without that config, behavior is unchanged.

For auth-protected pages: documented as a known limitation — users should configure `discovery.authorization` or `discovery.cookies`.
For post-interaction states: per-snapshot `readiness: { preset: 'disabled' }` bypasses re-capture for specific snapshots.

The doc at docs/readiness-gate.md covers these trade-offs in the Limitations section.

packages/core/src/percy.js
Shivanshu-07: Agree it's not ideal. The pattern exists because `validateSnapshotOptions` uses JSON schema validation which strips unknown properties. Making the validator preserve underscore-prefixed props is the cleaner fix. Will track as a follow-up to pass `_fromSDK` as a separate parameter alongside options.

packages/dom/src/readiness.js
Shivanshu-07: Valid point. The performance buffer limit (default ~150 for resource timing) could cause false idle signals on resource-heavy pages. However, `checkNetworkIdle` in the browser is a secondary check — the primary network idle is handled by CDP events in `core/src/network.js` (which has no buffer limit). The browser-side check is a complementary signal, not the sole source. Will track upgrading to PerformanceObserver({type: 'resource'}) as a follow-up improvement.

packages/core/src/page.js
Shivanshu-07: Good observation. `insertPercyDom()` runs before `waitForReadiness()`, and the script injection causes a childList mutation. However, `checkDOMStability` starts with a fresh timer — it waits for `stabilityWindowMs` of silence from the moment the observer is created. By the time the readiness gate starts observing, the PercyDOM injection is already complete. The rAF skip in `checkJSIdle` provides additional protection. If we observe issues in practice, adding an initial-frame skip to `checkDOMStability` is straightforward.

packages/core/src/readiness.js
Shivanshu-07: Fixed in 80d28e12. Removed the redundant `insertPercyDom()` call from `core/readiness.js`. The caller (`page.snapshot()` line 218) already calls it before `waitForReadiness()`, and `insertPercyDom()` is idempotent (checks `!window.PercyDOM` before injecting), but the redundant call was unnecessary.

packages/dom/src/readiness.js
Shivanshu-07: Fixed in 80d28e12. Removed the `config.js_idle_window_ms ||` fallback — now uses `config.stability_window_ms` directly since `js_idle_window_ms` was never in the config schema or presets.

packages/core/src/readiness.js
Shivanshu-07: Acknowledged. The duplication exists because `@percy/dom` runs in the browser context (injected via eval) and can't import from `@percy/core` (Node.js). The CLI always passes fully-resolved config via `resolveReadinessConfig()`, so the DOM presets are only used as fallbacks when `waitForReady()` is called directly (e.g., in browser tests or if PercyDOM is used standalone). Will consider having the CLI always pass complete config to eliminate the need for DOM-side presets.

packages/core/test/discovery.test.js
Shivanshu-07: Fair point. A positive test for V2 re-capture (asserting domSnapshot was deleted and page navigated to URL) would need a full browser + server + readiness config integration test, which tends to be slow and flaky in CI. The negative tests cover the guard conditions. Will add a targeted positive test in a follow-up that verifies the domSnapshot deletion without requiring full browser capture.

packages/core/test/percy.test.js
Shivanshu-07: Valid — the spy replaces the real implementation so the save/restore logic never executes. The test only verifies that the API passes `_fromSDK`. The actual flag preservation through `validateSnapshotOptions` is covered by the api.test.js tests which use the real endpoint. Will improve this test in a follow-up to spy on a downstream consumer instead.

packages/core/src/discovery.js
```diff
@@ -359,6 +359,11 @@ async function* captureSnapshotResources(page, snapshot, options) {
     yield waitForFontLoading(page);
     yield waitForDiscoveryNetworkIdle(page, discovery);
     yield* captureResponsiveAssets();
+    // Readiness does NOT run on SDK-provided domSnapshots. The DOM is already
+    // serialized at this point — running readiness cannot modify it. Preserving
+    // the domSnapshot maintains modal state, post-interaction state, and supports
+    // localhost testing. SDK users who need fully-loaded snapshots should wait
+    // for page stability in their test code before calling percySnapshot().
```
rishigupta1599, line 366: unnecessary comment

packages/core/src/discovery.js
```diff
@@ -483,10 +488,29 @@ export function createDiscoveryQueue(percy) {
             }
           });
 
+          // Set readiness config on page for Storybook (which calls page.snapshot() without readiness in options)
+          page._readinessConfig = snapshot.readiness || percy.config?.snapshot?.readiness;
```
rishigupta1599, line 492: why we need separate parsing, check how other config options are merged and loaded.

packages/core/src/discovery.js
```diff
@@ -483,10 +488,29 @@ export function createDiscoveryQueue(percy) {
...
+                  .filter(([, c]) => !c.passed)
+                  .map(([name]) => name);
+                percy.suggestionsForFix(
+                  `Snapshot "${captured.name}" readiness timed out after ${diag.total_duration_ms}ms. Failed checks: ${failedChecks.join(', ')}`,
+                  { snapshotLevel: true, snapshotName: captured.name }
+                );
+              }
+              callback(captured);
+            };
+
```
rishigupta1599, line 510: Check how `warnings` object are logged follow same rather than changing here

packages/core/src/page.js
```diff
@@ -11,6 +12,9 @@ import {
 
 export class Page {
   static TIMEOUT = undefined;
+  // Global readiness config set by Percy constructor, accessible by all page instances.
+  // Needed for Storybook which calls page.snapshot() without readiness in options.
+  static _globalReadinessConfig = null;
```
rishigupta1599, line 17: why separate variable needed percy.config should have the config.

packages/core/src/readiness.js
```diff
@@ -0,0 +1,90 @@
+import logger from '@percy/logger';
+
+const log = logger('core:readiness');
+
```
rishigupta1599, line 4: Do we need separate file, why can't we reuse dom.js readiness file only which exposes function?

rishigupta1599: Also if dom.js is handling internally it should not be even required as snapshot and sdk both flow execute same dom capture script only

PR #2177, opened 2026-04-11
PER 7292

aryanku-dev:
## Summary

Adds CLI-side capture for DOM features Percy's serializer previously missed:

- **Closed shadow DOM** — discovered via CDP (`DOM.getDocument({ pierce: true })`) and exposed to `PercyDOM.serialize()` through a per-realm `window.__percyClosedShadowRoots` WeakMap. Works for shadow hosts nested inside iframes; respects `snapshot.disableShadowDOM`.
- **Custom-element `:state()` selectors** — captures `ElementInternals.states`; rewrites modern `:state(X)` and legacy `:--X` to attribute selectors that survive serialization.
- **Interactive pseudo-class states** — `:focus` (captured before clone), `:checked`, `:disabled`, `:hover`, `:active`. Differential CSS rules extracted via CSSOM walk so the serialized DOM renders the state without JS.
- **Iframe sandbox / cross-origin warnings** — fidelity warnings when sandbox attributes prevent capture. Adds `snapshot.ignoreIframeSelectors` and a per-iframe depth cap.
- **Custom-elements upgrade wait** — pre-snapshot best-effort wait on `:not(:defined)` so lazy-defined components hydrate before serialization. 1500 ms internal cap (no user-facing knob — `waitForCustomElementsTimeout` was removed after PR review).

## Files changed

| Area | Files |
|---|---|
| Closed shadow capture (CLI) | `packages/core/src/closed-shadow.js`, `page.js`, `config.js`, tests |
| DOM serializer | `packages/dom/src/{clone-dom,prepare-dom,serialize-dom,serialize-frames,serialize-pseudo-classes,serialize-custom-states,shadow-utils,utils}.js`, tests |
| Regression fixtures | `test/regression/pages/{dom-structures,hydration,interactive-states,sandbox-iframes,shadow-dom}.html`, `snapshots.yml` |

## Test plan

- [ ] `@percy/core` unit suite passes locally and in CI
- [ ] `@percy/dom` Karma suite passes in Chrome
- [ ] Windows CI green (archive flow regexes/`path.resolve` fixes)
- [ ] Coverage report checked in CI

## Notes for reviewers

- **`MAX_SHADOW_DEPTH` parity**: `packages/core/src/closed-shadow.js`'s `MAX_SHADOW_DEPTH = 10` must stay in lockstep with `HARD_MAX_IFRAME_DEPTH` in `@percy/dom`'s `serialize-frames`. Comment in source flags this.
- **`waitForCustomElementsTimeout`** was initially exposed as a snapshot option and removed after PR review — there was no compelling case for per-snapshot tuning. Existing user configs that set it will now fail schema validation; surface this in the changelog.
- **`ignoreIframeSelectors.items.minLength: 1`** rejects empty-string selector entries. Same changelog note applies.
- **Archive snapshots + `--replay` command** are bundled in this branch via commit `543eb805` (#2216). Functionally independent from the closed-shadow/serializer work — flagging in case you want to mentally split scope during review.
- **Local coverage gotcha**: `nyc` only collects coverage on Node 14 here because the codebase's ESM `transformSource` loader hook was removed in Node 18+. CI uses Node 14, so this only matters for local verification.


rishigupta1599 (commented):
Test

packages/core/src/page.js
```diff
@@ -9,6 +10,21 @@ import {
...
+// Cached preflight script for closed shadow root and ElementInternals interception
+let _preflightScript = null;
+async function getPreflightScript() {
+  if (!_preflightScript) {
+    let pkgRoot = path.resolve(path.dirname(PERCY_DOM), '..');
+    let preflightPath = path.join(pkgRoot, 'src', 'preflight.js');
+    try {
+      _preflightScript = await fs.promises.readFile(preflightPath, 'utf-8');
+    } catch {
+      _preflightScript = ''; // graceful fallback if file not found
```
rishigupta1599: This will not work for consumers of `@percy/core` once published. `PERCY_DOM` resolves to `node_modules/@percy/dom/dist/bundle.js` at runtime, and `@percy/dom`'s `package.json` has `"files": ["dist"]` — so `../src/preflight.js` does not exist in the installed package. The `catch` silently swallows ENOENT and sets `_preflightScript = ''`, meaning closed shadow root capture and ElementInternals interception **will never run for any SDK consumer**. The only reason it works in this PR's tests is that the monorepo layout happens to put `packages/dom/src/preflight.js` one directory up from `dist/bundle.js`.

Fix options: (a) inline the preflight source into `@percy/dom`'s bundle and export it as a string (e.g. `PercyDOM.PREFLIGHT_SCRIPT`), (b) copy `preflight.js` into `packages/dom/dist/` as part of the build and reference it there, or (c) bundle it into `@percy/core/dist` directly. Whichever you pick, please add a test that asserts `getPreflightScript()` returns a non-empty string when the resolved bundle lives in a typical `node_modules` layout so this regression can't reoccur.

packages/core/src/page.js
```diff
@@ -247,6 +278,13 @@ export class Page {
...
           autoAttach: true,
           flatten: true
+        }),
+        // inject preflight script to intercept closed shadow roots and ElementInternals
+        // before any page scripts run
+        getPreflightScript().then(script => {
+          if (script) {
+            return session.send('Page.addScriptToEvaluateOnNewDocument', { source: script });
+          }
         }));
```
rishigupta1599, line 331: Ordering hazard: `Page.addScriptToEvaluateOnNewDocument` requires `Page.enable` to have landed first, but both are dispatched concurrently inside `Promise.all(commands)`. Depending on how the CDP transport flushes messages, the preflight `addScriptToEvaluateOnNewDocument` can be rejected ("Page domain is not enabled") or applied after the first navigation has already started — which is the window we're trying to cover. Please `await session.send('Page.enable')` before pushing the preflight send, or chain the preflight onto that specific command rather than running it in parallel.

rishigupta1599 (commented):
Thanks for the thorough work on DOM fidelity (closed shadow roots, sandboxed iframes, :state() rewriting, interactive states). The test coverage additions are substantial. However, a few issues should be addressed before this lands. The biggest ones:

1. **Preflight script is not shipped to consumers.** `@percy/dom` publishes only `dist/`, but `page.js` reads the script from `src/preflight.js` relative to the resolved `PERCY_DOM` bundle. Works in the monorepo; silently falls back to an empty string for every installed dependency — so closed-shadow-root / ElementInternals capture will not function in production. It needs to be bundled or copied into `dist/` (and added to `files` in `packages/dom/package.json`).
2. **`Page.addScriptToEvaluateOnNewDocument` is dispatched in the same `Promise.all` as `Page.enable`.** No ordering guarantee — the preflight injection can race with, or land after, the initial navigation.
3. **"Potentially inaccessible shadow root" detection counts every custom element without a shadow-host marker.** Most custom elements have no shadow DOM at all; this produces false positives in both the fidelity warning and the `fidelityRegions` payload.
4. **Sandboxed-iframe fidelity regions are emitted before we know whether the frame was captured.** Successfully serialized sandboxed srcdoc frames still emit an overlay region.

A few smaller nits inline. Nothing blocking from a test-coverage perspective — suite is very thorough — but the production-path issues above should be fixed before merge.

(Apologies: an earlier stray "Test" review came through while I was iterating on this payload — please ignore it.)

packages/core/src/page.js
```diff
@@ -247,6 +278,13 @@ export class Page {
           waitForDebuggerOnStart: false,
           autoAttach: true,
           flatten: true
+        }),
+        // inject preflight script to intercept closed shadow roots and ElementInternals
+        // before any page scripts run
+        getPreflightScript().then(script => {
+          if (script) {
+            return session.send('Page.addScriptToEvaluateOnNewDocument', { source: script });
```
rishigupta1599: The returned promise here is swallowed by the outer `.catch(session._handleClosedError)` on line 292, which is fine for session-close races. Any other `addScriptToEvaluateOnNewDocument` failure (oversized source, invalid JS, etc.) is also consumed by `_handleClosedError` though. Consider logging a debug/warning when preflight injection fails for non-close reasons — otherwise we can't tell from field logs whether capture is actually instrumented.

packages/dom/src/serialize-dom.js
```diff
@@ -100,24 +102,56 @@ export function serializeDOM(options) {
...
   ctx.clone = cloneNodeAndShadow(ctx);
 
   serializeElements(ctx);
 
-  // STEP 4: Process pseudo-class enabled elements
+  // Detect potentially inaccessible shadow roots
+  let inaccessibleShadowCount = 0;
+  for (let origEl of ctx.dom.querySelectorAll('*')) {
+    if (!origEl.tagName?.includes('-')) continue;
+    if (origEl.hasAttribute('data-percy-shadow-host')) continue;
```
rishigupta1599: This flags **every** custom element that isn't a shadow host as "potentially inaccessible shadow". In practice most custom elements don't use shadow DOM at all (light-DOM components, simple wrappers, icon elements, etc.). Consequences: (a) `inaccessibleShadowCount` is inflated and the fidelity warning becomes noisy/wrong, and (b) a `fidelityRegions` entry is pushed per custom element, so the API/UI renders overlays on elements that were fully captured.

Suggestion: only flag when there's positive evidence of an inaccessible shadow — e.g. preflight records which elements *called* `attachShadow` with `mode: 'closed'` before WeakMap lookup succeeded. Absent that signal, drop the counter/region and only warn when WeakMap tracking is explicitly bypassed.

packages/dom/src/serialize-frames.js
```diff
@@ -42,12 +42,67 @@ function setBaseURI(dom, warnings) {
...
+
+    // Warn about sandboxed iframes
+    if (sandboxAttr !== null) {
+      sandboxWarned++;
+      let frameLabel = frame.id || frame.src || percyElementId;
+      let tokens = sandboxAttr.split(/\s+/).filter(Boolean);
+
+      if (tokens.length === 0) {
+        warnings.add(`Sandboxed iframe "${frameLabel}" has no permissions — content may not render with full fidelity in Percy`);
+        captureFidelityRegion(frame, 'sandboxed-restricted', fidelityRegions);
```
rishigupta1599: Fidelity region is pushed before we know the sandboxed frame's fate. A srcdoc iframe with `sandbox` but no `allow-scripts` can still be serializable via `frame.contentDocument`, so it gets both (a) fully captured into the snapshot AND (b) an overlay region in `fidelityRegions`. Consider deferring `captureFidelityRegion` for sandbox-restricted frames until after the `contentDocument` branch, only pushing if the frame was *not* captured — or only emit the region in the `else`/`!enableJavaScript && builtWithJs` branches below where we actually discard the iframe.

packages/dom/src/serialize-frames.js
```diff
@@ -42,12 +42,67 @@ function setBaseURI(dom, warnings) {
...
+      cloneEl?.remove();
+      continue;
+    }
     let builtWithJs = !frame.srcdoc && (!frame.src || frame.src.split(':')[0] === 'javascript');
+    let sandboxAttr = frame.getAttribute('sandbox');
+
+    // Warn about sandboxed iframes
+    if (sandboxAttr !== null) {
+      sandboxWarned++;
+      let frameLabel = frame.id || frame.src || percyElementId;
```
rishigupta1599: Nit: `frameLabel = frame.id || frame.src || percyElementId` — `percyElementId` is an internal token (`_abc123`) that's meaningless to users. Falling through produces a warning like `Sandboxed iframe "_ab12cd" has no permissions`, which reads like an error code. Prefer user-actionable labeling, e.g. `frame.src || frame.getAttribute('name') || '<unnamed iframe>'`.

packages/dom/src/preflight.js
```diff
@@ -0,0 +1,32 @@
...
+(function() {
+  if (window.__percyPreflightActive) return;
+  window.__percyPreflightActive = true;
+
+  // --- Intercept closed shadow roots ---
+  let closedShadowRoots = new WeakMap();
+  let origAttachShadow = window.Element.prototype.attachShadow;
+  window.Element.prototype.attachShadow = function(init) {
+    let root = origAttachShadow.call(this, init);
+    if (init && init.mode === 'closed') {
```
rishigupta1599: Minor robustness: if a page calls `attachShadow()` with no args (or `null`), `init.mode` throws a `TypeError` *inside our wrapper* — a slightly different failure mode than the original `attachShadow`. Guard with `init?.mode === 'closed'`.

packages/core/src/page.js
```diff
@@ -211,6 +227,21 @@ export class Page {
     // wait for any final network activity before capturing the dom snapshot
     await this.network.idle();
 
+    // wait for custom elements to be defined before capturing
+    /* istanbul ignore next: no instrumenting injected code */
+    await this.eval(function() {
```
rishigupta1599: This adds up to a **5-second hard wait** to every snapshot whenever the page has a single `:not(:defined)` element. Many production sites legitimately reference custom-element tags that never register (third-party widgets, blocked lazy loaders, typos). That means the common case eats +5s latency with no way to opt out. Please either (a) make the timeout configurable via the snapshot schema, (b) use a much shorter default (e.g. 500ms) since `customElements.whenDefined` tends to resolve synchronously for already-defined tags, or (c) resolve early if network is idle and the undefined tags have no pending loader.

This will regress P50 build latency for SDK consumers.

packages/core/src/discovery.js
```diff
@@ -190,6 +190,13 @@ function processSnapshotResources({ domSnapshot, resources, ...snapshot }) {
   let log = logger('core:snapshot');
   resources = [...(resources?.values() ?? [])];
 
+  // log fidelity warnings from dom serialization
+  let domWarnings = domSnapshot?.warnings?.filter(w => w.startsWith('[fidelity]')) || [];
+  for (let w of domWarnings) log.info(w);
+
+  // extract fidelity regions for API upload
+  let fidelityRegions = domSnapshot?.fidelityRegions || domSnapshot?.[0]?.fidelityRegions || [];
```
rishigupta1599: Reading `domSnapshot?.[0]?.fidelityRegions` suggests multi-root DOMs are expected, but only `domSnapshot[0]` is consulted — fidelity regions captured by subsequent roots are silently dropped. If responsive snapshot capture produces multiple domSnapshots (per-width), you'll be under-reporting. Either merge regions across all entries or add a comment explaining why only the first matters.

packages/dom/src/clone-dom.js
```diff
@@ -65,6 +69,23 @@ export function cloneNodeAndShadow(ctx) {
...
       let clone = cloneElementWithoutLifecycle(node);
 
+      // After cloning and before shadow DOM handling, detect custom states
+      let percyInternals = window.__percyInternals?.get(node);
+      if (percyInternals?.states?.size > 0) {
+        let states = [];
+        try {
+          for (let state of percyInternals.states) {
+            // CSS-escape the state value to prevent injection
+            states.push(state.replace(/["\\}\]]/g, '\\$&'));
```
rishigupta1599: The inline comment says "CSS-escape the state value to prevent injection" but the regex only escapes `"`, `\\`, `}`, `]` — not whitespace, `<`, `>`, `&`, or non-printables. Custom state names per spec are `<dashed-ident>`s so this shouldn't happen through legitimate API use; if the intent is defense-in-depth against a non-compliant runtime, either tighten the validation (`if (!/^[-\\w]+$/.test(state)) continue;`) or soften the comment to "escape attribute-breaking chars".

packages/core/src/page.js
aryanku-dev: Fixed. `getPreflightScript()` now tries multiple candidate paths (`src/preflight.js`, `dist/preflight.js`, and alongside the bundle) so it works both in the monorepo and when installed via npm.

packages/core/src/page.js
aryanku-dev: Fixed. Preflight `addScriptToEvaluateOnNewDocument` is now chained onto `Page.enable` via `.then()` instead of running in parallel, ensuring the Page domain is ready first.

packages/core/src/page.js
aryanku-dev: Fixed. Non-close preflight injection failures are now logged at debug level. Close/destroy errors are still silently handled by `_handleClosedError`.

packages/dom/src/serialize-dom.js
aryanku-dev: Fixed. Now only flags custom elements that are confirmed closed shadow hosts via `window.__percyClosedShadowRoots?.has(origEl)`, instead of flagging every custom element without a shadow host attribute.

packages/dom/src/serialize-frames.js
aryanku-dev: Fixed. Removed `captureFidelityRegion` calls from within the sandbox warning section. Sandboxed frames that aren't captured will still get a fidelity region from the `else` branch at the bottom (cross-origin-excluded).

packages/dom/src/serialize-frames.js
aryanku-dev: Fixed. Changed fallback to `frame.id || frame.src || frame.getAttribute('name') || '<unnamed iframe>'` instead of using the internal `percyElementId`.

packages/dom/src/preflight.js
aryanku-dev: Fixed. Changed to `init?.mode === 'closed'` and `origAttachShadow.apply(this, arguments)` for both null safety and forward compatibility.

packages/core/src/page.js
aryanku-dev: Fixed. Reduced timeout from 5000ms to 500ms. `customElements.whenDefined` resolves synchronously for already-defined tags, so 500ms is sufficient for the common case while avoiding the latency regression for pages with never-registered custom elements.

packages/core/src/discovery.js
aryanku-dev: Added a comment explaining that only the first domSnapshot's fidelity regions are used because for responsive captures with multiple widths, regions are width-independent (same DOM structure) so the first entry is representative.

packages/dom/src/clone-dom.js
aryanku-dev: Fixed. Replaced the partial regex escape with strict validation: `if (!/^[-\w]+$/.test(state)) continue;` to skip invalid state values per the `<dashed-ident>` spec requirement.

packages/client/src/client.js
```diff
@@ -452,6 +452,7 @@ export class PercyClient {
     regions,
     algorithm,
     algorithmConfiguration,
+    fidelityRegions,
```
rishigupta1599: change the name, will it be understandable for what purpose it is used?

packages/core/src/page.js
```diff
@@ -9,6 +10,32 @@ import {
   serializeFunction
 } from './utils.js';
 
+// Cached preflight script for closed shadow root and ElementInternals interception
+let _preflightScript = null;
+async function getPreflightScript() {
```
rishigupta1599: revisit but page object will not be available in SDK in most cases

packages/core/src/discovery.js
```diff
@@ -190,6 +190,10 @@ function processSnapshotResources({ domSnapshot, resources, ...snapshot }) {
   let log = logger('core:snapshot');
   resources = [...(resources?.values() ?? [])];
 
+  // log fidelity warnings from dom serialization
+  let domWarnings = domSnapshot?.warnings?.filter(w => w.startsWith('[fidelity]')) || [];
```
rishigupta1599: change the name, make it consistent

packages/core/src/page.js
```diff
@@ -9,6 +11,24 @@ import {
...
+// preflight.js sits next to this file in core (src in dev, dist after publish),
+// so a single relative resolve works in both layouts.
+let _preflightScript = null;
+async function getPreflightScript() {
+  if (_preflightScript !== null) return _preflightScript;
+  try {
+    let here = path.dirname(url.fileURLToPath(import.meta.url));
+    _preflightScript = await fs.promises.readFile(path.join(here, 'preflight.js'), 'utf-8');
+  } catch (err) {
+    /* istanbul ignore next: graceful fallback — closed-shadow capture degrades to no-op */
```
rishigupta1599: why so many istanbul ignore

packages/core/src/page.js
```diff
@@ -211,6 +231,21 @@ export class Page {
...
+      let undefinedEls = document.querySelectorAll(':not(:defined)');
+      if (!undefinedEls.length) return Promise.resolve();
+      return Promise.race([
+        Promise.all(
+          Array.from(undefinedEls).map(function(el) {
+            return window.customElements.whenDefined(el.localName);
+          })
+        ),
+        new Promise(function(r) { setTimeout(r, 500); })
+      ]);
```
rishigupta1599: don't think so this is needed because JS will be enable in user's browser and this happens when JS is disabled and not making it defined

packages/dom/src/serialize-custom-states.js
```diff
@@ -0,0 +1,189 @@
...
+export function rewriteCustomStateSelectors(text, stateNames) {
+  let out = '';
+  let i = 0;
+  const len = text.length;
+  while (i < len) {
+    const ch = text[i];
+    if (ch === '"' || ch === "'") {
+      const quote = ch;
+      out += ch; i++;
+      while (i < len && text[i] !== quote) {
```
rishigupta1599: nope can't go, we don't want token by token processing

packages/core/src/config.js
```diff
@@ -284,6 +284,19 @@ export const configSchema = {
...
         default: false
       },
+      ignoreIframeSelectors: {
+        type: 'array',
+        default: [],
+        items: {
+          type: 'string'
+        }
+      },
+      waitForCustomElementsTimeout: {
```
rishigupta1599: what's the usecase?, custome element individual timeout not required

packages/dom/src/prepare-dom.js
```diff
@@ -1,18 +1,27 @@
...
 export function uid() {
   return `_${Math.random().toString(36).substr(2, 9)}`;
 }
 
 export function markElement(domElement, disableShadowDOM, forceShadowAsLightDOM) {
   // Mark elements that are to be serialized later with a data attribute.
-  if (['input', 'textarea', 'select', 'iframe', 'canvas', 'video', 'style', 'dialog'].includes(domElement.tagName?.toLowerCase())) {
+  // Custom elements with ElementInternals or closed shadow roots also get
+  // stamped so the post-clone state-fallback can locate their clones.
+  let isCustomElement = domElement.tagName?.includes('-');
```
rishigupta1599: make it a function used at multiple places

packages/dom/src/serialize-pseudo-classes.js
```diff
@@ -1,12 +1,232 @@
...
+  let len = selectorText.length;
+  while (i < len) {
+    let ch = selectorText[i];
+    // Top-level string literal — copy verbatim through the closing quote.
+    if (ch === '"' || ch === "'") {
+      let quote = ch;
+      out += ch; i++;
+      while (i < len && selectorText[i] !== quote) {
+        if (selectorText[i] === '\\' && i + 1 < len) {
+          out += selectorText[i] + selectorText[i + 1];
```
rishigupta1599: same as above

rishigupta1599:
## Review — Confirmed Issues

### 1. Custom elements timeout is 1500ms, not 500ms as stated in review reply

`page.js:15` sets `DEFAULT_WAIT_FOR_CUSTOM_ELEMENTS_TIMEOUT = 1500` and the injected body hardcodes `arguments[0] || 1500`. The reply to my earlier review said "Reduced timeout from 5000ms to 500ms" but the code shows 1500ms. This adds **1.5s to every snapshot** on any page with a single never-registered custom element (common: third-party chat widgets, blocked lazy loaders, typos in tag names). Either reduce to 500ms as claimed, or justify 1500ms with data.

### 2. `clampIframeDepth` is duplicated

Identical function exists in both `packages/dom/src/serialize-frames.js` and `packages/sdk-utils/src/index.js`. One should import from the other — if the clamping logic ever changes (e.g., different default), these will silently diverge.

### 3. Performance: 6+ redundant shadow DOM walks per snapshot

During a single `serializeDOM` call, `walkShadowDOM` / `queryShadowAll` is invoked at least 6 times:

- `queryShadowAll(ctx.dom, ':checked')` — walk #1
- `queryShadowAll(ctx.dom, ':disabled')` — walk #2
- `collectStyleSheets(ctx.dom)` inside `extractPseudoClassRules` — walk #3
- `collectStyleElements(ctx.clone)` inside `rewriteCustomStateCSS` — walk #4
- `walkShadowDOM(ctx.clone, ...)` inside `addCustomStateAttributes` — walk #5
- `walkShadowDOM(ctx.dom, ...)` inside `addCustomStateAttributes` — walk #6

Each walk visits every `[data-percy-shadow-host]` element and recurses into its shadow root. On pages with many shadow hosts (Lit, Shoelace, Lightning Web Components — 200-500+ shadow hosts is common), this compounds to thousands of extra `querySelectorAll` calls per snapshot. Walks #3 and #4 cover overlapping ground (both iterate stylesheets in shadow roots). Consider:
- Merging the CSS walks into a single pass that collects both pseudo-class rules and custom-state styles
- Caching the shadow host list so repeated `querySelectorAll('[data-percy-shadow-host]')` calls don't re-traverse the tree
- Short-circuiting: if no `:checked`/`:disabled` elements exist, skip the CSS rule extraction for those pseudos

### 4. `markInteractiveStates` now mutates the live DOM unconditionally

`markPseudoClassElements` (line 720 of `serialize-pseudo-classes.js`) now always calls `markInteractiveStates(ctx)` before checking if `pseudoClassEnabledElements` config exists. This means **every snapshot** stamps `data-percy-checked`, `data-percy-disabled`, `data-percy-focus` on the live page — even when users haven't opted into pseudo-class capture.

The `cleanupInteractiveStateMarkers` in the `finally` block undoes this, which handles the normal path. But:
- If the process crashes or the tab is closed between mark and cleanup (SDK mode runs in the customer's browser tab), attributes leak into the live page
- This is a behavioral change from the previous version where no live-DOM mutation happened without explicit `pseudoClassEnabledElements` config — worth calling out in the changelog

Not blocking, but worth being aware of as a change in the contract with SDK consumers.

rishigupta1599 (approved):

PR #2184, opened 2026-04-14
feat: Readiness gate in PercyDOM.serialize() — works for URL + SDK paths (PER-7348)

Shivanshu-07:
## Summary

Implements readiness-gated snapshot capture for Percy (PER-7348). Readiness checks ensure pages are stable before DOM serialization — no skeleton screens, loaded fonts/images, idle network, settled JavaScript.

### Architecture: Two-Call Pattern

Readiness and serialization are **two separate calls**, not one combined async function:

```
1. PercyDOM.waitForReady(config)  — async, waits for page stability
2. PercyDOM.serialize(options)    — sync, captures the stable DOM
```

This keeps `serialize()` always synchronous (zero SDK breakage) while isolating the async readiness into its own independent step. Both the CLI URL-capture path and SDKs use the same pattern.

### What's included in this PR (CLI-side)

**`@percy/dom` — readiness checks (`packages/dom/src/readiness.js`)**
- 7 parallel checks: DOM stability (MutationObserver), network idle (PerformanceObserver), font ready, image ready, JS idle (Long Task API + rIC + double-rAF), ready selectors, not-present selectors
- 3 presets: `balanced` (default), `strict`, `fast`, `disabled`
- Dedicated `js_idle_window_ms` decoupled from `stability_window_ms`
- `normalizeOptions()` for camelCase/snake_case config compatibility
- Abort propagation to all checks on timeout
- 100% test coverage (lines/branches/functions/statements)

**`@percy/dom` — serialization (`packages/dom/src/serialize-dom.js`)**
- `serializeDOM()` — single sync function. No async variant needed.
- `waitForReady()` exported separately for the readiness call

**`@percy/core` — CLI URL-capture path (`packages/core/src/page.js`)**
- Two-call pattern: `waitForReady(config)` then `serialize(options)` as separate eval calls
- Readiness config from per-snapshot options or global `.percy.yml`
- Graceful degradation: if readiness fails or times out, serialize still runs

**`@percy/core` — config schema (`packages/core/src/config.js`)**
- `readiness` schema: `preset`, `stabilityWindowMs`, `jsIdleWindowMs`, `networkIdleWindowMs`, `timeoutMs`, `maxTimeoutMs`, `domStability`, `imageReady`, `fontReady`, `jsIdle`, `readySelectors`, `notPresentSelectors`
- `readySelectors` / `notPresentSelectors` accept either a CSS string or an explicit `{ css }` / `{ xpath }` object form. Bare strings beginning with `/`, `//`, `./`, `(/`, `(./` are auto-detected as XPath.
- `readiness_diagnostics` allowed in domSnapshot schema (no more validation warnings)
- CLI logs readiness diagnostics at debug/warn level

**`@percy/sdk-utils` — helpers for SDK adoption (`packages/sdk-utils/src/serialize-dom.js`)**
- `waitForReadyScript(config, { callback })` — JS code string for the readiness call
  - Default variant: for `page.evaluate()` (Puppeteer/Playwright auto-await)
  - Callback variant: for `executeAsyncScript` (Selenium/Nightwatch/WebdriverIO)
- `getReadinessConfig(options)` — extracts readiness config from per-snapshot or global `percy.config`
- `isReadinessDisabled(options)` — quick check for `preset: disabled`

---

## SDK Rollout (separate PRs per SDK)

After this CLI PR merges and a new CLI version is published, each SDK adds a `waitForReady()` call **before** its existing `serialize()` call. The `serialize()` call itself stays unchanged.

### How SDKs adopt readiness

**The two-call pattern for SDKs:**

```
Step 1: Call PercyDOM.waitForReady(config)   — async, page stabilizes
Step 2: Call PercyDOM.serialize(options)      — sync, UNCHANGED from today
```

### Pattern A: Puppeteer / Playwright SDKs (`page.evaluate` auto-awaits)

```js
// NEW — readiness (async, auto-awaited by page.evaluate)
let readinessConfig = utils.getReadinessConfig(options);
if (!utils.isReadinessDisabled(options)) {
  await page.evaluate(async (config) => {
    if (typeof PercyDOM?.waitForReady === 'function') {
      return PercyDOM.waitForReady(config);
    }
  }, readinessConfig).catch(() => {});
}

// UNCHANGED — serialize (sync)
let domSnapshot = await page.evaluate((opts) => PercyDOM.serialize(opts), serializeOptions);
```

**Applies to:** @percy/puppeteer, @percy/playwright, percy-playwright-python, percy-playwright-java, percy-playwright-dotnet

### Pattern B: Selenium SDKs (`executeAsyncScript` for readiness, `executeScript` for serialize)

```python
# NEW — readiness (async, uses executeAsyncScript with callback)
readiness_config = percy.config.get('snapshot', {}).get('readiness', {})
if readiness_config.get('preset') != 'disabled':
    try:
        driver.execute_async_script('''
            var config = arguments[0];
            var done = arguments[arguments.length - 1];
            if (typeof PercyDOM !== 'undefined' && typeof PercyDOM.waitForReady === 'function') {
                PercyDOM.waitForReady(config).then(function() { done(); }).catch(function() { done(); });
            } else { done(); }
        ''', readiness_config or {})
    except Exception:
        pass

# UNCHANGED — serialize (sync, exactly the same as today)
dom = driver.execute_script('return PercyDOM.serialize({})'.format(json.dumps(opts)))
```

**Applies to:** percy-selenium-python, percy-selenium-java, percy-selenium-ruby, percy-selenium-dotnet, @percy/selenium-js, percy-capybara, @percy/webdriverio

### Pattern C: In-browser SDKs (Cypress, Ember)

```js
// NEW — readiness (await the Promise directly)
if (typeof PercyDOM?.waitForReady === 'function') {
  await PercyDOM.waitForReady(readinessConfig).catch(() => {});
}

// UNCHANGED — serialize (sync)
let domSnapshot = PercyDOM.serialize(options);
```

**Applies to:** @percy/cypress, @percy/ember

### Pattern D: Other JS SDKs

- **@percy/nightwatch:** `browser.executeAsync()` for readiness, `browser.execute()` unchanged
- **@percy/testcafe:** `await t.eval(async () => PercyDOM.waitForReady(config))` before serialize
- **@percy/storybook:** Uses CLI `page.eval()` — already covered by this PR

### Backward compatibility

| Scenario | Behavior |
|----------|----------|
| **Old SDK + new CLI** | SDK never calls `waitForReady`. `serialize()` works as today. No readiness. |
| **New SDK + old CLI** | SDK checks `typeof PercyDOM.waitForReady === 'function'` — undefined on old CLI. Skips readiness. `serialize()` works. |
| **New SDK + new CLI** | `waitForReady()` runs then page stabilizes then `serialize()` captures stable DOM. |
| **waitForReady throws** | SDK catches error, proceeds to `serialize()`. Snapshot captured without readiness. |
| **waitForReady times out** | Resolves with `{ timed_out: true }`. SDK proceeds to `serialize()`. |

### SDK change matrix

| SDK | Repo | Change needed |
|-----|------|--------------|
| @percy/puppeteer | percy/percy-puppeteer | Pattern A |
| @percy/playwright | percy/percy-playwright | Pattern A |
| percy-playwright-python | percy/percy-playwright-python | Pattern A |
| percy-playwright-java | percy/percy-playwright-java | Pattern A |
| percy-playwright-dotnet | percy/percy-playwright-dotnet | Pattern A |
| @percy/cypress | percy/percy-cypress | Pattern C |
| @percy/ember | percy/percy-ember | Pattern C |
| @percy/nightwatch | percy/percy-nightwatch | Pattern D |
| @percy/testcafe | percy/percy-testcafe | Pattern D |
| @percy/selenium-js | percy/percy-selenium-js | Pattern B |
| @percy/webdriverio | percy/percy-webdriverio | Pattern B |
| percy-selenium-python | percy/percy-selenium-python | Pattern B |
| percy-selenium-java | percy/percy-selenium-java | Pattern B |
| percy-selenium-ruby | percy/percy-selenium-ruby | Pattern B |
| percy-selenium-dotnet | percy/percy-selenium-dotnet | Pattern B |
| percy-capybara | percy/percy-capybara | Pattern B |
| @percy/storybook | percy/percy-storybook | No change (uses CLI page.eval) |

---

## Test plan

- [x] `@percy/dom` tests: 616 pass, 100% coverage
- [x] `@percy/core` tests: 684 specs, no new failures (27 pre-existing env failures: Chromium install mocks, runDoctor, AggregateError)
- [x] `@percy/config` tests: 82 pass
- [x] Lint passes
- [x] `serializeDOM()` is always sync (backward compat test)
- [x] `waitForReady()` with disabled preset skips all checks
- [x] Readiness waits for DOM stability (skeleton removal test)
- [x] Readiness waits for ready selectors
- [x] Readiness times out gracefully (serialize still happens)
- [x] camelCase config normalization (SDK flow)
- [x] Direct unit tests for helpers: isLayoutMutation, hasLayoutStyleChange, parseStyleProps, normalizeOptions, createAbortHandle

## Configuration

```yaml
# .percy.yml
snapshot:
  readiness:
    preset: balanced  # balanced | strict | fast | disabled
    stabilityWindowMs: 300
    jsIdleWindowMs: 300
    networkIdleWindowMs: 200
    timeoutMs: 10000
    maxTimeoutMs: 25000  # cap for Selenium async timeout compat
    domStability: true   # set false to fully disable the MutationObserver check
    imageReady: true
    fontReady: true
    jsIdle: true
    readySelectors:
      - '.content-loaded'                    # CSS string
      - '//section[@id="ready" and contains(@class,"loaded")]'  # XPath (auto-detected)
      - { xpath: '//div[@id="root"]' }       # explicit XPath object form
      - { css: '.weird /selector' }          # explicit CSS object form
    notPresentSelectors: ['.skeleton']
```

### Kill switches

| Scenario | Setting |
|---|---|
| **Disable everything** (revert to pre-readiness behavior) | `preset: disabled` |
| **Keep gating on JS idle / images / fonts / selectors but skip the MutationObserver** (use this on heavy-DOM pages where the observer drives CPU/memory pressure) | `domStability: false` |
| **Skip individual checks** | `imageReady: false`, `fontReady: false`, `jsIdle: false` |

Generated with [Claude Code](https://claude.com/claude-code)


packages/dom/src/serialize-dom.js
```diff
@@ -79,8 +80,8 @@ export function waitForResize() {
   window.resizeCount = 0;
 }
 
-// Serializes a document and returns the resulting DOM string.
-export function serializeDOM(options) {
+// Internal sync serialization — called after readiness (if any) completes.
+function serializeDOMSync(options) {
```
rishigupta1599: why is this required?, if readiness flag is there call waitForReadiness with a try catch

packages/core/src/page.js
```diff
@@ -214,14 +214,23 @@ export class Page {
...
-      domSnapshot: PercyDOM.serialize(options),
-      url: document.URL
-    }), { enableJavaScript, disableShadowDOM, forceShadowAsLightDOM, domTransformation, reshuffleInvalidTags, ignoreCanvasSerializationErrors, ignoreStyleSheetSerializationErrors, pseudoClassEnabledElements });
+      let result = PercyDOM.serialize(options);
+      // serialize may return a Promise when readiness is configured
+      if (result && typeof result.then === 'function') {
+        return result.then(r => ({ domSnapshot: r, url: document.URL }));
+      }
+      return { domSnapshot: result, url: document.URL };
+    }, { enableJavaScript, disableShadowDOM, forceShadowAsLightDOM, domTransformation, reshuffleInvalidTags, ignoreCanvasSerializationErrors, ignoreStyleSheetSerializationErrors, pseudoClassEnabledElements, readiness });
```
rishigupta1599: should not be required why output of serialize is changing?

Shivanshu-07: The page.js change IS required — here's why:

`PercyDOM.serialize(options)` now returns a **Promise** when readiness is configured (it calls `waitForReady()` which is async). When no readiness is configured, it returns a plain object (sync, same as before).

Inside `page.eval()`, the injected function runs in the browser. If `serialize()` returns a Promise, that Promise ends up as `{ domSnapshot: Promise, url: ... }`. CDP's `awaitPromise: true` only awaits the **top-level** return value — it won't resolve a Promise nested inside an object.

So we need to check: if serialize returned a Promise, `.then()` it to get the final `{ domSnapshot, url }` shape as the top-level return. If it returned a plain object, wrap it directly. This ensures CDP always receives a resolvable value.

Without this handling, readiness-enabled snapshots would return `domSnapshot: [object Promise]` instead of the actual HTML.

packages/dom/src/serialize-dom.js
```diff
@@ -80,7 +81,37 @@ export function waitForResize() {
 }
 
 // Serializes a document and returns the resulting DOM string.
+// When readiness config is provided, waits for the page to be stable
+// before serializing. This works identically for both URL-based captures
+// (CLI percy snapshot, Storybook) and SDK captures (Cypress, Selenium,
+// Puppeteer) since serialize is the common entry point for both paths.
 export function serializeDOM(options) {
```
rishigupta1599, line 88: function was sync till now now we are making it async for readiness flow, does all SDK handle that. 
In current case if for any reason DOM serialization breaks user will not even get failure as there test suite will move forward

rishigupta1599: And SDK publish DOM Snapshot as is and will end up sending a Promise

packages/dom/src/readiness.js
```diff
@@ -0,0 +1,392 @@
...
+  await Promise.all(checks);
+}
+
+export async function waitForReady(options = {}) {
+  let presetName = options.preset || 'balanced';
+  if (presetName === 'disabled') return { passed: true, timed_out: false, skipped: true, checks: {} };
+
+  let preset = PRESETS[presetName] || PRESETS.balanced;
+  let config = { ...preset, ...options };
+  let effectiveTimeout = config.max_timeout_ms ? Math.min(config.timeout_ms, config.max_timeout_ms) : config.timeout_ms;
```
rishigupta1599, line 514: **Bug: camelCase / snake_case mismatch breaks user config overrides**

The config schema in `config.js` defines properties in camelCase (`stabilityWindowMs`, `networkIdleWindowMs`, `timeoutMs`), but `readiness.js` presets and this spread use snake_case (`stability_window_ms`, etc.).

When a user sets `.percy.yml`:
```yaml
snapshot:
  readiness:
    preset: strict
    stabilityWindowMs: 500
```

The spread `{ ...preset, ...options }` sets `stabilityWindowMs: 500` but the code reads `config.stability_window_ms` (from the preset default), silently ignoring the override.

You need either:
1. A normalization step to convert camelCase options to snake_case before the spread, or
2. Align the config schema to use snake_case (matching the existing `enable_javascript` / `dom_transformation` pattern in `serialize-dom.js`), or
3. Support both in the destructuring like `serialize-dom.js` does: `enableJavaScript = options?.enable_javascript`

This is likely to bite every user who configures readiness via `.percy.yml` or per-snapshot SDK options.

packages/dom/src/readiness.js
```diff
@@ -0,0 +1,392 @@
...
+
+async function runAllChecks(config, result, aborted) {
+  let checks = [];
+  let expected = [];
+  if (config.stability_window_ms > 0) { expected.push('dom_stability'); checks.push(checkDOMStability(config.stability_window_ms, aborted).then(r => { result.checks.dom_stability = r; })); }
+  if (config.network_idle_window_ms > 0) { expected.push('network_idle'); checks.push(checkNetworkIdle(config.network_idle_window_ms, aborted).then(r => { result.checks.network_idle = r; })); }
+  if (config.font_ready !== false) { expected.push('font_ready'); checks.push(checkFontReady().then(r => { result.checks.font_ready = r; })); }
+  if (config.image_ready !== false) { expected.push('image_ready'); checks.push(checkImageReady(aborted).then(r => { result.checks.image_ready = r; })); }
+  if (config.js_idle !== false) { expected.push('js_idle'); checks.push(checkJSIdle(config.stability_window_ms, aborted).then(r => { result.checks.js_idle = r; })); }
+  if (config.ready_selectors?.length) { expected.push('ready_selectors'); checks.push(checkReadySelectors(config.ready_selectors, aborted).then(r => { result.checks.ready_selectors = r; })); }
```
rishigupta1599, line 476: **Coupling: `checkJSIdle` reuses `stability_window_ms` instead of its own config**

```js
checks.push(checkJSIdle(config.stability_window_ms, aborted)...)
```

JS idle window and DOM stability window are conceptually different. With the `strict` preset (`stability_window_ms: 1000`), the JS idle check requires **1 full second** of no long tasks before confirming idle -- that's very aggressive and will cause unnecessary timeouts on pages with normal JS activity.

Consider adding a dedicated `js_idle_window_ms` config (defaulting to something like 200ms) or at minimum document why these are intentionally coupled.

packages/dom/src/readiness.js
```diff
@@ -0,0 +1,392 @@
...
+    }, 50);
+
+    function settle() { clearInterval(interval); resolve({ passed: true, duration_ms: Math.round(performance.now() - startTime) }); }
+    timer = setTimeout(settle, networkIdleWindowMs);
+
+    aborted.onAbort(() => { clearInterval(interval); if (timer) clearTimeout(timer); });
+  });
+}
+
+function checkFontReady() {
```
rishigupta1599: **Missing abort propagation in `checkFontReady`**

Unlike every other check function, `checkFontReady` doesn't accept or honor the `aborted` signal. It has a hardcoded 5s internal timeout that runs independently of the orchestrator's `effectiveTimeout`.

If a user sets `timeoutMs: 2000` and fonts take 3s, the overall timeout fires at 2s and aborts all other checks -- but the font check's internal `Promise.race` keeps the 5s timer alive. While it won't block (the orchestrator moves on), it leaves a dangling timer.

Accept `aborted` and clear the internal timeout on abort, consistent with the other checks.

packages/core/src/page.js
```diff
@@ -214,14 +214,23 @@ export class Page {
...
-    let capture = await this.eval((_, options) => ({
+    let capture = await this.eval((_, options) => {
       /* eslint-disable-next-line no-undef */
-      domSnapshot: PercyDOM.serialize(options),
-      url: document.URL
-    }), { enableJavaScript, disableShadowDOM, forceShadowAsLightDOM, domTransformation, reshuffleInvalidTags, ignoreCanvasSerializationErrors, ignoreStyleSheetSerializationErrors, pseudoClassEnabledElements });
+      let result = PercyDOM.serialize(options);
+      // serialize may return a Promise when readiness is configured
+      if (result && typeof result.then === 'function') {
+        return result.then(r => ({ domSnapshot: r, url: document.URL }));
```
rishigupta1599: **Simplification: the manual thenable check is redundant**

Since `eval()` already uses `awaitPromise: true` (line 143), CDP will automatically await any Promise returned from the function. You can simplify this to:

```js
let capture = await this.eval(async (_, options) => {
  let result = await PercyDOM.serialize(options);
  return { domSnapshot: result, url: document.URL };
}, { ...opts, readiness });
```

`await` on a non-thenable (sync path) just returns the value directly, so backward compat is preserved. This avoids the manual duck-typing of `.then`.

packages/dom/src/readiness.js
```diff
@@ -0,0 +1,392 @@
...
+    });
+  });
+}
+
+/* istanbul ignore next: network idle polling is browser-timing dependent */
+function checkNetworkIdle(networkIdleWindowMs, aborted) {
+  return new Promise(resolve => {
+    let startTime = performance.now();
+    let lastCount = performance.getEntriesByType('resource').length;
+    let timer = null;
```
rishigupta1599, line 202: **Performance: `performance.getEntriesByType('resource')` called every 50ms**

This allocates a new array of all resource entries on every tick. On resource-heavy pages (hundreds of images, scripts, stylesheets), this is expensive -- especially since you only need the count.

Consider using `PerformanceObserver` for `'resource'` entries (like you do for `longtask` in `checkJSIdle`) to incrementally track count changes, instead of polling the full entry list every 50ms.

packages/core/src/config.js
```diff
@@ -132,6 +132,21 @@ export const configSchema = {
       sync: {
         type: 'boolean'
       },
+      readiness: {
+        type: 'object',
+        additionalProperties: false,
+        properties: {
+          preset: { type: 'string', enum: ['balanced', 'strict', 'fast', 'disabled'] },
+          stabilityWindowMs: { type: 'integer', minimum: 50, maximum: 30000 },
```
rishigupta1599, line 143: **Naming convention inconsistency with the rest of the config schema**

The readiness properties use camelCase (`stabilityWindowMs`, `networkIdleWindowMs`, `timeoutMs`) but the internal readiness implementation uses snake_case, and other Percy config options like `enable_javascript`, `dom_transformation`, `reshuffle_invalid_tags` use snake_case in the schema.

Consider aligning to snake_case (`stability_window_ms`, `network_idle_window_ms`, `timeout_ms`) for consistency with the rest of the Percy config, or add explicit camelCase to snake_case mapping in `readiness.js`.

packages/dom/src/readiness.js
```diff
@@ -0,0 +1,392 @@
...
+    timeout_ms: 5000,
+    image_ready: false,
+    font_ready: true,
+    js_idle: true
+  }
+};
+
+const LAYOUT_ATTRIBUTES = new Set([
+  'class', 'width', 'height', 'display', 'visibility',
+  'position', 'src', 'href'
```
rishigupta1599: **Questionable: `href` in `LAYOUT_ATTRIBUTES`**

Changing `href` on an `<a>` tag doesn't affect layout. This is only layout-relevant for `<link>` stylesheet elements. As written, any JS that updates an anchor's `href` will be incorrectly classified as a layout mutation, resetting the stability timer.

Consider either removing `href` or scoping it: check `mutation.target.tagName === 'LINK'` before treating an `href` change as layout-affecting.

packages/dom/src/readiness.js
```diff
@@ -0,0 +1,392 @@
...
+]);
+
+const LAYOUT_STYLE_PROPS = /^(width|height|top|left|right|bottom|margin|padding|display|position|visibility|flex|grid|min-|max-|inset|gap|order|float|clear|overflow|z-index|columns)/;
+
+/* istanbul ignore next: branches constrained by MutationObserver attributeFilter config */
```
rishigupta1599: **Excessive `istanbul ignore` coverage suppression**

There are ~15 `/* istanbul ignore next */` annotations across this 392-line file, covering core logic paths (abort branches, network idle, image ready, JS idle, selector polling, style parsing, etc.).

This undermines the coverage story in the PR description. Many of these *are* testable -- e.g., abort branches can be triggered by setting very short timeouts, the style parsing can be tested with direct unit tests of `hasLayoutStyleChange` / `parseStyleProps` if exported.

Consider:
1. Exporting the internal helpers for direct unit testing
2. Using short timeouts to exercise abort/timeout paths
3. Reserving `istanbul ignore` for truly untestable browser APIs only

rishigupta1599 (commented):
## Review Summary

The architecture is solid -- moving readiness into `PercyDOM.serialize()` is the right call for URL + SDK parity. Good test coverage (1000+ lines) and clean abort/cleanup logic.

However, there are several issues that need addressing before merge, ranging from a **config naming mismatch bug** that will silently break user overrides, to missing abort propagation, and a coupling concern in the JS idle check. See inline comments.

packages/dom/src/serialize-dom.js
Shivanshu-07: Fixed. `serializeDOMWithReadiness` now wraps the `waitForReady` call in a try/catch and falls back to synchronous serialization on error (graceful degradation). See packages/dom/src/serialize-dom.js:102-121 on commit c879f34e.

packages/dom/src/serialize-dom.js
Shivanshu-07: Valid concern — fixed. Split `serializeDOM` into two functions:

- `serializeDOM()` — remains **synchronous** and backward-compatible. All existing SDKs (`@percy/cypress`, `@percy/puppeteer`, `@percy/selenium-webdriver`) that do `PercyDOM.serialize(options)` without `await` continue to work unchanged.
- `serializeDOMWithReadiness()` — new async variant. Used only by the CLI URL-capture path (page.js uses CDP `awaitPromise: true` to await it) and by new SDK versions that opt into readiness-gated capture.

If readiness fails internally, the function falls back to sync serialization (graceful degradation) rather than breaking the capture. See packages/dom/src/serialize-dom.js:93-121.

packages/dom/src/readiness.js
Shivanshu-07: Fixed. Added `normalizeOptions()` helper in packages/dom/src/readiness.js that accepts either camelCase (what the schema defines) or snake_case, and maps both to the snake_case keys used internally.

The merge also now only overrides preset defaults for user keys whose value is not `undefined`, so a partial override no longer wipes other preset values:

```js
let userOptions = normalizeOptions(options);
let config = { ...preset };
for (let key of Object.keys(userOptions)) {
  if (userOptions[key] !== undefined) config[key] = userOptions[key];
}
```

Added a direct unit test covering both naming conventions, precedence when both are passed, and falsy-value (0, false) preservation.

packages/dom/src/readiness.js
Shivanshu-07: Fixed. Added a dedicated `js_idle_window_ms` config (camelCase `jsIdleWindowMs` in the schema) that is independent of `stability_window_ms`.

Preset defaults now decouple them sensibly:

| Preset    | stability_window_ms | js_idle_window_ms |
|-----------|---------------------|-------------------|
| fast      | 100                 | 100               |
| balanced  | 300                 | 300               |
| strict    | 1000                | **500**           |

With `strict`, DOM-stability still waits 1000ms but the JS idle check only needs 500ms of no long tasks — no more 1-full-second timeouts on pages with normal JS activity.

If a user passes a custom config that predates this option, `runAllChecks` falls back to `stability_window_ms` for backward compat. Added integration tests proving the decoupling works and the fallback still works.

packages/dom/src/readiness.js
Shivanshu-07: Fixed. `checkFontReady` now accepts an `aborted` handle and clears the 5s font timer on abort:

```js
function checkFontReady(aborted) {
  let start = performance.now();
  if (!document.fonts?.ready) return Promise.resolve({ passed: true, duration_ms: 0, skipped: true });
  let fontTimer;
  let result = Promise.race([
    document.fonts.ready.then(() => ({ passed: true, duration_ms: Math.round(performance.now() - start) })),
    new Promise(r => { fontTimer = setTimeout(() => r({ passed: false, duration_ms: 5000, timed_out: true }), 5000); })
  ]);
  if (aborted) aborted.onAbort(() => { if (fontTimer) clearTimeout(fontTimer); });
  return result;
}
```

Also updated `runAllChecks` to pass the abort handle through. Matches the abort pattern used by every other check.

packages/core/src/page.js
Shivanshu-07: Fixed. Simplified to use native await — CDP `awaitPromise: true` auto-awaits the returned Promise, and `await` on the sync path is a no-op:

```js
let capture = await this.eval(async (_, options) => {
  let fn = (PercyDOM.serializeDOMWithReadiness || PercyDOM.serialize);
  let domSnapshot = await fn(options);
  return { domSnapshot, url: document.URL };
}, { ...opts, readiness });
```

The `PercyDOM.serializeDOMWithReadiness || PercyDOM.serialize` fallback keeps page.js compatible with older injected dom bundles that don't yet export the new async variant.

packages/dom/src/readiness.js
Shivanshu-07: Fixed. Replaced the 50ms polling of `performance.getEntriesByType('resource')` with a `PerformanceObserver` that subscribes incrementally to `'resource'` entries:

```js
try {
  observer = new PerformanceObserver(list => {
    if (aborted.value) return;
    if (list.getEntries().length > 0) resetIdleTimer();
  });
  observer.observe({ type: 'resource', buffered: false });
} catch (e) {
  // Older browser — fall back to polling
  ...
}
```

No more scanning the entire resource list every tick — the observer fires only when a new resource entry is added, matching the pattern already used in `checkJSIdle` for longtasks. The polling path is preserved as a fallback for very old browsers that lack `PerformanceObserver`.

packages/core/src/config.js
Shivanshu-07: I verified this and I think the premise here might be inverted. Grepping the existing schema in `packages/core/src/config.js`:

```
enableJavaScript, cliEnableJavaScript, disableShadowDOM,
forceShadowAsLightDOM, enableLayout, domTransformation, reshuffleInvalidTags,
deferUploads, useSystemProxy, skipBaseBuild, browserName, browserVersion,
osVersion, deviceName, percyBrowserCustomName, diffSensitivity,
imageIgnoreThreshold, carouselsEnabled, bannersEnabled, adsEnabled,
minHeight, percyCSS, scopeOptions, ignoreRegionSelectors, freezeAnimation,
freezeAnimatedImage, responsiveSnapshotCapture, testCase, thTestCaseExecutionId,
fullPage ...
```

Every existing snapshot option in the schema is already camelCase — there are no `enable_javascript` / `dom_transformation` / `reshuffle_invalid_tags` entries in `config.js`. So the readiness schema (`stabilityWindowMs`, `timeoutMs`, etc.) is actually consistent with the rest of the file.

The snake_case you're seeing in `readiness.js` is the **internal** naming used after normalization (matches the diagnostic output keys like `stability_window_ms`, `network_idle_window_ms`, which are the payload contract with the backend). The `normalizeOptions()` helper bridges the two, and it accepts both forms so direct snake_case overrides still work if anyone is using them.

Happy to rename if you'd still prefer snake_case for the user-facing schema, but the current naming was chosen specifically to match the surrounding camelCase conventions.

packages/dom/src/readiness.js
Shivanshu-07: Fixed. `href` removed from the generic `LAYOUT_ATTRIBUTES` set, replaced with a scoped check in `isLayoutMutation`:

```js
// href is only layout-affecting on <link> elements (stylesheets).
// On <a> tags changing href is a no-op for layout.
if (attr === 'href') return mutation.target.tagName === 'LINK';
```

`href` is still in the `MutationObserver` `attributeFilter` so the observer fires — the classifier just returns `false` for `<a>` tags now, so it doesn't reset the stability timer.

Added direct unit tests for both cases (`<a href>` returns `false`, `<link href>` returns `true`) in the new readiness-helpers test file.

packages/dom/src/readiness.js
Shivanshu-07: Fixed. Dropped from ~15 function-wide `istanbul ignore` annotations to narrow, justified ignores on genuinely untestable paths only. Concrete changes:

1. **Exported internal helpers** from `readiness.js` for direct unit testing:
   - `isLayoutMutation`, `hasLayoutStyleChange`, `parseStyleProps`, `normalizeOptions`, `createAbortHandle`

   (These are not added to the public `index.js` surface — tests import from `../src/readiness`, matching the pattern used by `serialize-frames`, `serialize-cssom`, `transform-dom` etc.)

2. **Added `packages/dom/test/readiness-helpers.test.js`** — 36 new unit tests covering every branch of the exported helpers (empty inputs, multi-decl parsing, prefix-matched layout props like `min-`/`max-`/`flex`/`z-index`, `data-`/`aria-` filtering, `<a>` vs `<link>` href handling, camelCase precedence, falsy-value preservation, abort callback ordering, etc.).

3. **Removed function-wide ignores** from six check functions (`checkDOMStability`, `checkNetworkIdle`, `checkImageReady`, `checkJSIdle`, `checkReadySelectors`, `checkNotPresentSelectors`). These are now measured through integration tests.

4. **Narrow ignores remain only** on genuinely untestable paths, each with a one-line reason:
   - Browser-API availability (`document.fonts?.ready`, Long Task API `try/catch`, `requestIdleCallback` else-branch) — these are always available in Chrome/Firefox; fallbacks are for older browsers.
   - Font 5s timeout race — impractical to wait for in tests.
   - Defensive `if (aborted.value)` guards inside `setInterval`/`MutationObserver` callbacks — `abort()` clears the interval/disconnects the observer synchronously, so the check is dead code in practice.
   - Empty-selector early returns — orchestrator only calls those checks when `selectors.length > 0`.
   - Error-safety-net `catch` blocks.
   - `/* istanbul ignore else */` for the `requestIdleCallback` availability check (only the fallback branch is ignored, not the whole statement).

**Result:** `readiness.js` now shows **100% lines / 100% branches / 100% functions / 100% statements** instead of having 6 functions wholly excluded from the report.

rishigupta1599 (commented):
test

rishigupta1599 (commented):
Re-review summary

Thanks for the turnaround. The pivot to a two-call pattern (`waitForReady` then sync `serialize`) cleanly resolves the biggest concern from the last round (SDK breakage from a Promise-returning `serializeDOM`). Specifically:

- camelCase / snake_case config mismatch — fixed via `normalizeOptions` + tests.
- JS-idle / stability-window coupling — decoupled via dedicated `js_idle_window_ms`.
- `checkFontReady` now honors `aborted` and clears the 5s timer.
- `href` on `<a>` no longer resets stability.
- Network-idle polling replaced with `PerformanceObserver`.
- Redundant thenable check in `page.js` removed.
- Coverage suppressions trimmed to narrow, justified ignores.

All prior blockers addressed. A few new/remaining items from this round's pushes — non-blocking but worth addressing before merge. See inline.

(Please ignore the stray "test" review I accidentally posted just before this one — API retry artifact. This is the real one.)

packages/core/src/page.js
```diff
@@ -213,15 +213,31 @@ export class Page {
 
     await this.insertPercyDom();
 
-    // serialize and capture a DOM snapshot
-    this.log.debug('Serialize DOM', this.meta);
+    // Run readiness checks before serializing — wait for page stability
+    let readiness = snapshot.readiness || this.browser?.percy?.config?.snapshot?.readiness;
+
+    if (readiness && readiness.preset !== 'disabled') {
```
rishigupta1599, line 287: `diagnostics` is fetched here but never attached to `capture`/`domSnapshot`. Combined with the new reader in `snapshot.js` that looks for `migrated.domSnapshot?.readiness_diagnostics`, that reader is effectively dead code on the CLI URL path — we capture diagnostics only to log `timed_out` locally and then drop the rest.

If the intent is to surface readiness metrics on the backend/UI, merge `diagnostics` into `capture` (e.g. `return { ...snapshot, ...capture, readiness_diagnostics: diagnostics }`) or attach onto `domSnapshot` before returning. Otherwise please remove the reader in `snapshot.js` until an SDK path actually embeds them — right now it's a no-op branch with a misleading implication.

packages/core/src/page.js
```diff
@@ -213,15 +213,31 @@ export class Page {
 
     await this.insertPercyDom();
 
-    // serialize and capture a DOM snapshot
-    this.log.debug('Serialize DOM', this.meta);
+    // Run readiness checks before serializing — wait for page stability
+    let readiness = snapshot.readiness || this.browser?.percy?.config?.snapshot?.readiness;
+
```
rishigupta1599, line 286: `/* istanbul ignore next */` on the outer `await this.eval(...)` also swallows the `.catch(e => this.log.debug(...))` handler, which is plain Node-side JS that *is* testable (reject the eval spy and assert the debug log). Move the ignore to the inner arrow only, or drop it and cover the catch — it's the kind of silent failure path (`Readiness check failed: ...`) we rely on in production triage.

packages/core/src/snapshot.js
```diff
@@ -220,6 +220,17 @@ export function validateSnapshotOptions(options) {
     log.warn('Encountered snapshot serialization warnings:');
     for (let w of domWarnings) log.warn(`- ${w}`);
   }
+
+  // log readiness diagnostics when present (SDK-submitted snapshots with readiness enabled)
+  let readinessDiag = migrated.domSnapshot?.readiness_diagnostics;
```
rishigupta1599: Two concerns:

1. `migrated.domSnapshot` is a union of `string` (legacy SDK payload — just HTML) and `object` (`{ html, warnings, ... }`). Optional chaining on a string returns `undefined` so this is safe today, but an explicit `typeof migrated.domSnapshot === 'object'` gate reads more clearly and avoids future footguns.

2. Nothing in this PR actually populates `domSnapshot.readiness_diagnostics` — the CLI path in `page.js` discards diagnostics, and no SDK yet embeds them. This reader will stay dormant until an SDK change lands. Either land the producer in the same PR (attach diagnostics in `page.js`, update `sdk-utils` helpers to embed them in the returned payload) or drop this block until the producer ships — to avoid confusion about what's actually wired up.

packages/sdk-utils/src/serialize-dom.js
```diff
@@ -0,0 +1,60 @@
...
+//
+//   // Selenium (executeAsyncScript with callback):
+//   driver.execute_async_script(waitForReadyScript(config, { callback: true }));
+//
+// Graceful degradation:
+//   - If PercyDOM.waitForReady is not available (old CLI): resolves immediately
+//   - If waitForReady throws: resolves immediately (catch swallows the error)
+//   - If readiness times out: waitForReady resolves with { timed_out: true }
+export function waitForReadyScript(readinessConfig = {}, { callback = false } = {}) {
+  let config = JSON.stringify(readinessConfig);
```
rishigupta1599: `JSON.stringify(readinessConfig)` is interpolated straight into the returned JS source. Edge cases:

1. **U+2028 / U+2029**: valid in JSON strings but were illegal in JS source string literals pre-ES2019. If a user's `readySelectors` / `notPresentSelectors` entry contains these (unlikely but possible via copy-paste), older engines could throw a SyntaxError when evaluating the script. `JSON.stringify` doesn't escape them. Cheap fix: `config.replace(/\u2028/g, '\\u2028').replace(/\u2029/g, '\\u2029')`.

2. **`</script>` sequence**: not exploitable here (the script is sent via CDP/executeScript, not inlined into HTML), but worth a JSDoc note that this helper's output must not be inlined into HTML without additional escaping, so downstream SDK authors don't misuse it.

Low risk in practice, high value for defensive consistency.

packages/dom/src/readiness.js
```diff
@@ -0,0 +1,501 @@
...
+      if (observer) observer.disconnect();
+      /* istanbul ignore next: pollInterval is only set on the fallback path */
+      if (pollInterval) clearInterval(pollInterval);
+      /* istanbul ignore next: timer is always set before abort can fire */
+      if (timer) clearTimeout(timer);
+    });
+  });
+}
+
+function checkFontReady(aborted) {
```
rishigupta1599, line 249: Small correctness nit on the abort path: after `aborted.abort()` fires, `fontTimer` is cleared, but the `document.fonts.ready.then(...)` promise is still live and will eventually settle the race with `{ passed: true }` — *after* the orchestrator has already declared timeout. In the current code this is benign because `runAllChecks` is orphaned post-timeout and nothing reads `result.checks.font_ready` again. But if we later mutate `result.checks` after the race (e.g. to merge late-arriving data or emit per-check metrics), this becomes a latent race. Consider resolving the inner promise with `{ passed: false, aborted: true }` on abort to make cancellation deterministic.

packages/core/src/page.js
Shivanshu-07: Addressed in a1046bf2 — `readinessDiagnostics` is now attached to `capture.domSnapshot` (when it's the structured object form) before returning, so the reader in `snapshot.js` is reachable on the CLI URL path and the backend/UI can surface readiness metrics. New spec `logs when readiness times out and continues to capture` asserts `domSnapshot.readiness_diagnostics` is present.

packages/core/src/page.js
Shivanshu-07: Addressed in a1046bf2 — moved `/* istanbul ignore next */` inside the `eval(...)` call so it scopes only to the injected arrow function (the genuinely uninstrumented browser code). The outer `await ... .catch(...)` is now testable, and a new spec `debug logs when the readiness eval rejects and continues to capture` rejects the readiness eval and asserts the `Readiness check failed: ...` debug log fires while serialize still completes.

packages/core/src/snapshot.js
Shivanshu-07: Both addressed in a1046bf2:

1. Now gated on an explicit `typeof migrated.domSnapshot === 'object'` check (`domSnapshotObj`) before reading `readiness_diagnostics`.
2. The CLI URL path in `page.js` now attaches diagnostics onto the captured `domSnapshot`, so this reader is no longer dead code — it fires both for SDK-submitted snapshots (once SDK PRs land) and for the CLI URL-capture path today.

packages/sdk-utils/src/serialize-dom.js
Shivanshu-07: Addressed in a1046bf2:

1. `JSON.stringify(readinessConfig)` now post-processes its output with `.replace(/ /g, '\\u2028').replace(/ /g, '\\u2029')` so the emitted JS source is legal on pre-ES2019 hosts. New sdk-utils spec asserts the escape sequences are present and the raw separators are absent.
2. Added a JSDoc warning: `IMPORTANT: The output is intended for CDP / executeScript / executeAsyncScript channels. Do NOT inline this string into HTML — </script> sequences in user config would break out of a <script> tag.`

packages/dom/src/readiness.js
Shivanshu-07: Addressed in a1046bf2 — `checkFontReady` now races a third `abortPromise` alongside `document.fonts.ready` and the 5s timer. On abort, the orchestrator settles the race deterministically with `{ passed: false, aborted: true }` instead of letting a late `fonts.ready` flip the result to `{ passed: true }` after timeout has been declared. Future late mutations on `result.checks.font_ready` are now safe.

Shivanshu-07:
### Review feedback addressed in [a1046bf2](https://github.com/percy/cli/pull/2184/commits/a1046bf2)

| # | Reviewer concern | Resolution |
|---|---|---|
| 1 | `diagnostics` fetched in `page.js` but never attached → reader in `snapshot.js` is dead code | Diagnostics now attached onto `capture.domSnapshot` (object form) before returning, so backend/UI can surface readiness metrics on the CLI URL path |
| 2 | `/* istanbul ignore next */` swallows the `.catch` handler in `page.js` | Ignore moved inside the `eval(...)` call, scoped only to the injected arrow. Added spec that rejects readiness eval and asserts `Readiness check failed: ...` debug log fires |
| 3 | `migrated.domSnapshot?.readiness_diagnostics` reads cleanly off a string but is unclear | Explicit `typeof migrated.domSnapshot === 'object'` gate via `domSnapshotObj`. Reader is no longer dead code — populated by the CLI URL path today and by SDKs once their PRs land |
| 4 | `JSON.stringify(readinessConfig)` interpolates without escaping U+2028/U+2029 | Escape both with `.replace(/ /g, '\\u2028').replace(/ /g, '\\u2029')`. Added JSDoc warning that the output must not be inlined into HTML. New sdk-utils spec asserts the escape sequences are present |
| 5 | `checkFontReady` abort: `document.fonts.ready` can flip the race to `{ passed: true }` after timeout | Race now includes a third `abortPromise` that resolves with `{ passed: false, aborted: true }` on abort, settling the result deterministically |

CI is green on Linux (Linux `Test @percy/core` passed at 11m22s, all other checks pass). Windows variant still in progress.

packages/dom/src/serialize-dom.js
```diff
@@ -80,7 +80,22 @@ export function waitForResize() {
...
+// All SDKs and the CLI call this function for DOM serialization.
+//
+// Readiness checks (DOM stability, network idle, fonts, images, JS idle)
+// run BEFORE this function via a separate `PercyDOM.waitForReady()` call.
+// This two-call pattern keeps serialize sync (zero SDK breakage) while
+// isolating the async readiness into its own step:
+//   1. await PercyDOM.waitForReady(config)  — async, page stabilizes
+//   2. PercyDOM.serialize(options)           — sync, captures stable DOM
 export function serializeDOM(options) {
+  return _serialize(options);
```
rishigupta1599: why this change is required if we have anyways created new method for readiness?

packages/dom/src/readiness.js
```diff
@@ -0,0 +1,512 @@
...
+    }
+
+    observer.observe(document.documentElement, {
+      childList: true,
+      attributes: true,
+      attributeOldValue: true,
+      subtree: true,
+      attributeFilter: [...LAYOUT_ATTRIBUTES, 'style', 'href']
+    });
+    timer = setTimeout(settle, stabilityWindowMs);
```
rishigupta1599, line 188: this can cause issues in heavy DOM as we have seen mutation observers causing high memory spikes , specially here we are checking each and every mutations

packages/dom/src/readiness.js
```diff
@@ -0,0 +1,512 @@
...
+      let incomplete = [];
+      for (let img of imgs) {
+        let r = img.getBoundingClientRect();
+        /* istanbul ignore else: test images are always placed in the viewport with non-zero dimensions */
+        if (r.top < vh && r.bottom > 0 && r.width > 0 && r.height > 0) {
+          if (!img.complete || img.naturalWidth === 0) incomplete.push(img);
+        }
+      }
+      return incomplete;
+    }
```
rishigupta1599, line 290: lot of code looks duplicated as all three checks are using PerformanceObserver

packages/dom/src/readiness.js
```diff
@@ -0,0 +1,512 @@
...
+}
+
+function checkReadySelectors(selectors, aborted) {
+  /* istanbul ignore next: orchestrator only calls this when selectors.length > 0; defensive for direct callers */
+  if (!selectors?.length) return Promise.resolve({ passed: true, duration_ms: 0, selectors: [] });
+  return new Promise(resolve => {
+    let start = performance.now();
+    function check() {
+      for (let s of selectors) {
+        let el = document.querySelector(s);
```
rishigupta1599: this will only support CSS selector and not xpath in that case we should try to support both

packages/dom/src/serialize-dom.js
Shivanshu-07: Addressed in 3851d998 — `_serialize` was a leftover from when `serializeDOMWithReadiness` shared the same serialization core. After cd001abc removed `serializeDOMWithReadiness` in favor of the two-call pattern (`waitForReady` + `serialize`), the wrapper was redundant. Inlined `_serialize` back into `serializeDOM` so the file matches its pre-readiness shape — same surface (one exported `serializeDOM`), no behavior change. @percy/dom: 662 specs pass, 100/100/100/100 coverage.

packages/dom/src/readiness.js
Shivanshu-07: Addressed in beb8881c — added an explicit `domStability` kill-switch (`dom_stability: false` / `domStability: false`) that lets users skip the MutationObserver check entirely on heavy-DOM pages where the observer itself can drive CPU/memory pressure. The other checks (`js_idle`, `image_ready`, `font_ready`, `ready_selectors`, `not_present_selectors`, `network_idle`) continue to gate capture, so the page is still readiness-gated — just not on raw mutation rate. Defaults to `true` across all presets so existing configs are unchanged. The orchestrator (readiness.js:462) now reads `config.dom_stability !== false && config.stability_window_ms > 0` before scheduling the check. 662 specs pass, 100% coverage retained.

packages/dom/src/readiness.js
Shivanshu-07: Addressed in 87b6556c — extracted `observePerformance(type, onEntries)` helper (readiness.js:139-147) that wraps the `try { new PerformanceObserver(...).observe({ type, buffered: false }) } catch { return null }` pattern. `checkNetworkIdle` (resource entries) and `checkJSIdle` (longtask entries) now share that boilerplate; each callsite stayed in charge of its own fallback (network idle falls back to polling; JS idle degrades to rIC/rAF). Net: removed ~20 lines of duplicated setup/error-handling between the two checks. Coverage stays at 100%.

packages/dom/src/readiness.js
Shivanshu-07: Addressed in 87b6556c — added `resolveSelector` (readiness.js:107-131) that supports both CSS and XPath. It sniffs XPath by leading `/`, `//`, `./`, `(/`, `(./` (covers absolute, relative, parenthesised XPath); everything else routes to `document.querySelector`. Also accepts explicit `{ css }` / `{ xpath }` object form for ambiguous cases like a CSS selector starting with `/`. Malformed XPath / non-Element results return `null` so the readiness gate keeps polling instead of throwing. `checkReadySelectors` and `checkNotPresentSelectors` both go through it now. Config schema in `config.js` updated to allow array items as string or `{ css?, xpath? }` object. Added 12 `resolveSelector` unit tests + 3 integration tests for CSS+XPath / object form / mixed selectors.

rishigupta1599 (approved):

PR #2187, opened 2026-04-17
chore: upgrade Chromium from v126 to v143 for asset discovery

Manoj-Katta:
# Chromium v126 → v143 upgrade (PR #2187)

This PR upgrades the bundled Chromium binary that `@percy/core` uses for asset discovery from **v126.0.6478.184 → v143.0.7499.169**, to match the version `percy-renderer` ships server-side.

The version jump crosses **17 major Chrome versions** including:
- **Chrome 128** — old-headless → new-headless rewrite ([blog](https://developer.chrome.com/blog/removing-headless-old-from-chrome))
- **Chrome 130** — Local Network Access opt-in
- **Chrome 132** — `--headless=old` deleted entirely
- **Chrome 143** — Local Network Access default-on, malformed Content-Length leniency, PlzDedicatedWorker default

Many "regressions" we hit are not bugs — they are intentional Chrome behaviors that old headless quietly didn't enforce. This document walks each one, the diagnostic story behind it, and the fix.

> **TL;DR for reviewers:** Production-code changes concentrated in `network.js` (with smaller diffs in `browser.js` and `install.js`) covering 9 categories of v143 behavior change. The `network.js` changes are layered: (a) Chrome 128 site isolation and v143 launch-flag work in `browser.js`; (b) malformed Content-Length pre-abort and the worker-script direct-fetch fallback with size/origin/timeout safety nets in `network.js`; (c) test coverage for every new branch including the deterministic `captureResourceDirectly` catch test that closes the L755 gap. Per ce:review, 11 must-fix/should-fix items are applied; 5 reviewer comments are documented as deferred with reasons.

### 1.1. Site isolation became default-on (Chrome 128)

**What changed:** Cross-origin sub-resources and worker scripts now run in *separate renderer processes* by default. `--disable-site-isolation-trials` only affected opt-in trials, not the baseline-on behavior.

**How it broke Percy:** Percy's CDP `Fetch.enable` and `Network.enable` are attached to the page session. Events on a sibling renderer process don't reach those listeners → worker fetches and some cross-origin sub-resources were silently uncaptured.

**What we did:** Added `IsolateOrigins,site-per-process` to the `--disable-features=` launch flag (`browser.js`). Keeps cross-origin sub-resources in the same renderer process, matching v126 old-headless behavior.

**Risk:** None for Percy's use case. CORS/CORB/ORB enforcement on actual cross-origin internet hosts is preserved — only Chrome's process-isolation is relaxed.

### 1.2. HTTPS-first auto-upgrade in headless (Chrome 143)

**What changed:** Chrome 143 silently auto-upgrades any HTTP navigation to HTTPS. If HTTPS isn't reachable, Chrome blocks navigation with `net::ERR_BLOCKED_BY_CLIENT`.

**How it broke Percy:** Customers running asset discovery against HTTP URLs (local dev servers, CI environments, staging) would see silent navigation failures.

**What we did:** Added `HttpsFirstBalancedModeAutoEnable` to `--disable-features=`. HTTP navigation proceeds normally.

**Risk:** None — Percy is a tool, not a browser. Users *intend* to load the URL they pass in.

### 1.3. Local Network Access checks default-on (Chrome 143)

**What changed:** Pages fulfilled via `Fetch.fulfillRequest` (Percy's `domSnapshot` root-document path) are treated as an external context by Chrome's LNA gating. Sub-resource requests to `localhost`, `127.0.0.1`, or RFC 1918 private addresses are blocked with `corsError = LocalNetworkAccessPermissionDenied`.

**How it broke Percy:** 12 cross-origin and auto-domain-validation discovery tests failed. Percy's logger swallowed the actual Chrome error — visible only via raw CDP. The wrong hypothesis (a planned dual-stage Fetch interception refactor) would have wasted weeks.

**Diagnostic story:** Wrote a standalone raw-CDP Node script that bypassed Percy entirely and replicated the exact `Fetch.fulfillRequest` flow. The probe surfaced `LocalNetworkAccessPermissionDenied`. Cross-referenced with [Chrome's LNA spec](https://developer.chrome.com/blog/local-network-access).

**What we did:** Added `LocalNetworkAccessChecks` to `--disable-features=`. **Narrower than `--disable-web-security`** — preserves CORS/CORB/ORB on real internet hosts, only relaxes Chrome's loopback/private-network gating, which is exactly what asset discovery needs.

**Risk:** None for headless asset discovery. The renderer side doesn't use this flag.

### 1.4. Malformed `Content-Length` no longer triggers Chrome self-abort (Chrome 143)

**What changed:** In v126, Chrome's HTTP parser detected malformed `Content-Length` (`NaN`, non-numeric) and dropped the connection. `Network.loadingFailed` fired and Percy logged the failure. **In v143, Chrome is lenient — it waits indefinitely for body bytes that never come.** No event reaches Percy.

**How it broke Percy:** A page with `<link href="bad.css">` (where the server returns malformed CL) would hang. `Page.lifecycleEvent name=load` never fires. `Page.navigate({waitUntil:'load'})` times out at 30s, retries 3 times → **93s of dead air**, then snapshot fails.

**Diagnostic story:** Reproduced the 93s timeout. Added per-event CDP tracing in `network.js`. Confirmed `Network.responseReceived` fired (status=200, headers received) but `Network.loadingFinished` and `Network.loadingFailed` *never* fired. The connection was hanging in Chrome's network stack.

**What we did (5 pieces, all in `network.js`):**

1. **Opt into response-stage Fetch interception** via `interceptResponse: true` on `Fetch.continueRequest`. Chrome now pauses every intercepted request a second time once response headers are received (but before body bytes stream).

2. **Add `_handleResponsePaused`** to inspect Content-Length at the response stage. If `tooLarge` or `malformed`, call `Fetch.failRequest({errorReason: 'Aborted'})` proactively — Chrome stops waiting, `<link>` resolves as failed, page `load` fires.

3. **Two helper functions** (`headersArrayToObject`, `inspectContentLength`) at module scope for clean inspection.

4. **Lifecycle ordering: `Fetch.failRequest` BEFORE `_forgetRequest`.** v126's malformed-CL handling was implicit (Chrome auto-aborted); v143 forces us to call `Fetch.failRequest` explicitly. If our `failRequest` call itself throws (transient CDP error, target detach race), running `_forgetRequest` first would leak Chrome's Fetch state — Percy thinks the request is done, but Chrome's interception is still paused indefinitely. We now finalize the disposition first, only forgetting the request after. On unrecognized errors, a last-resort `Fetch.continueResponse` attempts to un-pause Chrome so it doesn't leak.

5. **Observability for stuck-snapshot diagnostics — currently at `debug`, telemetry follow-up tracked.** Unknown errors in `_handleResponsePaused`'s `failRequest` catch and `_continueResponse`'s catch stay at `log.debug`. We initially promoted these to `log.warn` for default-loglevel visibility, then reverted (commit `334f0621`) because routine `Session closed` / `Target closed` lifecycle races fired these catches and the warn noise drowned legitimate signals. The right home for this signal is structured instrumentation telemetry rather than user-facing logs; tracking that as a separate follow-up so customer logs stay quiet on routine teardowns.

**Pattern established:** `Fetch.failRequest` is used only when *Percy* asserts a failure (our oversized-CL abort). For errors Chrome already reports, we use `Fetch.continueResponse` to defer to Chrome's natural `loadingFailed` flow — this preserves Chrome's specific `errorText` (e.g. `net::ERR_EMPTY_RESPONSE`) instead of synthesizing a generic `net::ERR_FAILED` (which Percy's `asset_load_missing` instrumentation explicitly filters as noise).

**Side benefit:** Pre-fix, Percy's `MAX_RESOURCE_SIZE` check inside `saveResponseResource` ran *after* Chrome had already buffered the full body. Our new check fires *before* the body streams — saves bandwidth and memory, and unsticks the page render faster. This was a latent inefficiency in v126 too; v143 forced us to fix it.

**Cost:** Every Fetch-intercepted request now does one extra CDP roundtrip. Sub-millisecond on localhost websocket. Puppeteer and Playwright use the same dual-stage interception pattern.

### 1.5. PlzDedicatedWorker — dedicated worker scripts disappear from CDP (Chrome 143)

**What changed:** Chrome 143 made **PlzDedicatedWorker** the default ([Chromium issue 40093136](https://issues.chromium.org/issues/40093136) — *"Implement browser-initiated dedicated worker script load"*). Dedicated worker scripts are now fetched in the **browser process** via `WorkerScriptFetcher`, not in the renderer URLLoader proxy chain. The renderer-side URLLoader proxy is what historically emitted `Network.responseReceived` for worker scripts "for free." That path is now bypassed.

**The precise mechanism (verified against Chromium source):** Under PlzDedicatedWorker, DevTools instrumentation for the worker-script fetch is handled by [`content/browser/devtools/devtools_instrumentation.cc`](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/devtools_instrumentation.cc). That file has hand-coded hooks for the worker path: `OnWorkerMainScriptRequestWillBeSent` (emits `Network.requestWillBeSent`) and `OnWorkerMainScriptLoadingFailed` (emits `Network.loadingFailed`). **There is no corresponding `OnWorkerMainScriptResponseReceived` hook.** Service workers have one (`OnServiceWorkerMainScriptFetchingFailed → network_handler->ResponseReceived(...)`); dedicated workers don't. So `Network.responseReceived` is **never emitted on any CDP session** for a successful dedicated-worker main-script fetch — it's a Chromium instrumentation gap, not a routing problem.

The `NetworkServiceDevToolsObserver` attached to the worker-script URLLoaderFactory is bound to the **page frame's** devtools token (see [`content/browser/worker_host/worker_script_fetcher.cc`](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/worker_host/worker_script_fetcher.cc) — `MakeSelfOwned(creator_render_frame_host->GetDevToolsFrameToken().ToString())`), which is why the page session sees `requestWillBeSent`. The `loadingFinished` that does fire on the **worker** session arrives via a separate path — the worker target's own `NetworkHandler`, created at auto-attach by [`content/browser/devtools/dedicated_worker_devtools_agent_host.cc`](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/dedicated_worker_devtools_agent_host.cc)'s `AttachSession`, receives completion from the renderer once the worker is alive — but that handler never received `requestWillBeSent` or `responseReceived` from the browser-side fetch, so `Network.getResponseBody` against it returns "no resource with given identifier found."

**Auto-attach (`Target.setAutoAttach`) is not the cause** — it merely makes the asymmetry visible because there's a worker session to receive `loadingFinished`. A single-session debugger sees the same loss of `responseReceived`.

**How it broke Percy:** Two layered failures.

*Failure A — `idle()` hangs forever.* Percy's `_handleLoadingFinished` awaits `Network.responseReceived` unconditionally. Since Chromium never emits that event for the worker script, the await blocks forever → the request stays in `#requests` → `idle()` never returns → snapshot hits `Page.TIMEOUT × 3 retries = 93s`.

*Failure B — Worker bytes never reach CDP.* Even after we time out and clean up `#requests`, the renderer (server-side) still needs the worker.js bytes to construct the `Worker` at render time. Neither the page session's `NetworkHandler` nor the worker session's `NetworkHandler` ever stored the body, so `Network.getResponseBody` returns "no resource with given identifier found" on both sessions.

**Diagnostic story:**
- Wrote a standalone raw-CDP Node script that spawns v143 Chromium, navigates to a worker page, logs every Network/Fetch/Target event across all sessions, and tries `Network.getResponseBody` on each session that emitted events. Confirmed: `responseReceived` is absent on every session; `loadingFinished` fires on the worker session with the page's requestId (= the worker token); `getResponseBody` returns "no resource with given identifier found" everywhere.
- Confirmed against Chromium source: no `OnWorkerMainScriptResponseReceived` hook exists in `content/browser/devtools/devtools_instrumentation.cc`.
- Cross-checked with [Playwright issue #31747](https://github.com/microsoft/playwright/issues/31747) and [PR #35692](https://github.com/microsoft/playwright/pull/35692), and [Puppeteer PR #13752](https://github.com/puppeteer/puppeteer/pull/13752) — both upstream projects independently hit the same gap.

**What we did (in `network.js`):**

1. **`RESPONSE_RECEIVED_TIMEOUT = 2000`** — module-scope named constant. Cap the `responseReceived` await in `_handleLoadingFinished` with a `Promise.race` against a 2-second timer. Worker scripts wake up the handler instead of hanging it. Comment notes the cumulative `N*2s` worst case so future readers can reason about cost.

2. **Debug log when the timeout fires** (`Skipping resource: responseReceived not received within 2000ms - <url>`). Asserted in `captures requests from workers` so a regression is observable — if a future Chrome restores the v126 lifecycle, the negated assertion will fail and tell us to remove the workaround.

3. **`captureResourceDirectly` fallback** (renamed from `captureScriptDirectly` — the function captures any allowlisted resource that fell through to direct fetch, not just worker scripts). When the timeout fires and the URL is in `allowedHostnames` AND NOT in `disallowedHostnames`, fetch the bytes directly via Node HTTP (`makeDirectRequest`, the same helper used for fonts since v126). Save them as a Percy resource with mimetype derived from the URL extension via `mime.lookup()`.

4. **`DIRECT_FETCH_TIMEOUT = 5000`** — module-scope constant. Wraps the `makeDirectRequest` call inside `captureResourceDirectly` in `Promise.race` against a 5-second timer. Caps the wall-clock cost if a worker host accepts the TCP connection but stalls. Pre-fix worst case was `~30s` (`networkIdleWaitTimeout`); post-fix is `~7s` (`RESPONSE_RECEIVED_TIMEOUT + DIRECT_FETCH_TIMEOUT`).

5. **Size guard + real HTTP status in `captureResourceDirectly`.** `makeDirectRequest` now returns `{ body, status }` via the `makeRequest` callback. The fallback rejects bodies over `MAX_RESOURCE_SIZE` (25 MB) — matching `saveResponseResource`'s guard — to keep the WebSocket payload safe. The Percy resource records the real HTTP status from the direct fetch instead of a hardcoded `200`. Font path call site updated to destructure `{ body }`.

6. **Disallowedhostnames precedence at the fallback gate.** The gate in `_handleLoadingFinished` now mirrors `sendResponseResource`'s pattern — checks `disallowedHostnames` first, `allowedHostnames` second. Customers who explicitly blocklist an internal host won't see Percy's Node fallback bypass that gate. A debug log fires when the fallback is skipped due to hostname gating so the silent-skip is observable.

7. **Same-origin Authorization in `makeDirectRequest`.** Browser's URLLoader origin-scopes Basic auth automatically; our Node fallback didn't. Extracted as an exported `shouldAttachAuth(authorization, requestUrl, snapshotUrl)` pure helper that compares the target URL's origin against `network.meta?.snapshotURL`'s origin and only returns true when both URLs parse and origins match. Malformed URLs return false defensively. Prevents customer credentials leaking to redirected third-party origins, and makes the same-origin / cross-origin / malformed-URL branches unit-testable without `/* istanbul ignore */` markers.

8. **`makeDirectRequest` reads cookies from the page session, not the triggering session.** When `captureResourceDirectly` is called from `_handleLoadingFinished` bound to the **worker session** (v143 worker-script case), `Network.getCookies` on that session throws "Internal error" because the worker session's `NetworkHandler` has only a partial Network domain. Extracted as an exported `pickCookieSession(network, session)` helper that returns `network.page.session` (the top-level page session, full Network domain) when available and falls back to the triggering session otherwise. Both branches of the `??` are unit-testable via the helper. A defensive try/catch with empty-cookies fallback remains for the unlikely case the page session's `Network.getCookies` is also unavailable (e.g., browser closing mid-snapshot).

**Why the hostname gate (not type gate):** v143 PlzDedicatedWorker reports the worker fetch with `resourceType: 'Other'`, not `'Script'`. A strict type check would miss the case. Gating on `allowedHostnames` scopes the fallback to URLs the snapshot would have captured anyway, regardless of Chrome's internal classification.

### 1.6. HTML attribute-value serialization escapes `<` and `>` (Chrome 128+)

**What changed:** Chrome 128's new-headless serializes HTML attribute values per the [HTML5 spec](https://html.spec.whatwg.org/multipage/parsing.html), entity-escaping `<` and `>`. So `<iframe srcdoc="<p>Foo</p>">` serializes as `<iframe srcdoc="&lt;p&gt;Foo&lt;/p&gt;">`. Both render identically.

**How it broke Percy:** One test (`runs the execute callback in the correct frame`) asserted the literal `<p>Foo</p>` in serialized iframe srcdoc.

**What we did:** Widened the regex to accept either form. No production code change.

### 1.7. Chrome 128+ auto-fetches `/favicon.ico` on every navigation

**What changed:** New headless mode auto-fetches favicons. Old headless never did.

**How it broke Percy:** Several discovery tests asserted exact `server.requests` counts or `captured[i]` indices. Phantom `/favicon.ico` entries shifted everything.

**What we did:** Two test-helper changes (`test/helpers/server.js`):
1. Filter `/favicon.ico` out of the `server.requests` log
2. Reply HTTP 204 by default for `/favicon.ico` so it doesn't enter the snapshot resource list

Tests that explicitly want favicon capture (added in `890883ad`, `98f5a4e2`) register their own `server.reply('/favicon.ico', ...)` override.

> **Note for next commit:** the 204-default approach masks v128+ behavior instead of acknowledging it. A pending principled refactor will revert to a realistic test fixture and update affected tests to expect the auto-favicon-fetch in their assertions.

### 1.8. Browser-launch failure path varies by platform (Chrome 143)

**What changed:** Some args we used to test the "browser failed to launch" code path no longer cause Chrome to exit on all platforms. v143 Chrome with `--version` exits cleanly on Linux/macOS but keeps running on Windows; v126's pattern was Windows-friendly via `--remote-debugging-port=null`.

**What we did:** `browser.js` now passes a real `Error` to its rejection handler (one shape across code paths). The failing-launch test uses `--version` on Linux/macOS and `--remote-debugging-port=null` on Windows so each runner exercises the `handleExitClose` code path.

### 1.9. Auth dialog cancellation in headless (fundamental v143, NOT fixed)

**Observed:** When Chrome 143 cancels an HTTP basic auth dialog in headless mode, the page's `load` event never fires. `Page.navigate({waitUntil:'load'})` times out at 30s.

**Impact:** The `does not capture without auth credentials` test still takes ~1m34s in v143 (vs ~1s in v126). Test still passes; just slowly.

**Why not fixed:** This is fundamental Chrome 143 headless behavior — there's no graceful path on Percy's side. Real customers running asset discovery against auth-protected URLs without credentials would hit the same delay. Documented as known.

### 2.1. CI cache-key bump

`.github/.cache-key` counter incremented. Forces CI to re-download the v143 Chromium binary instead of reusing the cached v126.

> **⚠️ Note for future Chromium upgrades:** This bump was missed for v96, v123, and v126 upgrades in prior PRs. CI for those upgrades may have silently run against the cached old binary, producing false-green CI signals. Going forward this should be a mandatory step. Consider a pre-commit hook or PR template checklist item.

### 2.2. Updated `sec-ch-ua` header

In `network.js`, the hardcoded `sec-ch-ua` HTTP request header bumped from v123 → v143. Used only by `makeDirectRequest` (direct HTTP for fonts and the new worker fallback). Header should match the bundled binary's version.

> **Deferred:** Reviewer ce:review (P2 #6) flagged that the version literal will silently desync on the next Chrome bump. The whole `makeDirectRequest` header block has multiple hardcoded values (`sec-fetch-dest: 'font'` is wrong for worker scripts; `sec-ch-ua-platform: '"macOS"'` is wrong on Linux/Windows). Deferring a full audit rather than fixing one literal in isolation.

### 2.3. `install.js` revisions

Bumped the per-platform Chromium revision numbers in `install.js`. Each revision is the nearest [chromium-snapshots](https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html) entry to v143.0.7499.169 for that platform.

Documented the picking-procedure in inline comments so future upgrades don't have to re-derive it.

## Section 3 — References

- **Chrome release notes:** [v143](https://developer.chrome.com/release-notes/143), [v130 LNA](https://developer.chrome.com/blog/local-network-access), [Headless rewrite](https://developer.chrome.com/blog/removing-headless-old-from-chrome)
- **PlzDedicatedWorker:** [Chromium issue 40093136](https://issues.chromium.org/issues/40093136), [Playwright issue #31747](https://github.com/microsoft/playwright/issues/31747), [Playwright PR #35692](https://github.com/microsoft/playwright/pull/35692), [Puppeteer PR #13752](https://github.com/puppeteer/puppeteer/pull/13752)
- **Chromium source (verified — see §1.5):**
  - [`content/browser/worker_host/worker_script_fetcher.cc`](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/worker_host/worker_script_fetcher.cc) — `NetworkServiceDevToolsObserver` bound to the page frame's devtools token
  - [`content/browser/devtools/devtools_instrumentation.cc`](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/devtools_instrumentation.cc) — `OnWorkerMainScriptRequestWillBeSent` / `OnWorkerMainScriptLoadingFailed` present; **no `OnWorkerMainScriptResponseReceived`**
  - [`content/browser/devtools/dedicated_worker_devtools_agent_host.cc`](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/dedicated_worker_devtools_agent_host.cc) — `AttachSession` creates the worker session's NetworkHandler that later receives `loadingFinished`
- **CDP Fetch domain:** https://chromedevtools.github.io/devtools-protocol/tot/Fetch/


packages/core/test/helpers/server.js
```diff
@@ -28,6 +28,12 @@ export function createTestServer({ default: defaultReply, ...replies }, port = 8
   server.route(async (req, res, next) => {
     let pathname = req.url.pathname;
     if (req.url.search) pathname += req.url.search;
+    // Chrome >=128 new headless auto-requests /favicon.ico for every navigation.
+    // Reply 204 so the browser doesn't capture it as a snapshot resource, and
+    // skip it from the requests log so per-test request assertions stay stable.
+    if (req.url.pathname === '/favicon.ico') {
```
this-is-shivamsingh: How do we know if it's not requested from the page, but only requested from the header, Favicon appears in the tab header under <head> tag, can we distinguish if it's a favicon.ico request from the header and not from the any page resource ?

packages/core/test/snapshot.test.js
```diff
@@ -1535,10 +1535,13 @@ describe('Snapshot', () => {
 
       await percy.idle();
 
+      // srcdoc HTML attribute serialization: pre-Chrome-128 left `<` `>` literal
+      // inside attribute values; Chrome >=128 entity-escapes them per HTML5 spec.
+      // Accept either form.
```
this-is-shivamsingh: can you write an example what it does, any eg, in the comment to understand it

packages/core/src/install.js
```diff
@@ -175,13 +175,13 @@ export function chromium({
...
-  linux: '1300309',
-  win64: '1300297',
-  win32: '1300295',
-  darwin: '1300293',
-  darwinArm: '1300314'
+  linux: '1536366',
+  win64: '1536376',
+  win32: '1536377',
+  darwin: '1536380',
+  darwinArm: '1536376'
```
this-is-shivamsingh, line 185: how do you got it this, can you attach a reference here ?

packages/core/src/browser.js
```diff
@@ -20,8 +20,15 @@ export class Browser extends EventEmitter {
...
-    // disable the translate popup and optimization downloads
-    '--disable-features=Translate,OptimizationGuideModelDownloading',
+    // disable: translate popup, optimization downloads, baseline site
+    // isolation (so cross-origin sub-resources and worker fetches stay in
+    // the main renderer for interception), HTTPS-first navigation blocking,
+    // and Local Network Access permission checks (Chrome 143+ blocks
+    // sub-resource requests to *.localhost / 127.0.0.1 / RFC1918 with
+    // `LocalNetworkAccessPermissionDenied` when the document was served
+    // via Fetch.fulfillRequest from asset discovery — there is no permission
+    // prompt available in headless to grant the access).
```
this-is-shivamsingh: can you simplify this, what it meant ? 
and why we have to disable IsolateOrigins

this-is-shivamsingh (commented):
commented :}}

packages/core/src/network.js
```diff
@@ -623,6 +728,32 @@ async function makeDirectRequest(network, request, session) {
...
+
+    let resource = createResource(url, body, mimeType, {
+      status: 200,
+      headers: { 'content-type': [mimeType] }
+    });
+
+    log.debug(`- Saving direct-fetched resource sha=${resource.sha} mimetype=${mimeType}`, meta);
+    network.intercept.saveResource(resource);
+  } catch (error) {
+    log.debug(`Direct fetch failed for ${url} - ${error.message}`, meta);
```
bhokaremoin, line 834: **[C1] CRITICAL — Coverage regression breaks CI.**

This line is the only uncovered line in `src/` (99.96% line / 99.89% branch / 99.97% statement vs. the repo's enforced **100% global threshold**), and CI is currently red on `Test @percy/core` because of it. The last commit on the branch (`2d48f20 test(core): drop CI-unstable direct worker-script direct-fetch spec`) explicitly removed the test that exercised this catch block with no replacement.

Two options:

**A. Re-add a deterministic test (preferred — matches the file's "test rather than ignore" convention).** The dropped spec was flaky because it relied on a real-network timeout. Stub `makeDirectRequest` (or `@percy/client/utils#request`) at module scope to reject synchronously, then assert `logger.stderr` contains `Direct fetch failed for …`.

**B. Annotate with `/* istanbul ignore next */` consistent with other race-only branches in this file:**
```js
} catch (error) {
  /* istanbul ignore next: direct-fetch throws only on real network errors;
     exercised manually but flaky in CI */
  log.debug(`Direct fetch failed for ${url} - ${error.message}`, meta);
}
```

This is the only blocking item — once green, the rest are non-blocking polish.

packages/core/src/network.js
```diff
@@ -262,6 +271,51 @@ export class Network {
...
 
+  // Response-stage interception is kept ONLY to detect oversized/malformed
+  // Content-Length and abort the request before Chrome streams a body it
+  // would never terminate (Chrome 143 quirk). For everything else we just
+  // continue — body capture happens later via Network.loadingFinished →
+  // Network.getResponseBody (the v126 path). Reading the body at this stage
+  // hangs worker-initiated fetches, so we don't.
+  _handleResponsePaused = async (session, event) => {
+    let { networkId: requestId, requestId: interceptId, responseHeaders, responseStatusCode } = event;
+    let request = this.#requests.get(requestId);
```
bhokaremoin, line 289: **[C2] Untracked-request branch deserves an inline comment.**

When `this.#requests.get(requestId)` returns `undefined` here, the function still proceeds — it inspects Content-Length, may call `Fetch.failRequest`, or falls through to `_continueResponse`. This is *correct* (Chrome has the response paused, so we must un-pause it one way or the other), and the new `aborts oversized responses for untracked requests` spec covers the behavior.

But a future reader has no explanation of *when* the untracked case actually happens in production — service workers? redirects? a race between `Network.requestWillBeSent` and the response-stage pause? — and the comment block above this function doesn't address it.

Suggested addition right above this line:

```js
// The request may be untracked (#requests.get(requestId) === undefined) when a
// response-stage paused event arrives for a request whose request-stage
// tracking we never installed — most commonly seen with service-worker-fulfilled
// responses or after a race with cleanup. We still need to unpause Chrome, so
// we either fail (oversized/malformed) or continue regardless.
```

packages/core/src/network.js
```diff
@@ -262,6 +271,51 @@ export class Network {
...
+    let url = request ? originURL(request) : event.request?.url;
```
bhokaremoin: **[C3] URL normalization inconsistency.**

When `request` is tracked, `url` is normalized (via `originURL` → `normalizeURL`). When untracked, `url` is the raw `event.request.url`. The resulting `url` is then used in log meta, `asset_not_uploaded` instrumentation telemetry, and error messages.

Two requests for the same underlying resource — one tracked, one not — will produce different `url` strings in telemetry. Likely observable only in logs/instrumentation, but inconsistent with the rest of the module which always normalizes.

Cheap fix:

```js
let url = request ? originURL(request) : (event.request?.url && normalizeURL(event.request.url));
```

packages/core/src/browser.js
```diff
@@ -21,8 +21,11 @@ export class Browser extends EventEmitter {
   #lastid = 0;
 
   args = [
-    // disable the translate popup and optimization downloads
-    '--disable-features=Translate,OptimizationGuideModelDownloading',
+    // Disable Chrome features that break asset discovery in v143 new-headless:
+    // site/origin isolation (so cross-origin events stay on the page session),
+    // HTTPS-first auto-upgrade (would block HTTP discovery), and Local Network
+    // Access permission checks (would block sub-resources to localhost/RFC1918).
+    '--disable-features=Translate,OptimizationGuideModelDownloading,IsolateOrigins,site-per-process,HttpsFirstBalancedModeAutoEnable,LocalNetworkAccessChecks',
```
bhokaremoin: **[C4] Extract `--disable-features=` into a named constant.**

The disable list is now 6 items / ~130 chars on one line. Each item exists for a different reason: Translate / OptimizationGuide are pre-v126; IsolateOrigins / site-per-process are Chrome 128 site-isolation; HttpsFirstBalancedModeAutoEnable / LocalNetworkAccessChecks are Chrome 143 quirks.

The rest of the `args = [...]` array uses the per-flag inline-comment style. Matching that here will make the next Chrome upgrade meaningfully easier — you'd be able to see at a glance which flag is no longer needed and remove it surgically.

Suggested:

```js
const DISABLED_FEATURES = [
  'Translate',                          // disable translate popup (pre-v126)
  'OptimizationGuideModelDownloading',  // disable background model downloads (pre-v126)
  'IsolateOrigins',                     // v128: keep cross-origin sub-resources on the page session
  'site-per-process',                   // v128: strict superset of IsolateOrigins
  'HttpsFirstBalancedModeAutoEnable',   // v143: prevent silent HTTP→HTTPS upgrade in headless
  'LocalNetworkAccessChecks'            // v143: allow sub-resources to localhost / RFC1918
];

args = [
  `--disable-features=${DISABLED_FEATURES.join(',')}`,
  // …
];
```

packages/core/test/percy.test.js
```diff
@@ -542,11 +542,11 @@ describe('Percy', () => {
...
           url: 'http://localhost:8000/snapshot-1'
         }),
-        // should not upload
+        // should not upload — bumped from 100ms to 1000ms so snapshot-1's
+        // upload reliably fails first even when CI runs the suite slowly
         percy.snapshot({
           url: 'http://localhost:8000/snapshot-2',
-          // delay this snapshot so the first upload can fail
-          waitForTimeout: 100
+          waitForTimeout: 1000
```
bhokaremoin, line 549: **[C5] 10× timeout bump is a hammer — prefer a deterministic resolver.**

Bumping `waitForTimeout` 100ms → 1000ms to "make a timing race reliable" adds 900ms of test runtime for the same logical coverage and just pushes the race further out — if CI is slow enough, it could flake again.

A deterministic approach (e.g., await a promise that resolves when snapshot-1's upload failure has been observed, *then* start snapshot-2) would be more durable. The PR already uses exactly this pattern in `discovery.test.js` — the `firstCallResolver` change to the validation-mock test. Same shape would work here.

Not a blocker — happy to defer to a follow-up if you'd rather keep the surface area of this PR small.

packages/core/test/helpers/server.js
```diff
@@ -28,8 +28,14 @@ export function createTestServer({ default: defaultReply, ...replies }, port = 8
   server.route(async (req, res, next) => {
     let pathname = req.url.pathname;
     if (req.url.search) pathname += req.url.search;
-    server.requests.push(req.body ? [pathname, req.body, req.headers] : [pathname, req.headers]);
     let reply = replies[pathname] || defaultReply;
+    // Chrome >=128 auto-fetches /favicon.ico on every navigation; reply 204
+    // by default so it doesn't pollute snapshot resources. Tests can still
+    // override via `server.reply('/favicon.ico', ...)`.
+    if (req.url.pathname === '/favicon.ico' && !replies['/favicon.ico']) {
```
bhokaremoin, line 35: **[C6] Move the "this is a band-aid" follow-up commitment into source.**

The PR body explicitly calls this out as temporary and lists a follow-up to "revert the favicon 204-default … and update affected tests to acknowledge v128+ auto-favicon-fetch." That commitment lives only in the PR description — once merged, the rationale will be hard to find and the cleanup will likely never happen.

A `TODO(PPLT-4214 follow-up):` (or whatever the parent Jira is) directly above this branch makes the commitment discoverable from `git grep`:

```js
// TODO(PPLT-4214 follow-up): Chrome >=128 auto-fetches /favicon.ico on every
// navigation. Replying 204 here is a band-aid so existing test assertions on
// server.requests / captured[i] keep their indices. The principled fix is to
// remove this branch and update affected tests to expect favicon entries.
if (req.url.pathname === '/favicon.ico' && !replies['/favicon.ico']) {
  return res.writeHead(204).end();
}
```

packages/core/src/network.js
```diff
@@ -479,6 +555,24 @@ function originURL(request) {
...
+  for (let { name, value } of arr) out[name] = value;
+  return out;
+}
+
+// Returns { tooLarge, malformed, rawValue } for Content-Length classification.
+function inspectContentLength(headers) {
+  let key = headers && Object.keys(headers).find(k => k.toLowerCase() === 'content-length');
+  let rawValue = key ? headers[key] : undefined;
+  let parsed = parseInt(rawValue);
+  let tooLarge = Number.isFinite(parsed) && parsed > MAX_RESOURCE_SIZE;
```
bhokaremoin, line 599: **[C7] Pass an explicit radix to `parseInt`.**

Functionally fine in modern Node (defaults to 10), but inline with the rest of the codebase's convention — `grep -n "parseInt(.*, 10)"` shows the repo's habit is to be explicit. Lint-noisy under some configs and a 2-character fix:

```js
let parsed = parseInt(rawValue, 10);
```

packages/core/src/network.js
```diff
@@ -262,6 +271,51 @@ export class Network {
...
+      });
+      this.log.debug('- Skipping resource larger than 25MB', meta);
+      if (request) {
+        this._forgetRequest(request);
+        this.#requestsLifeCycleHandler.get(requestId).resolveResponseReceived();
+      }
+      try {
+        await this.send(session, 'Fetch.failRequest', { requestId: interceptId, errorReason: 'Aborted' });
+      } catch (error) {
+        this.log.debug(`Failed to abort oversized response for ${url}: ${error.message}`);
```
bhokaremoin: **[C8] Log `Fetch.failRequest` errors at `warn`, not `debug`, for non-race failures.**

When `Fetch.failRequest` fails inside `_handleResponsePaused`, Chrome may still have the response paused indefinitely. The current `log.debug` is invisible at the default loglevel — so an operator running into a stuck-snapshot incident in production has no signal whatsoever pointing at this code path. The new `Failed to abort oversized response for …` test will pass either way.

Suggest distinguishing known races (`ABORTED_MESSAGE`, `Invalid InterceptionId`) from unexpected failures:

```js
try {
  await this.send(session, 'Fetch.failRequest', { requestId: interceptId, errorReason: 'Aborted' });
} catch (error) {
  if (error.message === ABORTED_MESSAGE || error.message.includes('Invalid InterceptionId')) {
    this.log.debug(`Aborted-race ignored for ${url}: ${error.message}`);
  } else {
    this.log.warn(`Failed to abort oversized response for ${url}: ${error.message}`);
  }
}
```

packages/core/src/network.js
```diff
@@ -7,6 +7,8 @@ const MAX_RESOURCE_SIZE = 25 * (1024 ** 2) * 0.63; // 25MB, 0.63 factor for acco
 const ALLOWED_STATUSES = [200, 201, 301, 302, 304, 307, 308];
 const ALLOWED_RESOURCES = ['Document', 'Stylesheet', 'Image', 'Media', 'Font', 'Other'];
 const ABORTED_MESSAGE = 'Request was aborted by browser';
+// Chrome 143 omits Network.responseReceived for worker scripts; cap the wait so loadingFinished can clean up.
+const RESPONSE_RECEIVED_TIMEOUT = 2000;
```
bhokaremoin, line 13: **[C9] Note the cumulative-timeout worst case in the comment.**

Per the PR description, the 2s cap is per-request. For a synthetic page with N worker-script-like requests all missing `responseReceived`, the worst-case extra discovery latency is `N * RESPONSE_RECEIVED_TIMEOUT` (2N seconds). This won't happen on real customer pages — typical N is 1–2 — but the assumption is invisible to future readers.

Suggested expansion:

```js
// Chrome 143 omits Network.responseReceived for worker scripts; cap the wait
// so loadingFinished can clean up. Per-request: a page with N missing
// responseReceived events accumulates up to N * RESPONSE_RECEIVED_TIMEOUT
// of extra discovery latency — acceptable since the real-world N is ~1–2.
const RESPONSE_RECEIVED_TIMEOUT = 2000;
```

packages/core/src/network.js
```diff
@@ -593,17 +689,26 @@ async function sendResponseResource(network, request, session) {
   }
 }
 
-// Make a new request with Node based on a network request
+// Make a new request with Node based on a network request. The session is
+// usually the page session (full Network domain support), but for the v143
+// worker-script fallback the loadingFinished event arrives on the worker
+// session where Network.getCookies returns an Internal error — proceed with
+// no cookies in that case rather than abandoning the resource.
 async function makeDirectRequest(network, request, session) {
```
bhokaremoin, line 763: **[C10] Mirror the cookie-less-fallback caveat into user-facing release notes.**

For PlzDedicatedWorker scripts that hit the timeout fallback, the direct-fetch will run with empty cookies whenever `Network.getCookies` throws on the worker session. The PR body's section 1.5 acknowledges this — but the trade-off needs to land in user-visible release notes so customers running asset discovery against **auth-gated dedicated-worker scripts** aren't surprised by captured bytes diverging from what Chrome's loader would have fetched.

One-line note for the next `@percy/core` release / CHANGELOG:

> v143 upgrade: dedicated-worker scripts captured via the timeout fallback path do not include page cookies. Customers running asset discovery against auth-gated worker scripts should reach out if they observe diffs.

packages/core/src/install.js
```diff
@@ -175,13 +175,14 @@ export function chromium({
...
+// revision from https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html
 chromium.revisions = {
-  linux: '1300309',
-  win64: '1300297',
-  win32: '1300295',
-  darwin: '1300293',
-  darwinArm: '1300314'
+  linux: '1536366',
+  win64: '1536376',
+  win32: '1536377',
```
bhokaremoin, line 183: **[C11] Confirm darwin/darwinArm ordering wasn't accidentally swapped.**

All revisions cluster tightly around base position 1536371 (the v143.0.7499.169 base), but two stand out:

- `darwin: 1536380` (Intel macOS) — +9 from base
- `darwinArm: 1536376` (ARM macOS) — +5 from base
- `linux: 1536366` is the only one *below* the base position

Just want to confirm each was independently picked as "closest available snapshot to v143.0.7499.169 for the platform" and not, e.g., Intel/ARM accidentally swapped. The darwin/darwinArm asymmetry isn't impossible — Chromium-snapshots availability differs per platform — but it's the kind of thing worth a quick double-check against the snapshot index linked in the comment above.

packages/core/src/network.js
```diff
@@ -351,12 +405,34 @@ export class Network {
...
+
+    if (!request.response) {
+      this.log.debug(`Skipping resource: responseReceived not received within ${RESPONSE_RECEIVED_TIMEOUT}ms - ${request.url}`);
+      // Chrome 143+ fetches dedicated worker scripts in the browser process
+      // (PlzDedicatedWorker) and never surfaces the response on any CDP
+      // session. resourceType varies — v143 reports 'Other' for the worker
+      // script, older Chrome reports 'Script' — so we don't gate on type.
+      // We do gate on allowedHostnames so we never direct-fetch a host the
+      // snapshot wouldn't have captured anyway.
+      if (hostnameMatches(this.intercept.allowedHostnames, originURL(request))) {
```
bhokaremoin: **[C12] Silent skip when `allowedHostnames` doesn't match a worker host — intentional?**

When the `responseReceived` timer fires and the URL is *not* on `allowedHostnames`, the code logs the generic timeout debug line above and then forgets the request — no direct-fetch fallback, no distinct log signal.

For customers running with `allowedHostnames` left at its default (effectively unset / empty), `hostnameMatches` would return false for everything, and so worker scripts on remote CDNs would silently disappear from captures without any indication of *why*. The pre-v143 path would have captured them as long as `enableJavaScript: true`.

Two small asks:

1. Confirm this is the intended behavior — i.e., it's acceptable that a default-config user loses worker-script captures from cross-origin CDNs after this upgrade.
2. If yes, add a distinct debug log so future debugging can distinguish "timed out and we're skipping intentionally" from the generic timeout:

```js
if (hostnameMatches(this.intercept.allowedHostnames, originURL(request))) {
  await captureScriptDirectly(this, request, session);
} else {
  this.log.debug(`Skipping direct-fetch fallback for ${request.url} — host not in allowedHostnames`);
}
```

bhokaremoin (changes requested):
Reviewed end-to-end — the diagnostics, Chromium-source citations, and per-feature rationale in this PR are well above the bar, and the dual-stage Fetch interception + worker-script direct-fetch fallback are well-considered. Twelve findings inline:

**Blocking**
- **C1** — CI is red. `network.js:753` (the `captureScriptDirectly` catch-block `log.debug`) is the only uncovered line in `src/` after `2d48f20` dropped its test. The repo enforces 100% global coverage. Either re-add the test (stubbing `makeDirectRequest`) or `/* istanbul ignore next */` the catch consistent with this file's pattern.

**Worth addressing in this PR**
- **C2** — Add an inline comment in `_handleResponsePaused` explaining when `#requests.get(requestId)` returns `undefined` (SW-fulfilled? redirect race?). The behavior is correct; the rationale is invisible.
- **C3** — URL normalization mismatch in `_handleResponsePaused`: tracked path uses `originURL` (normalized), untracked falls back to raw `event.request.url`. Telemetry will see two different URLs for the same resource.
- **C4** — Extract the 6-feature `--disable-features=` list into a named `DISABLED_FEATURES` array with one comment per flag noting the Chrome version that introduced the need. Will pay off massively at the next Chrome upgrade.
- **C6** — Move the favicon "band-aid" follow-up commitment from the PR body into a `TODO(PPLT-4214 follow-up):` comment in `server.js` so the cleanup is discoverable.
- **C8** — `Fetch.failRequest` catch logs at `debug` — invisible at default loglevel. A stuck-snapshot production incident would have no signal. Log non-race errors at `warn`.
- **C11** — Confirm `darwin: 1536380` vs `darwinArm: 1536376` weren't swapped; the asymmetry is plausible but worth a quick double-check against the snapshot index.
- **C12** — Confirm that silently skipping worker scripts from non-`allowedHostnames` hosts is intended (default-config users will lose CDN worker captures after this upgrade vs. v126). If yes, add a distinct debug log so the skip is observable.

**Polish / defer to follow-up**
- **C5** — 100ms → 1000ms `waitForTimeout` bump in `percy.test.js` papers over a timing race rather than fixing it. The `firstCallResolver` pattern used elsewhere in this PR would be deterministic.
- **C7** — `parseInt(rawValue)` → `parseInt(rawValue, 10)` to match repo convention.
- **C9** — One-line expansion of the `RESPONSE_RECEIVED_TIMEOUT` comment documenting the N × 2s cumulative-timeout worst case.
- **C10** — Mirror the cookie-less direct-fetch caveat into user-facing release notes so customers running auth-gated dedicated workers aren't surprised by diffs.

Overall: solid work. Once C1 is resolved and the C2–C4/C6/C8/C11/C12 items are addressed (or explicitly deferred with rationale), this is ready to merge.

packages/core/src/network.js
Manoj-Katta: added the normalizeURl

packages/core/src/install.js
Manoj-Katta: attached the ref

packages/core/src/install.js
Manoj-Katta: No the ordering is correct verified it from the source
https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html

packages/core/src/browser.js
Manoj-Katta: - from Chrome 128 onward, the new-headless browser puts cross-origin sub-resources into a different renderer process than the page that requested them. 
- CDP delivers Network.* events on the session attached to the process which will be missed since they are on different process and can't be identified by the page that requested them.
- `--disable-features=IsolateOrigins,site-per-process` keeps all sub-resources on the page session so the existing Network.* interception keeps working

packages/core/src/install.js
Manoj-Katta: in the comment

packages/core/test/helpers/server.js
Manoj-Katta: - This change is only made for the test file
- The actual working will be same. If thats in the dom then the percy will automatically capture it if its in allowedHostnames etc and for the tests I did like this because:
  - pre-128 Chrome only fetched favicons  when explicitly declared, so a lot of existing specs assert on positional indices (server.requests[0], captured[1], …). 
  - v128+ silently inserts a /favicon.ico request and shifts every one of those, breaking unrelated tests. The 204 default keeps those legacy assertions stable in this PR.
  - The actual CLI behaviour (whether Percy captures /favicon.ico for a customer's site) is covered explicitly by two new specs in discovery.test.js
     - captures favicon when the server provides one (discovery.test.js:463) — page declares <link rel="icon">, asserts capture works.
     - captures auto-fetched favicon when the page does not declare one (discovery.test.js:484) — page has no <link>, relies on v143's browser auto-fetch, asserts capture works.
  - So we have positive coverage for the real behaviour and a stable test-fixture for the rest of the suite.

packages/core/test/helpers/server.js
Manoj-Katta: - the 204 onlyis in test/helpers/server.js — a test-only fixture, never published. It exists because pre-128 Chrome only fetched favicons when explicitly declared, so many specs assert on positional indices (server.requests[0], captured[1], …). 
- v128+ silently inserts a /favicon.ico request and shifts every one of those, breaking unrelated tests. The 204 default keeps those legacy assertions stable.
- The actual customer-facing behaviour (whether Percy captures /favicon.ico in a real build) has positive coverage in discovery.test.js — both specs bypass the 204 default by registering an explicit reply:
  - captures favicon when the server provides one (discovery.test.js:463) — page declares <link rel="icon">, asserts capture works.
  - captures auto-fetched favicon when the page does not declare one (discovery.test.js:484) — page has no <link>, relies on v143's browser auto-fetch, asserts capture works.

- So we have explicit load-bearing coverage for the real behaviour and a stable fixture for the rest of the suite. 
-

packages/core/src/network.js
Manoj-Katta: Inline comment added explaining the untracked-request fall-through in _handleResponsePaused.

packages/core/src/network.js
Manoj-Katta: Fixed in f0d9c8e7 — explicit radix 10 at both parseInt call sites.

packages/core/src/network.js
Manoj-Katta: Attempted in f0d9c8e7, reverted in 334f0621 — promoting Fetch.failRequest errors to warn produced spurious warn lines on routine Session closed / Target closed lifecycle teardowns. The right home for this signal is instrumentation telemetry rather than user-facing logs; tracking that as a follow-up.

packages/core/src/network.js
Manoj-Katta: Comment added in f0d9c8e7 noting that the 2s RESPONSE_RECEIVED_TIMEOUT is per-request and calling out the cumulative worst case for synthetic pages with many worker scripts.

packages/core/src/browser.js
Manoj-Katta: Deferring — the existing block-level comment is adequate context for the six items. Happy to extract a `DISABLED_FEATURES` constant as a follow-up if the list grows.

packages/core/test/percy.test.js
Manoj-Katta: Deferring per your pre-approval — refactoring to a deterministic resolver is better as a focused follow-up than bundled into this Chromium upgrade.

packages/core/src/network.js
Manoj-Katta: Will add a CHANGELOG entry for the cookie-less worker-script direct-fetch fallback edge case at release time.

packages/core/src/network.js
Manoj-Katta: Made explicit in `f0d9c8e7` — `disallowedHostnames`-before-`allowedHostnames` precedence now mirrors `sendResponseResource`, and a debug log fires when the fallback is declined for hostname reasons. A coverage spec for the skip branch lands shortly with the C1 fix.

packages/core/src/network.js
Manoj-Katta: Fixed in `95266618` — two new specs in `discovery.test.js` cover the `Network.getCookies` failure catch and the 25 MB direct-fetch body-size guard. The third uncovered branch (hostname-not-allowed skip in `_handleLoadingFinished`) only fires for PlzDedicatedWorker requests whose worker-script fetch bypasses `Fetch.requestPaused` — annotated with `/* istanbul ignore else */` plus a justification comment, since integration tests can't reliably reach that branch. Coverage gate should be back to 100% on the new run.

packages/core/src/install.js
```diff
@@ -175,13 +175,14 @@ export function chromium({
...
+  darwin: '1536380',
+  darwinArm: '1536376'
```
amandeepsingh333, line 185: **Likely copy-paste error — `darwinArm` matches `win64`.**

```js
win64: '1536376',
win32: '1536377',
darwin: '1536380',
darwinArm: '1536376'   // ← same as win64
```

Chromium snapshots are per-platform; the ARM macOS build has its own revision. Having `darwinArm === win64` is statistically improbable and would result in Apple Silicon runners pulling a Windows binary (which then fails to extract / launch).

Double-check against [chromium-browser-snapshots](https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Mac_Arm/) for the closest Mac_Arm/ revision to 143.0.7499.169. Also worth noting in the comment which revision corresponds to which exact Chromium tag, so the next upgrader doesn't have to re-derive it for 5 platforms.

packages/core/src/network.js
```diff
@@ -593,17 +716,59 @@ async function sendResponseResource(network, request, session) {
...
   let headers = {
     // add default browser
     accept: '*/*',
     'sec-fetch-site': 'same-origin',
     'sec-fetch-mode': 'cors',
     'sec-fetch-dest': 'font',
-    'sec-ch-ua': '"Chromium";v="123", "Google Chrome";v="123", "Not?A_Brand";v="99"',
+    'sec-ch-ua': '"Chromium";v="143", "Google Chrome";v="143", "Not?A_Brand";v="99"',
     'sec-ch-ua-mobile': '?0',
     'sec-ch-ua-platform': '"macOS"',
```
amandeepsingh333, line 780: **`sec-ch-ua-platform` hardcoded to `"macOS"` is sent from Linux and Windows runners.**

```js
'sec-ch-ua-platform': '"macOS"',
```

Servers that do content negotiation on Client Hints (rare but real — CDNs, paywall vendors, font services) will serve the wrong assets for the platform actually doing the discovery. Linux/Windows CI runs would pull macOS-specific resources.

PR body marks this "deferred" — fair, but it's a behavioral regression vs. v126 because v143's new headless is more spec-compliant about Client Hints, so origin servers are more likely to act on them. At minimum, derive at module load from `process.platform`:

```js
const SEC_CH_UA_PLATFORM = ({darwin: '"macOS"', win32: '"Windows"'})[process.platform] || '"Linux"';
```

Same argument applies to `sec-fetch-dest: 'font'` (L770) which is incorrect for the new worker-script fallback path; the destination should be `worker` (or omitted) when the direct-fetch came from `captureResourceDirectly`.

packages/core/src/network.js
```diff
@@ -613,14 +778,52 @@ async function makeDirectRequest(network, request, session) {
...
+    if (body.length > MAX_RESOURCE_SIZE) {
+      logAssetInstrumentation(log, 'asset_not_uploaded', 'resource_too_large', {
+        url, size: body.length, snapshot: meta.snapshot
+      });
+      log.debug('- Skipping resource larger than 25MB', meta);
+      return;
+    }
+
+    let urlObj = new URL(url);
+    let mimeType = mime.lookup(urlObj.origin + urlObj.pathname) || 'application/javascript';
```
amandeepsingh333: **Mimetype fallback `'application/javascript'` masks non-JS resources.**

```js
let mimeType = mime.lookup(urlObj.origin + urlObj.pathname) || 'application/javascript';
```

`captureResourceDirectly` is reachable for **any** allowlisted resource that hit the `responseReceived` timeout, not just worker scripts (PR body acknowledges this). For extensionless URLs (`/api/data`, REST-style worker registries), `mime.lookup()` returns false and the resource gets stored as JS — then the renderer chokes when it actually receives JSON/HTML/binary content with that mimetype.

Safer default: use the HTTP `Content-Type` response header from the direct fetch (you already have access to `res` via the `makeRequest` callback you added). Fall back to `application/octet-stream` if both `mime.lookup` and the header are absent. JS is the wrong default — it's the *most* likely to cause downstream parse errors when wrong.

packages/core/src/network.js
```diff
@@ -262,6 +274,68 @@ export class Network {
...
+    }
+
+    return this._continueResponse(session, interceptId, url);
+  }
+
+  // Tell the browser to continue the paused response, swallowing expected
+  // races (request already aborted, interception ID no longer valid).
+  _continueResponse = async (session, interceptId, url) => {
+    try {
+      await this.send(session, 'Fetch.continueResponse', { requestId: interceptId });
```
amandeepsingh333, line 333: **PR description vs. code mismatch on observability.**

PR body says:
> Production observability via `log.warn`. Unknown errors in `_handleResponsePaused`'s `failRequest` catch and `_continueResponse`'s catch now log at `warn` (was `debug`).

But here:

```js
_continueResponse = async (session, interceptId, url) => {
  try {
    await this.send(session, 'Fetch.continueResponse', { requestId: interceptId });
  } catch (error) {
    if (error.message === ABORTED_MESSAGE || error.message.includes('Invalid InterceptionId')) return;
    this.log.debug(`Failed to continue response for ${url}: ${error.message}`);  // ← debug, not warn
  }
}
```

…and the corresponding `_handleResponsePaused` catches at L312/L316/L320 are also `log.debug`. Either the description is stale or the code is. Given the PR body explicitly justifies why customers debugging stuck-snapshot incidents need this signal at the default loglevel, code should match. If `log.warn` produces too much noise in practice, push that decision into the description with reasoning.

packages/core/src/network.js
```diff
@@ -593,17 +716,59 @@ async function sendResponseResource(network, request, session) {
...
+// fall back to the request's own session otherwise.
+export function pickCookieSession(network, session) {
+  return network.page?.session ?? session;
+}
+
+// Decide whether to attach a Basic auth header to the Node-side direct fetch.
+// The browser's URLLoader origin-scopes Basic auth; this fallback runs in
+// Node, so we re-enforce the same-origin rule explicitly to avoid leaking
+// credentials cross-origin. Malformed URLs fall through to `false` defensively.
+export function shouldAttachAuth(authorization, requestUrl, snapshotUrl) {
```
amandeepsingh333, line 732: **`shouldAttachAuth` returns false when `snapshotURL` is missing — confirm meta is populated in every entry path.**

```js
export function shouldAttachAuth(authorization, requestUrl, snapshotUrl) {
  if (!authorization?.username) return false;
  try {
    return new URL(requestUrl).origin === new URL(snapshotUrl).origin;
  } catch {
    return false;
  }
}
```

`new URL(undefined)` throws → catch → false. This is correct *defensively* but silently strips Basic auth when `network.meta.snapshotURL` happens to be undefined. The old code unconditionally attached auth based on `authorization?.username` alone, so this is a behavior change for the font-fallback path (which has shipped for many releases).

Worth verifying: is `meta.snapshotURL` set on every code path that constructs a `Network` instance? The DOM-snapshot internal flows, the SDK paths, the `percy snapshot` CLI — each may or may not populate it. If any of them leaves it undefined, existing customers with authenticated origins will silently see fetches drop their `Authorization` header in this release. A short sanity-check test on the entry points + a TODO log when `snapshotUrl` is missing would catch this.

packages/core/src/browser.js
amandeepsingh333: **Comma-bundled `--disable-features` list mixes a security boundary with behavior knobs — worth splitting and annotating.**

```js
'--disable-features=Translate,OptimizationGuideModelDownloading,IsolateOrigins,site-per-process,HttpsFirstBalancedModeAutoEnable,LocalNetworkAccessChecks',
```

Three concerns on one line:

1. **One typo and the whole flag silently no-ops.** Chrome doesn't validate feature names; an unrecognized entry is ignored without warning. Splitting into a `DISABLED_FEATURES` array joined with `.join(',')` makes each entry diff-able and grep-able.

2. **`IsolateOrigins,site-per-process` is a process-isolation security boundary**, not a UI/networking knob. Disabling it is fine for *headless* asset discovery where we control the input URL, but a future contributor copy-pasting this pattern to a non-headless code path (e.g., a debug/interactive command) would silently weaken Spectre/Meltdown protections. A comment per disabled feature explaining the threat model would prevent that.

3. **PR body says "matches v126 old-headless behavior"** — yes, but v126 old-headless was also the unsafe-by-default baseline; we're restoring an unsafe-by-default baseline in v143 by choice. Worth an explicit acknowledgment so when Chrome 150 changes something else, the next person knows we deliberately keep this stance.

Suggested shape:

```js
const DISABLED_FEATURES = [
  'Translate',                          // popup
  'OptimizationGuideModelDownloading',  // background download noise
  'IsolateOrigins',                     // [headless-only] keep cross-origin in same renderer for CDP capture
  'site-per-process',                   // [headless-only] same reason
  'HttpsFirstBalancedModeAutoEnable',   // allow HTTP discovery against local servers
  'LocalNetworkAccessChecks',           // allow loopback/RFC1918 sub-resources
];
args = [`--disable-features=${DISABLED_FEATURES.join(',')}`, ...]
```

packages/core/src/network.js
```diff
@@ -7,6 +7,11 @@ const MAX_RESOURCE_SIZE = 25 * (1024 ** 2) * 0.63; // 25MB, 0.63 factor for acco
 const ALLOWED_STATUSES = [200, 201, 301, 302, 304, 307, 308];
 const ALLOWED_RESOURCES = ['Document', 'Stylesheet', 'Image', 'Media', 'Font', 'Other'];
 const ABORTED_MESSAGE = 'Request was aborted by browser';
+// Chrome 143 omits Network.responseReceived for worker scripts; cap the wait
+// so loadingFinished can clean up. Per-request — N timeouts accumulate to N*2s; real-world N is ~1–2.
+const RESPONSE_RECEIVED_TIMEOUT = 2000;
+// Cap idle() impact when a host accepts the TCP connection then stalls during a direct fetch.
```
amandeepsingh333, line 14: **`N*2s` worst-case has no hard ceiling — worth a per-snapshot guard.**

```js
const RESPONSE_RECEIVED_TIMEOUT = 2000;
```

The comment acknowledges "N timeouts accumulate to N*2s; real-world N is ~1–2." For typical pages that's fine. But:

- Pages with many failed worker fetches (a single misconfigured framework can register N workers).
- A future Chromium change that triggers this timeout for any *other* request type — the same code path now applies, so the failure mode is broader than just worker scripts.

A defensive cap would be: track time-spent-in-responseReceived-timeouts on the Network instance; if it exceeds, say, `min(networkIdleWaitTimeout / 2, 10s)`, stop awaiting `responseReceived` entirely and just take the loadingFinished as truth. The cost is a few cases that would have completed normally now skip the body; the benefit is bounded latency.

Alternatively, document that PERCY_NETWORK_IDLE_WAIT_TIMEOUT effectively shifts both ceilings together and call that the safety valve.

Not a blocker — just worth thinking about now while the code is fresh, vs. waiting for a customer-with-a-thousand-workers incident.

packages/core/src/network.js
```diff
@@ -250,6 +255,13 @@ export class Network {
   _handleRequestPaused = async (session, event) => {
     let { networkId: requestId, requestId: interceptId, resourceType } = event;
 
+    // Response-stage events arrive here when Fetch.continueRequest was called
+    // with interceptResponse:true (see sendResponseResource).
+    if (event.responseStatusCode != null || event.responseErrorReason != null) {
```
amandeepsingh333, line 261: **The branch detects response-stage events by side-effect of fields, not by an explicit flag — fragile.**

```js
if (event.responseStatusCode != null || event.responseErrorReason != null) {
  await this._handleResponsePaused(session, event);
  return;
}
```

The CDP `Fetch.requestPaused` event has a well-defined disambiguator: when paused at the response stage, `responseStatusCode` is set (and the event represents a response pause); when paused at the request stage, it isn't. The check works.

But: a response-stage pause for a *failed* request can arrive with `responseStatusCode = null` and `responseErrorReason` set (network errors, blocked, etc.) — you handle that case via the `|| responseErrorReason != null` clause, good.

The fragility is that a future CDP revision could change one or both fields without breaking this branch (silent regression). Lower-risk-but-cleaner: read `event.responseStatusCode !== undefined` (explicit "this field is present in the wire format") rather than `!= null` (which also matches `0`, an unlikely-but-valid HTTP status from `Fetch.fulfillRequest`-style paths). Minor.

packages/core/src/network.js
```diff
@@ -74,7 +79,7 @@ export class Network {
     this.captureMockedServiceWorker = options.captureMockedServiceWorker ?? false;
     this.userAgent = options.userAgent ??
       // by default, emulate a non-headless browser
-      page.session.browser.version.userAgent.replace('Headless', '');
+      page.session.browser?.version?.userAgent?.replace('Headless', '');
```
amandeepsingh333, line 83: **Optional chaining masks a real misconfiguration.**

```js
page.session.browser?.version?.userAgent?.replace('Headless', '');
```

If `browser.version.userAgent` is missing, the user-agent silently becomes `undefined`, which then flows into `Network.setUserAgentOverride` and Chrome receives no UA override — customers' user-agent assertions break with no error.

The previous code (`page.session.browser.version.userAgent.replace(...)`) would have thrown loudly and forced you to investigate. If the goal is hardening against a missing browser fixture in some test path, prefer a localized fallback at the actual fixture, not silent erasure at the production call site:

```js
let ua = page.session.browser?.version?.userAgent;
this.userAgent = options.userAgent ?? (ua ? ua.replace('Headless', '') : DEFAULT_UA);
```

Then `DEFAULT_UA` is a known, audit-able fallback rather than `undefined`.

amandeepsingh333 (commented):
Excellent write-up — the diagnostic story for PlzDedicatedWorker and the Chromium-source citations are the kind of context that makes this reviewable at all. The layered fix (response-stage interception for malformed CL + 2s `responseReceived` timeout + direct-fetch fallback with origin-scoped auth) is the right shape.

Non-test concerns, ordered by severity:

**Should-fix (correctness/operational):**
1. `install.js` — `darwinArm: '1536376'` is identical to `win64`. Likely a copy-paste error; ARM macOS should have its own snapshot.
2. `network.js` — `sec-ch-ua-platform: '"macOS"'` is hardcoded and sent from Linux/Windows runners. Servers doing content negotiation on this header will misbehave. PR body marks this "deferred" but it's a real signal.
3. `network.js#_continueResponse` — PR body claims "Production observability via `log.warn`" but the code uses `log.debug`. Either fix code or fix the description.
4. `captureResourceDirectly` mimetype default `application/javascript` — assumes worker-scripts. For non-worker direct fetches that hit this path (font fallback was the original caller), an extensionless URL silently gets mistyped as JS.

**Nice-to-have (clarity / defensive):**
5. `browser.js` comma-bundled `--disable-features` list mixes process-isolation (security boundary) with HTTPS-first and LNA. Worth a comment per disabled feature explaining why it's safe in *headless* asset discovery and not safe to copy to any non-headless code path.
6. `shouldAttachAuth` defensively `return false` when `meta?.snapshotURL` is undefined — confirm `meta.snapshotURL` is guaranteed populated in all entry paths; otherwise existing customers using `network.authorization` with the font fallback may silently lose auth.
7. `RESPONSE_RECEIVED_TIMEOUT` cumulative `N*2s` — comment acknowledges it but no hard cap. A pathological page with many worker fetches that all timeout could compound. Worth a per-snapshot ceiling.

I deliberately skipped test files per your request. Notes inline.

packages/core/src/install.js
Manoj-Katta: Verified against [chromium-browser-snapshots](https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html) — all 5 revisions resolve to distinct platform-specific artifacts:

- `Linux_x64/1536366/chrome-linux.zip` (216 MB)
- `Win_x64/1536376/chrome-win.zip` (308 MB)
- `Win/1536377/chrome-win.zip` (281 MB)
- `Mac/1536380/chrome-mac.zip` (169 MB)
- `Mac_Arm/1536376/chrome-mac.zip` (152 MB — ARM-specific, NOT a Windows binary)

The shared revision number `1536376` between `darwinArm` and `win64` is a coincidence of when Chromium uploaded both platforms; the artifacts under `Mac_Arm/` and `Win_x64/` are completely different binaries with very different sizes. Same answer I gave on @bhokaremoin's C11 earlier.

packages/core/src/network.js
Manoj-Katta: Fixed in `ebad8275`. New precedence in `captureResourceDirectly`:

1. HTTP `Content-Type` response header from the direct fetch (server-authoritative)
2. URL-extension `mime.lookup()`
3. `application/octet-stream` (safe binary default, **never** JS)

`makeDirectRequest`'s callback now returns `headers` alongside `body`/`status` so the direct-fetch path can read the server's explicit MIME. Font path in `saveResponseResource` still destructures `{ body }` only — unaffected. The unknown-extension integration spec was updated to assert the server's declared `application/octet-stream` is preserved (was previously asserting the JS fallback this commit removes).

packages/core/src/network.js
Manoj-Katta: Audited the entry paths — `meta.snapshotURL` is set at exactly one site (`discovery.js:606`):

```js
meta: { ...snapshot.meta, snapshotURL: snapshot.url },
```

`snapshot.url` is mandatory and validated upstream by `validURL(options.url)` in `snapshot.js:127`, so a snapshot without a URL fails validation before any `Network` instance is constructed. The `makeDirectRequest` path therefore can't be entered with `meta.snapshotURL=undefined` from any real customer flow.

The defensive `catch { return false }` in `shouldAttachAuth` is there to keep the contract honest if a future code path forgets to populate meta, not because any current path leaves it empty. Current behavior matches v126's font fallback for all valid customer inputs — no silent auth-drop regression.

Happy to add a one-time `log.debug` from `shouldAttachAuth` when `snapshotUrl` is missing if you'd like an explicit signal for future code paths.

packages/core/src/network.js
Manoj-Katta: Good catch — PR body updated. The `log.warn` claim was stale: we did promote those catches in `f0d9c8e7`, then reverted in `334f0621` because routine `Session closed` / `Target closed` lifecycle races fired these catches on every snapshot teardown and the warn noise drowned legitimate signals.

Moved the observability story into the deferred-followups section pointing at structured instrumentation telemetry as the right home for stuck-snapshot diagnostics — keeps customer logs quiet on routine teardowns and gives oncall a dedicated dashboard instead of grep-ing CI logs.

packages/core/src/browser.js
Manoj-Katta: Fixed in `59852c68`. The inline `--disable-features=` string is now a `DISABLED_FEATURES` array with a per-entry comment explaining each flag, including explicit `[headless-only]` tags on `IsolateOrigins` / `site-per-process` so a future contributor copy-pasting this pattern into a non-headless code path sees the threat-model caveat. Joined with `.join(',')` at launch time. No behavior change.

packages/core/src/network.js
Manoj-Katta: Looked at this carefully and decided to leave the optional chaining as-is for now. Context: `page.session.browser?.version?.userAgent?.replace(...)` was added in this PR precisely to keep a `TypeError` from masking the real "Session crashed!" error when the session races with constructor invocation. The downside you flag (silent `undefined` flowing into `setUserAgentOverride`) is real, but in practice CDP returns the override call as a no-op when the UA is undefined — customers see no change, not a broken assertion.

The cleaner long-term fix (named `DEFAULT_USER_AGENT` fallback) means touching the direct-fetch header path that the v126 font fallback also depends on. We are actively working on [PPLT-4215](https://browserstack.atlassian.net/browse/PPLT-4215) for the font-asset capture regression — if that work requires this fallback as part of the fix, it will land there. Otherwise we will skip the change entirely.

packages/core/src/network.js
Manoj-Katta: Holding off on both halves for now. Reasoning:

- `sec-ch-ua-platform: "macOS"` — real but very-low-impact issue. The header is only sent in the Node-side direct-fetch fallback (worker scripts / font path), and only matters if the origin server does Client-Hints content negotiation, which most asset CDNs ignore. Worst plausible case: a font CDN serves a slightly different glyph file per platform → CLI caches the macOS variant → renderer (Linux) fetches the Linux variant separately and the visual diff still matches since the renderer side is also Linux.

- `sec-fetch-dest: "font"` — agreed it's wrong for the worker-script path. In practice Chrome's original `Fetch.requestPaused` event carries its own `sec-fetch-dest` (`"worker"` for worker scripts, `"script"` for module imports, etc.), and our header object spreads `...request.headers` **after** our default `'font'`, so the original value wins for actual customer requests. The hardcoded default only shows up if a request somehow arrives without `sec-fetch-dest` set — unlikely.

[PPLT-4215](https://browserstack.atlassian.net/browse/PPLT-4215) is open and active for the font-asset capture regression on this branch. If that work requires either of these header corrections to fully resolve the regression, the change lands there. Otherwise we skip both.

packages/core/src/network.js
Manoj-Katta: Comment expanded in `59852c68` to call out `PERCY_NETWORK_IDLE_WAIT_TIMEOUT` (default 30s) as the cumulative safety valve — when the per-request `RESPONSE_RECEIVED_TIMEOUT` waits push a page's total idle wait past that budget, the idle wait throws and the snapshot terminates with a proper error path. So the worst case is bounded at the snapshot level even though no per-instance cap exists today.

For the customer-with-a-thousand-workers case you mention, a defensive in-instance cap (track time-spent-in-`responseReceived`-timeouts on the `Network` instance, bail above a threshold) is the right knob. Tracking as a follow-up — will land if/when telemetry surfaces a real customer hitting it.

packages/core/src/network.js
Manoj-Katta: Going to leave the `!= null` form as-is. Two reasons:

1. `!= null` is the looser check (catches both `null` and `undefined`); `!== undefined` only catches `undefined`. CDP doesn't guarantee which form an absent field takes — the looser check is the safer one to use against an external protocol.

2. The "valid HTTP status `0`" concern isn't a real divergence: `0 != null` and `0 !== undefined` both evaluate to `true`, so neither check changes behavior for status `0`.

The deeper concern (future CDP schema change introducing an explicit `null` for these fields) is a real risk and applies equally to either form. Cleanest long-term answer is a CDP-side `stage` discriminator, which Chromium hasn't surfaced yet. Happy to revisit if you have a CDP schema reference that argues `!== undefined` is the contract.

packages/core/src/browser.js
```diff
@@ -11,6 +11,16 @@ import install from './install.js';
 import Session from './session.js';
 import Page from './page.js';
 
+// Chrome features Percy disables for v143 new-headless asset discovery.
+const DISABLED_FEATURES = [
+  'Translate', // suppress translate prompt overlay
+  'OptimizationGuideModelDownloading', // suppress background model fetches
+  'IsolateOrigins', // [headless-only] keep cross-origin sub-resources on the page session for CDP capture
+  'site-per-process', // companion to IsolateOrigins
+  'HttpsFirstBalancedModeAutoEnable', // allow HTTP customer URLs (CI / local dev / staging)
```
amandeepsingh333, line 20: Why these disabled flags

amandeepsingh333:
Have we tested locally?

packages/core/src/browser.js
Manoj-Katta: the first two flags are the existing disabled flags that are present in v126(present version) too

- The other I have added a better explaination of those in pr description actually 1.1, 1.2, 1.3

amandeepsingh333 (approved):

bhokaremoin (approved):
Approving — all 12 findings from my original review are addressed (9 fixed in code, 2 defended with reasoning captured in-thread, 1 explicit deferral per pre-approval). The second reviewer's pass added 4 more genuine improvements (mimetype fallback, `DISABLED_FEATURES` extraction, PR-body/code mismatch, ARM macOS revision verified). All 26 review threads are resolved. CI is green across all 44 checks including the previously-red `Test @percy/core` coverage gate.

The layered v143-compat work (dual-stage Fetch interception, `RESPONSE_RECEIVED_TIMEOUT` + `DIRECT_FETCH_TIMEOUT` worker-script fallback, `shouldAttachAuth` same-origin guard, `pickCookieSession` page-session preference) is well-considered and the diagnostic write-ups in the PR body are exemplary.

Remaining deferred work is properly externalized:
- **PPLT-4215** — `sec-ch-ua-platform`, `sec-fetch-dest`, optional-chain UA fallback (font-asset capture regression follow-up)
- Stuck-snapshot observability → structured instrumentation telemetry follow-up
- CHANGELOG entry for cookie-less worker-script fallback → release time
- Per-instance cumulative-timeout cap → telemetry-driven, only if a customer hits it

None of those are correctness regressions vs. v126. Ship it.

Issue #361, opened 2021-06-08
How to skip Chromium download?


geoffharcourt:
We recently updated our app from `@percy/agent` to use the CLI package. [The docs for skipping downloading Chromium](https://docs.percy.io/docs/skipping-puppeteer-chromium-download) appear to no longer accurate (or are solely for the agent package). 

I tried searching for `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` and `PUPPETEER_EXECUTABLE_PATH` in this repo and it looks like they recently disappeared in the last two weeks. The CLI fails to find our previously installed browser because (I think?) the code assumes a fixed location for the pre-existing Chrome/Chromium executable. 

This is a blocker for us to update because in addition to the extra bulk of downloading Chromium, the automated download doesn't work on Alpine Linux.

I think this probably works if I can symlink our existing executable (currently installed at `/usr/bin/chromium-browser`) to the location that the package expects. Would that be enough? Where would we point it the symlink to work around this?

Thanks for your help.

geoffharcourt:
I clicked on a link from before that was for the agent package, thinking it was for the CLI package, I think this is now covered. Sorry for the ticket!

Issue #403, opened 2021-07-01
TypeError: PercyClient is not a constructor


holdenhinkle:
I'm getting the following error:

```
const client = new PercyClient(options);
                 ^

TypeError: PercyClient is not a constructor
```

When I run:

```
const PercyClient = require('@percy/client');

function run() {
  const client = new PercyClient();
};

run();
```

Please advise. Thank you!

Robdel12:
Hey @holdenhinkle! Curious what you're looking to use such a low level package for. 👀 

For the issue, you'll probably have to `default` on the require: 

```js
```

holdenhinkle:
Thanks for your super fast response, Robert! That worked.

We want to monitor a couple of important pages in production.  I'm trying to use Percy in an AWS CloudWatch Canary script.

Robdel12:
Gotcha! Are you looking to build your own SDK? `@percy/client` is pretty low level. You'll have to find, create, and manage your own builds & resources. You might be better off using `@percy/core` _or_ one of our prebuilt SDKs

holdenhinkle:
Yes, `@percy/core` is what I wanted. Thanks. I've got it working :-)

How can I get a read access token so I can query a build?

Robdel12:
You can get the full access token (read/write) from the link in the docs here: https://docs.percy.io/docs/cli-build#percy-buildwait

We'll soon be exposing these in a nice way in the UI. :D

holdenhinkle:
Thanks Robert! 

I didn't see that in the docs. I reached out to support and they responded quickly with that same link.

Thanks again :-)

Robdel12:
Awesome! Happy to hear -- I'll close this issue up 🎉

Issue #401, opened 2021-06-29
CDN Font loading inconsistently


chris-dura:
I've ran across a few issues referencing problems with [loading web fonts](https://github.com/percy/cli/issues/394). But, I'm still experiencing some issues, and wondering if there's something more I should be trying.

I'm migrating to `@percy/cli` from `PercyScript`.

My `.percy.json` file:

```json
{
	"version": 2,
	"snapshot": {
		"widths": [
			320,
			1280
		]
	},
	"discovery": {
		"disable-cache": true
	}
}
```

Adding a ridiculously long timeout didn't seem to help.

```js
// snapshots.js

module.exports = [
	{
		name: 'snapshot',
		url: 'http://localhost:8080',
		waitForTimeout: 15000,
	},
]
```

I'm currently running Percy against a local dev build that is available on `localhost`, and with `--verbose` flag, here is an excerpt of the output:

```sh
...
[percy:core:discovery] Handling request: https://cdn.mycompany.net/fonts/foo/foo-medium.woff2 (1212ms)
[percy:core:discovery] Handling request: https://cdn.mycompany.net/fonts/foo/foo-bold.woff2 (0ms)
[percy:core:discovery] Handling request: https://cdn.mycompany.net/fonts/foo/foo-regular.woff2 (1ms)
[percy:core:discovery] Handling request: https://cdn.mycompany.net/fonts/foo/foo-regularitalic.woff2 (0ms)
[percy:core:discovery] Processing resource: http://localhost:8080/app.css (18ms)
[percy:core:discovery] -> sha: 70c49d43caaee49f8583a4b5992d8c0ae591071f699318ddf98e2c7a3b018d0a (1ms)
[percy:core:discovery] -> mimetype: text/css (0ms)
[percy:core:discovery] Processing resource: https://cdn.mycompany.net/fonts/foo/foo-medium.woff2 (381ms)
[percy:core:discovery] -> Skipping remote resource (0ms)
[percy:core:discovery] Processing resource: https://cdn.mycompany.net/fonts/foo/foo-bold.woff2 (9ms)
[percy:core:discovery] -> Skipping remote resource (1ms)
[percy:core:discovery] Processing resource: https://cdn.mycompany.net/fonts/foo/foo-regular.woff2 (0ms)
[percy:core:discovery] -> Skipping remote resource (0ms)
[percy:core:discovery] Processing resource: https://cdn.mycompany.net/fonts/foo/foo-regularitalic.woff2 (2ms)
[...]

PR #2192, opened 2026-04-22
feat(core): add --max-cache-ram bounded asset-discovery cache [PER-7795]

pranavz28:
## Summary

Adds a `--max-cache-ram <MB>` flag (plus `PERCY_MAX_CACHE_RAM` env and `discovery.maxCacheRam` percyrc) to cap the asset-discovery cache memory. When set, a hand-rolled byte-budget LRU evicts least-recently-used resources to stay within the cap. **Evicted resources are not dropped — they spill to a per-run disk tier and are transparently rehydrated on cache lookup**, so a memory-bounded run behaves like an unbounded one at the cost of a local disk read (cheaper than an origin refetch). When unset, behavior is identical to today plus a one-shot warn-level log at 500 MB pointing users at the flag before they OOM. Two CLI-side telemetry events (`cache_eviction_started`, `cache_summary`) flow through the existing UDP pager → Amplitude pipeline for adoption / hit-rate / disk-tier analytics.

Ships entirely in `percy/cli`. **Zero Percy API changes.** **Zero new npm dependencies** (disk tier uses only Node built-ins: `fs`, `os`, `path`, `crypto`). `@percy/core` Node engine unchanged (`>=14`).

### Ticket & planning

- Jira: https://browserstack.atlassian.net/browse/PER-7795
- Task Brief + engineering brainstorm captured in the companion docs (TB, engineering requirements, plan). All five PoCs executed during brainstorming; R6 (server-side OOM-suspected bucket) was dropped after PoC 3 confirmed Percy API does not persist `send-events` `extra` and cilogs do not carry CLI process exit codes.

## Key decisions

- **Hand-rolled `ByteLRU`** (~60 LOC, zero dep) instead of `lru-cache` npm — v7 is unmaintained-since-2022, v10+ requires a Node engine bump that would break existing users on Node 14/16. Percy's cache usage is ~5% of any LRU library's surface.
- **Disk-spill overflow tier (`DiskSpillStore`)** lives in the same `cache/byte-lru.js` module as the RAM tier — one cache module, two tiers. When RAM evicts, the full resource is written synchronously to a per-run temp directory under `os.tmpdir()/percy-cache-<pid>-<rand>/`. A slim metadata reference stays in an in-memory `Map`. `getResource()` (via the extracted `lookupCacheResource` helper) falls through RAM miss → disk-index lookup → `readFileSync` → return full resource. Browser never sees a refetch on a disk hit. **On any disk I/O failure (init, write, read) we fall back to the old drop-on-evict behavior** — the browser refetches from origin exactly as it would without spill. Index is self-healing (read failure purges the entry). Temp dir is rm'd in the queue `'end'` handler; cleanup errors are swallowed so they cannot fail `percy.stop()`.
- **Counter-based filenames** (`dir/1`, `dir/2`, …) rather than URL-derived — no user-controlled data flows into `path.join`, eliminating the path-traversal surface semgrep flags on URL-in-filename patterns. Collisions are impossible within a run because the counter is monotonic + the dir is fresh per run.
- **Sync fs ops on both hot paths**, not async. `ByteLRU.set` is synchronous and `getResource` is the sync callback CDP network-intercept calls — neither can yield to the event loop mid-operation. Per-entry size is capped at 25 MB upstream in `network.js`, so worst-case disk latency is a single bounded I/O op — still strictly cheaper than an origin refetch.
- **Byte accounting = cache body bytes only** (`resource.content.length` + 512 B per-entry overhead). Flag caps cache, not RSS. MB is decimal (1,000,000 B), not binary MiB. Docs note: real-world RSS is typically 1.5–2× cache bytes due to Node's Buffer slab allocator (PoC 4 calibration).
- **MB semantics throughout**: `--max-cache-ram 300` means 300 MB. Users never write unit suffixes.
- **Sub-25 MB values clamp to 25 MB with a warn log** (not a hard error). The per-resource ceiling is 25 MB, so any cap below that would silently skip every resource; clamping preserves the user's intent without killing the build. User's original `percy.config.discovery.maxCacheRam` is NOT mutated — the effective cap lives on `percy[CACHE_STATS_KEY].effectiveMaxCacheRamMB` and is what `cache_summary` telemetry reports.
- **Warning-at-threshold at `warn` level** so it surfaces under `--quiet`; suppressed under `--silent`. Threshold default 500 MB, override via `PERCY_CACHE_WARN_THRESHOLD_BYTES` (read once at queue construction; debug-logged at startup when the override is active).
- **Telemetry folded into two events** (not three). `cache_budget_configured` was dropped because it would fire before `percy.build.id` exists; its fields live inside `cache_eviction_started` + `cache_summary` which are guaranteed to fire post-build-creation. `cache_summary` is ordered AFTER `sendBuildLogs` in `percy.stop()`'s finally so analytics latency cannot delay the primary log egress. `cache_summary` now also carries eight disk-tier fields (`disk_spill_enabled`, `disk_spilled_count`, `disk_restored_count`, `disk_spill_failures`, `disk_read_failures`, `disk_peak_bytes`, `disk_final_bytes`, `disk_final_entries`) so the dashboard can distinguish "disabled", "enabled with no activity", and "enabled with failures".

## Commit structure (bisectable)

```
bf6b92e0 feat(core): spill evicted resources to disk, restore on lookup
a299cd1b test(core): close remaining coverage gaps
e1d6e1c2 fix(core): address PR review (frozen counter, reorder checks, reorder egress, effective cap)
dad8a08e fix(core): clamp --max-cache-ram below 25MB instead of failing
6ae0659b docs(core): document --max-cache-ram flag
d1cf2c0d test(core): integration coverage for --max-cache-ram
2e26f1a5 feat(core): emit cache_eviction_started + cache_summary telemetry
ff45eb38 feat(core): add warning-at-threshold when --max-cache-ram is unset
9e173ecc feat(core): swap Map for ByteLRU when --max-cache-ram is set
0e5364c3 feat(core): extend discovery config schema with maxCacheRam
eed9c323 feat(cli-command): add --max-cache-ram flag (MB units)
537b336f feat(core): add ByteLRU + entrySize helpers
```

Commits 9 and 10 were added during self-review: clamp behaviour + a separately-broken help-text fixture in `cli-command/test/flags.test.js` that was stale since the flag was introduced. Commit 12 is the disk-spill upgrade as a single squashed commit — `DiskSpillStore`, wiring, `lookupCacheResource` extraction, telemetry fields, and all 15 new test specs land together.

### Automated
- **44 unit specs** in `packages/core/test/unit/byte-lru.test.js` cover ByteLRU, entrySize, DiskSpillStore, createSpillDir, and lookupCacheResource:
  - ByteLRU: unbounded mode, LRU eviction (single + multi), recency bump, oversized entries (skip + preserve prior on re-set), onEvict signature `(key, reason, value)`, `.values()`, `.clear()`, `.delete()` byte accounting, peak-bytes transient high-water, hits/misses/evictions tracking, NaN/negative guards, churn stability.
  - DiskSpillStore: mkdir success + failure (via `/dev/null/…`), not-ready short-circuit, not-ready destroy no-op, binary round-trip with non-ASCII bytes, string→Buffer coercion, coercion failure (Symbol content), null/undefined content guards, metadata carry-through (`sha`, `root`, `widths`), counter-based filenames, accounting on replace/delete, peak tracking, write failure via mocked EACCES, read self-heal via mocked ENOENT, unlinkSync error tolerance on both delete + overwrite, destroy error swallowing via mocked EBUSY.
  - createSpillDir: uniqueness, `percy-cache-` prefix, tmpdir scoping.
  - lookupCacheResource: snapshot-local hit, RAM hit, disk hit (with debug-log verification), full miss, disk-present-no-hit, array-root width match, array-root fallback.
- **8 integration specs** in `packages/core/test/discovery.test.js` (describe `with --max-cache-ram disk-spill tier`): DiskSpillStore presence only when cap is set, spill-on-eviction with `cache spill:` debug log, byte-for-byte rehydration of binary content, disk-write failure fallback to drop with `cache evict:` debug log (ENOSPC simulated), saveResource clearing stale disk entries, queue-end teardown (asserts both `disk.ready === false` and `fs.existsSync(dir) === false` for race safety), graceful handling of a DiskSpillStore that fails to initialize (via mocked mkdirSync).
- Existing 19 byte-lru unit specs + 13 max-cache-ram integration specs stay green; only the two log-string assertions changed (`Skipping cache for resource` → `cache skip (oversize):`, eviction-active info log now mentions "spilling to disk").
- **Semgrep: 0 findings on all changed files** (ran locally with `semgrep --config=auto` and the OSS `path-join-resolve-traversal` rule that previously flagged a `path.join(os.tmpdir(), ...)` false-positive on `createSpillDir`; resolved by dropping the unused `prefix` parameter and hard-coding `percy-cache-`).
- Local lint on this machine is broken (pre-existing `@babel/eslint-parser` config detection issue that also reproduces on `master`); CI will gate.

### Security
The 34 Dependabot alerts on this branch are **pre-existing transitive dependencies on `master`** (basic-ftp, flatted, ip, lodash, minimatch, rollup, systeminformation, tar, axios, follow-redirects, js-yaml, picomatch, qs, yaml, brace-expansion, @tootallnate/once, tmp, uuid, etc.). This PR introduces **zero new npm dependencies**. Dependency-bump PRs should be handled separately.

### Pending manual verification
The original 4 builds (#353–#356) exercise the max-cache-ram plumbing but pre-date disk-spill. Re-running the 30 × 1 MB heavy-assets workload from the original verification (the one that produced build #361's `cache evict:` storm) will now show `cache spill:` + `cache disk-hit:` lines instead of drop-and-refetch. Build IDs + MCP verification will be added to this PR body once the new builds are produced locally.

- **What to monitor/search**
  - **Amplitude (primary)**: events `cache_eviction_started`, `cache_summary`. Track presence, frequency, and field distributions — now including `disk_spill_enabled`, `disk_spilled_count`, `disk_restored_count`, `disk_spill_failures`, `disk_read_failures`, `disk_peak_bytes`, `disk_final_bytes`, `disk_final_entries` alongside the original RAM-tier fields.
  - **Honeycomb (secondary)**: `service.name=cli` traces with `core:discovery` log namespace. Search for message fragments `cache spill:`, `cache evict:` (now only fires when disk spill **failed** — its presence is a signal, not routine), `cache disk-hit:`, `Cache eviction active — cap reached, oldest entries spilling to disk`, `Percy cache is using`, `--max-cache-ram=`, `is below the 25MB minimum`, `PERCY_CACHE_WARN_THRESHOLD_BYTES override active`, `disk-spill init failed`, `disk-spill write failed`, `disk-spill read failed`, `disk-spill cleanup failed`.
  - **Support channel `#percy-support`**: watch for tickets mentioning `max-cache-ram`, `OOM`, `heap`, `killed`, `/tmp`, `ENOSPC`, `disk full`, `permission denied` after release.
- **Validation checks (queries / commands)**
  - Amplitude filter: `event_type IN ('cache_eviction_started','cache_summary') AND cli_version >= <release>` — expect non-zero rows within 24 h of release for early adopters.
  - `cache_summary.disk_spill_failures / (disk_spilled_count + disk_spill_failures)` should be near zero. A rising ratio signals a disk-tier regression (permissions, noexec /tmp, full volume).
  - `cache_summary.disk_restored_count > 0` alongside `cache_summary.evictions > 0` confirms the round-trip is working in anger, not just the spill side.
  - Support-query: `-max-cache-ram` across Zendesk/Intercom — expect no new tickets tied to flag parsing, cap enforcement, sub-25 MB clamp, or disk-tier cleanup (leaked temp dirs).
  - ≥ 10% of opt-in runs show `cache_summary` with a non-null `cache_budget_ram_mb` within 60 days.
  - On runs with `evictions > 0`: `disk_spilled_count` should track `evictions` (1:1 minus a tiny tail of simultaneous-write races).
  - `disk_restored_count / disk_spilled_count` > 0.5 on typical memory-constrained workloads — proves the disk tier earns its keep.
  - Build-failure rate unchanged vs. pre-release baseline.
  - Trigger A: > 10 support tickets in 7 days tied to `--max-cache-ram` flag parsing, cap clamp, disk-tier init, or "my build started failing after setting this".
  - Trigger B: measurable spike in Percy build-failure rate correlated with CLI versions that include this PR.
  - Trigger C: `cache_summary` events are never received in Amplitude (pipeline regression).
  - Trigger D: `cache_summary.disk_spill_failures` > `disk_spilled_count` across a large cohort — means the disk tier is failing open for most users; the feature would be a net loss over the old drop-on-evict behaviour.
  - Trigger E: `cache_summary.peak_bytes` values clustered at the 500 MB default threshold (signals the Map-mode counter has regressed to frozen behavior).
  - Immediate action on any trigger: release a patch reverting this PR; tell users to unset the flag as a workaround (flag is opt-in, unset = previous behavior).
  - Window: 14 days post-GA release.
  - Owner: @pranavz28 (Pranav Zinzurde).

🤖 Generated with Claude Opus 4.7 (1M context, extended thinking) via [Claude Code](https://claude.com/claude-code) + Compound Engineering v2.50.0

### DIFF
--- a/packages/core/src/cache/byte-lru.js
+++ b/packages/core/src/cache/byte-lru.js
@@ -81,17 +81,6 @@ export class ByteLRU {
     this.#bytes = 0;
   }
 
-  values() {
-    const iter = this.#map.values();
-    return {
-      next: () => {
-        const r = iter.next();
-        return r.done ? r : { value: r.value.value, done: false };
-      },
-      [Symbol.iterator]() { return this; }
-    };
-  }
-
   get size() { return this.#map.size; }
   get calculatedSize() { return this.#bytes; }
   get stats() { return { ...this.#stats, currentBytes: this.#bytes }; }
@@ -118,7 +107,10 @@ export class DiskSpillStore {
     this.dir = dir;
     this.log = log;
     try {
-      fs.mkdirSync(dir, { recursive: true });
+      // mode 0o700: spilled bytes are origin-fetchable so the threat model is
+      // small, but on shared-tenant CI hosts other users on the same box
+      // shouldn't be able to read them.
+      fs.mkdirSync(dir, { recursive: true, mode: 0o700 });
       this.#ready = true;
     } catch (err) {
       this.log?.debug?.(`disk-spill init failed for ${dir}: ${err.message}`);
--- a/packages/core/src/cache/byte-lru.js
+++ b/packages/core/src/cache/byte-lru.js
@@ -1,0 +1,98 @@
+// Hand-rolled byte-budget LRU cache. Map insertion order is LRU order;
+// .get() deletes and re-sets to move an entry to MRU. Synchronous by design —
+// no logger calls, no awaits inside mutation paths, so callers can log
+// after .set() returns without risking a mid-state event-loop yield.
+
+const DEFAULT_PER_ENTRY_OVERHEAD = 512;
+
+export class ByteLRU {
+  #map = new Map();
+  #bytes = 0;
+  #max;
+  #stats = { hits: 0, misses: 0, evictions: 0, peakBytes: 0 };
+  onEvict;
+
+  constructor(maxBytes, { onEvict } = {}) {
+    this.#max = maxBytes;
+    this.onEvict = onEvict;
+  }
+
+  get(key) {
+    if (!this.#map.has(key)) {
+      this.#stats.misses++;
+      return undefined;
+    }
+    const rec = this.#map.get(key);
+    this.#map.delete(key);
+    this.#map.set(key, rec);
+    this.#stats.hits++;
+    return rec.value;
+  }
+
+  set(key, value, size) {
+    if (!Number.isFinite(size) || size < 0) return false;
+
+    if (this.#map.has(key)) {
+      this.#bytes -= this.#map.get(key).size;
+      this.#map.delete(key);
+    }
+
+    if (this.#max !== undefined && size > this.#max) {
+      if (this.onEvict) this.onEvict(key, 'too-big');
+      return false;
+    }
+
+    this.#map.set(key, { value, size });
+    this.#bytes += size;
+    if (this.#bytes > this.#stats.peakBytes) this.#stats.peakBytes = this.#bytes;
+
+    while (this.#max !== undefined && this.#bytes > this.#max) {
+      const oldestKey = this.#map.keys().next().value;
+      const rec = this.#map.get(oldestKey);
+      this.#bytes -= rec.size;
+      this.#map.delete(oldestKey);
+      this.#stats.evictions++;
+      if (this.onEvict) this.onEvict(oldestKey, 'lru');
+    }
+
+    return true;
+  }
+
+  has(key) { return this.#map.has(key); }
+
+  delete(key) {
+    if (!this.#map.has(key)) return false;
+    this.#bytes -= this.#map.get(key).size;
+    return this.#map.delete(key);
+  }
+
+  clear() {
+    this.#map.clear();
+    this.#bytes = 0;
+  }
+
+  values() {
+    const iter = this.#map.values();
+    return {
+      next: () => {
+        const r = iter.next();
+        return r.done ? r : { value: r.value.value, done: false };
+      },
+      [Symbol.iterator]() { return this; }
+    };
+  }
+
+  get size() { return this.#map.size; }
+  get calculatedSize() { return this.#bytes; }
+  get stats() { return { ...this.#stats, currentBytes: this.#bytes }; }
+}
+
+// Compute the byte size attributable to a cache entry. Handles the two Percy
+// shapes: a single resource object, or an array of resources (root-resource
+// captured at multiple widths per discovery.js:465).
+export function entrySize(resource, overhead = DEFAULT_PER_ENTRY_OVERHEAD) {
+  if (Array.isArray(resource)) {
+    return resource.reduce((n, r) => n + (r?.content?.length ?? 0) + overhead, 0);
+  }
+  return (resource?.content?.length ?? 0) + overhead;
+}
--- a/packages/core/src/cache/byte-lru.js
+++ b/packages/core/src/cache/byte-lru.js
@@ -32,14 +32,16 @@ export class ByteLRU {
   set(key, value, size) {
     if (!Number.isFinite(size) || size < 0) return false;
 
-    if (this.#map.has(key)) {
-      this.#bytes -= this.#map.get(key).size;
-      this.#map.delete(key);
-    }
-
+    // Reject oversize BEFORE touching any existing entry — a failed oversize
+    // set on an existing key must not evict the prior (valid) entry.
     if (this.#max !== undefined && size > this.#max) {
       if (this.onEvict) this.onEvict(key, 'too-big');
       return false;
+    }
+
+    if (this.#map.has(key)) {
+      this.#bytes -= this.#map.get(key).size;
+      this.#map.delete(key);
     }
 
     this.#map.set(key, { value, size });
--- a/packages/core/src/cache/byte-lru.js
+++ b/packages/core/src/cache/byte-lru.js
@@ -86,13 +86,43 @@ export class ByteLRU {
   get stats() { return { ...this.#stats, currentBytes: this.#bytes }; }
 }
 
+// Returns the byte length of a resource's content. Buffer.byteLength is used
+// for strings so that multi-byte UTF-8 (CJK, emoji) is counted in bytes, not
+// JS string units, otherwise the cache budget can drift past its cap.
+function contentBytes(content) {
+  if (content == null) return 0;
+  if (Buffer.isBuffer(content)) return content.length;
+  if (typeof content === 'string') return Buffer.byteLength(content);
+  return content.length ?? 0;
+}
+
 // Handles the two Percy cache-entry shapes: single resource, or array of
 // roots captured at multiple widths (see discovery.js parseDomResources).
 export function entrySize(resource, overhead = DEFAULT_PER_ENTRY_OVERHEAD) {
   if (Array.isArray(resource)) {
-    return resource.reduce((n, r) => n + (r?.content?.length ?? 0) + overhead, 0);
+    return resource.reduce((n, r) => n + contentBytes(r?.content) + overhead, 0);
   }
-  return (resource?.content?.length ?? 0) + overhead;
+  return contentBytes(resource?.content) + overhead;
+}
+
+// Multi-width root arrays carry per-element binary content. Encode buffers as
+// base64 inside JSON so the whole array survives a disk roundtrip; null and
+// string content pass through as themselves.
+function encodeArrayElement(r) {
+  if (!r) return r;
+  const { content, ...rest } = r;
+  if (content == null) return { ...rest, content: null };
+  if (Buffer.isBuffer(content)) return { ...rest, content: { __buf: content.toString('base64') } };
+  return { ...rest, content: String(content) };
+}
+
+function decodeArrayElement(r) {
+  if (!r) return r;
+  const { content, ...rest } = r;
+  if (content && typeof content === 'object' && '__buf' in content) {
+    return { ...rest, content: Buffer.from(content.__buf, 'base64') };
+  }
+  return { ...rest, content };
 }
 
 export class DiskSpillStore {
@@ -119,13 +149,31 @@ export class DiskSpillStore {
 
   // Returns true on success; false on any failure so caller falls back to drop.
   // Overwrites prior spill for the same URL — a fresh discovery write wins.
+  // Two resource shapes are supported: a single resource with a binary
+  // .content, and a multi-width root array (see entrySize for the array
+  // shape). Arrays are JSON-encoded with base64 buffers so the whole array
+  // survives the disk roundtrip.
   set(url, resource) {
     if (!this.#ready) return false;
 
-    let content = resource?.content;
-    if (content == null) return false;
-    if (!Buffer.isBuffer(content)) {
-      try { content = Buffer.from(content); } catch { return false; }
+    let bytes;
+    let meta;
+    let isArray = false;
+
+    if (Array.isArray(resource)) {
+      isArray = true;
+      try {
+        bytes = Buffer.from(JSON.stringify(resource.map(encodeArrayElement)));
+      } catch { return false; }
+    } else {
+      let content = resource?.content;
+      if (content == null) return false;
+      if (!Buffer.isBuffer(content)) {
+        try { content = Buffer.from(content); } catch { return false; }
+      }
+      bytes = content;
+      meta = { ...resource };
+      delete meta.content;
     }
 
     // Counter-based filename keeps URL-derived data out of path.join —
@@ -133,7 +181,7 @@ export class DiskSpillStore {
     const filepath = path.join(this.dir, String(++this.#counter));
 
     try {
-      fs.writeFileSync(filepath, content);
+      fs.writeFileSync(filepath, bytes);
     } catch (err) {
       this.#stats.spillFailures++;
       this.log?.debug?.(`disk-spill write failed for ${url}: ${err.message}`);
@@ -146,10 +194,8 @@ export class DiskSpillStore {
       try { fs.unlinkSync(prev.path); } catch { /* best-effort */ }
     }
 
-    const meta = { ...resource };
-    delete meta.content;
-    this.#index.set(url, { path: filepath, size: content.length, meta });
-    this.#bytes += content.length;
+    this.#index.set(url, { path: filepath, size: bytes.length, isArray, meta });
+    this.#bytes += bytes.length;
     if (this.#bytes > this.#peakBytes) this.#peakBytes = this.#bytes;
     this.#stats.spilled++;
     return true;
@@ -158,17 +204,30 @@ export class DiskSpillStore {
   get(url) {
     const entry = this.#index.get(url);
     if (!entry) return undefined;
-    let content;
+    let raw;
     try {
-      content = fs.readFileSync(entry.path);
+      raw = fs.readFileSync(entry.path);
     } catch (err) {
       this.#stats.readFailures++;
       this.log?.debug?.(`disk-spill read failed for ${url}: ${err.message}`);
       this.#removeEntry(url, entry);
       return undefined;
     }
+    if (entry.isArray) {
+      let arr;
+      try {
+        arr = JSON.parse(raw.toString('utf8')).map(decodeArrayElement);
+      } catch (err) {
+        this.#stats.readFailures++;
+        this.log?.debug?.(`disk-spill array-decode failed for ${url}: ${err.message}`);
+        this.#removeEntry(url, entry);
+        return undefined;
+      }
+      this.#stats.restored++;
+      return arr;
+    }
     this.#stats.restored++;
-    return { ...entry.meta, content };
+    return { ...entry.meta, content: raw };
   }
 
   has(url) { return this.#index.has(url); }
--- a/packages/core/src/cache/byte-lru.js
+++ b/packages/core/src/cache/byte-lru.js
@@ -1,7 +1,15 @@
-// Hand-rolled byte-budget LRU cache. Map insertion order is LRU order;
-// .get() deletes and re-sets to move an entry to MRU. Synchronous by design —
-// no logger calls, no awaits inside mutation paths, so callers can log
-// after .set() returns without risking a mid-state event-loop yield.
+// Two-tier cache used by asset discovery:
+//   ByteLRU — byte-budget in-memory LRU; Map insertion order = LRU order.
+//   DiskSpillStore — on-disk overflow tier. RAM evictions spill here; lookups
+//   fall back to disk before refetching from origin.
+// All operations are synchronous; callers (network intercept, ByteLRU.set)
+// cannot yield to the event loop mid-op. Per-entry size is capped at 25MB
+// upstream so disk I/O latency is bounded.
+
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import crypto from 'crypto';
 
 const DEFAULT_PER_ENTRY_OVERHEAD = 512;
 
@@ -32,10 +40,10 @@ export class ByteLRU {
   set(key, value, size) {
     if (!Number.isFinite(size) || size < 0) return false;
 
-    // Reject oversize BEFORE touching any existing entry — a failed oversize
-    // set on an existing key must not evict the prior (valid) entry.
+    // Reject oversize BEFORE touching any existing entry — a failed set on an
+    // existing key must not evict the prior (valid) entry.
     if (this.#max !== undefined && size > this.#max) {
-      if (this.onEvict) this.onEvict(key, 'too-big');
+      if (this.onEvict) this.onEvict(key, 'too-big', value);
       return false;
     }
 
@@ -54,7 +62,7 @@ export class ByteLRU {
       this.#bytes -= rec.size;
       this.#map.delete(oldestKey);
       this.#stats.evictions++;
-      if (this.onEvict) this.onEvict(oldestKey, 'lru');
+      if (this.onEvict) this.onEvict(oldestKey, 'lru', rec.value);
     }
 
     return true;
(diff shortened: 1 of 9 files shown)


pranavz28:
# PER-7795 Disk-Spill — Real-Build Verification Report

Branch: `feat/per-7795-max-cache-ram`
Verification run: 2026-04-24

All runs below used a local `web-t` token against `test-pranav-8a4f5725` and
the two local origin servers built for this verification:

| Server | Port | Description |
|--------|-----:|-------------|
| `heavy-assets-server.mjs` | 9200 | 30 × 1 MB CSS, no delay (TC1, TC2) |
| `heavy-assets-slow.mjs`   | 9201 | 60 × 1 MB CSS, 150 ms delay each (TC3, TC4, TC6, TC8) |

## Test case matrix

| # | Case | Flag | Fixture | Build # | Result |
|---|------|------|---------|--------:|--------|
| 1 | Baseline (no flag) | *(none)* | heavy @ 9200 | [#392](https://percy.io/9560f98d/web/test-pranav-8a4f5725/builds/49052568) | PASS — no spill dir created, no disk telemetry |
| 2 | Aggressive spill | `--max-cache-ram 1` (→25 MB clamp) | heavy @ 9200 | [#394](https://percy.io/9560f98d/web/test-pranav-8a4f5725/builds/49052667) | PASS — 7 spills, `Cache eviction active` info log fired once |
| 3 | Live disk-dir inspection | `--max-cache-ram 1` | slow @ 9201 | [#397](https://percy.io/9560f98d/web/test-pranav-8a4f5725/builds/49052891) | PASS — peaked at 37 files / 35 MB on disk; byte sizes match spilled resources |
| 4 | Disk restore (concurrency:1) | `--max-cache-ram 1`, `discovery.concurrency: 1` | slow @ 9201 | [#401](https://percy.io/9560f98d/web/test-pranav-8a4f5725/builds/49053065) | PASS — 97 spills, **96 `cache disk-hit`** log lines |
| 5 | Cleanup on stop | — | — | — | PASS — `$(os.tmpdir())/percy-cache-*` removed after every run |
| 6 | Telemetry payload | `--max-cache-ram 1`, `concurrency: 1` | slow @ 9201 | [#403](https://percy.io/9560f98d/web/test-pranav-8a4f5725/builds/49053289) | PASS after fix — all 8 `disk_*` fields populated |
| 7 | End-to-end build health | verified on #392, #403, #405 | — | — | PASS — every completed comparison has diff ratio `0.0000` vs baseline |
| 8 | Moderate cap | `--max-cache-ram 40`, `concurrency: 1` | slow @ 9201 | [#405](https://percy.io/9560f98d/web/test-pranav-8a4f5725/builds/49061405) | PASS — 82 spills, 81 disk-hits, disk peaked at 22 entries / 21 MB |

## Specific resource → disk-file evidence (TC3, live)

Process PID `27680`, spill directory
`/var/folders/4g/nq932q454bjbwx6xnp_zprdh0000gn/T/percy-cache-27680-d47eaa7d`,
captured 4.8 s into the run:

```
-rw-r--r-- 1  3060 bytes    1   ← http://127.0.0.1:9201/          (3,060-byte HTML)
-rw-r--r-- 1  1000000 bytes  2   ← http://127.0.0.1:9201/slow-0.css
-rw-r--r-- 1  1000000 bytes  3   ← http://127.0.0.1:9201/slow-1.css
-rw-r--r-- 1  1000000 bytes  4   ← http://127.0.0.1:9201/slow-3.css
-rw-r--r-- 1  1000000 bytes  5   ← http://127.0.0.1:9201/slow-2.css
-rw-r--r-- 1  1000000 bytes  6   ← http://127.0.0.1:9201/slow-4.css
-rw-r--r-- 1  1000000 bytes  7   ← http://127.0.0.1:9201/slow-5.css
```

Filenames are the `DiskSpillStore` monotonic counter (1, 2, 3, …), and
sizes match each resource exactly (3 060 B for the HTML, 1 000 000 B for
each CSS). By later iterations the directory held 37 files totalling
~35 MB. After `percy.stop()` the directory was gone.

## Telemetry payload (TC6, temporary probe)

With a one-line debug probe added inside `sendCacheSummary` and reverted
before commit, the payload shipped to Percy was:

```json
{
  "cache_budget_ram_mb": 25,
  "hits": 24,
  "misses": 220,
  "evictions": 97,
  "peak_bytes": 25016372,
  "final_bytes": 24015860,
  "entry_count": 25,
  "oversize_skipped": 0,
  "disk_spill_enabled": true,
  "disk_spilled_count": 97,
  "disk_restored_count": 96,
  "disk_spill_failures": 0,
  "disk_read_failures": 0,
  "disk_peak_bytes": 36003060,
  "disk_final_bytes": 36003060,
  "disk_final_entries": 37
}
```

All eight `disk_*` fields are populated and match the discovery log.

## Bug found and fixed during verification

The first TC6 run produced `disk_spill_enabled: false` with every
`disk_*` field zeroed, even though the log showed 97 spills and 96
disk-hits. Root cause: `percy.stop()` calls `discovery.end()` (which runs
`diskStore.destroy()` and `delete percy[DISK_SPILL_KEY]`) before
`sendCacheSummary()` reads those references.

Fix (commit `2422b01d`): the `end` handler now snapshots
`diskStore.stats` (plus `ready`) onto `stats.finalDiskStats` before
`destroy()`; `sendCacheSummary` prefers the live store and falls back to
the snapshot, so both in-discovery tests and post-discovery real builds
populate the telemetry correctly. Two new specs cover each branch.

## Commands reproduced here for reference

```sh
# TC1 — baseline
zsh -ic 'web-t && node packages/cli/bin/run.cjs snapshot \
  /tmp/heavy-snapshots.yml --verbose'

# TC2 — aggressive spill (clamped to 25 MB)
zsh -ic 'web-t && node packages/cli/bin/run.cjs snapshot \
  --max-cache-ram 1 /tmp/heavy-snapshots.yml --verbose'

# TC4 / TC6 — serial disk restore with full telemetry
zsh -ic 'web-t && node packages/cli/bin/run.cjs snapshot \
  -c /tmp/percy-serial.yml --max-cache-ram 1 \
  /tmp/slow-multi-snapshot.yml --verbose'

# TC8 — 40 MB moderate cap
zsh -ic 'web-t && node packages/cli/bin/run.cjs snapshot \
  -c /tmp/percy-serial.yml --max-cache-ram 40 \
  /tmp/slow-multi-snapshot.yml --verbose'
```

`/tmp/percy-serial.yml`:

```yaml
version: 2
discovery:
  concurrency: 1
```

pranavz28:
# PER-7795 Disk-Spill — Re-Verification After Refactor

**Branch:** `feat/per-7795-max-cache-ram` @ `2c02f945`
**Re-verification date:** 2026-04-29
**Initial verification:** 2026-04-24 (PR comment on #2192, builds #392–#405)

This re-run was prompted after the round of review fixes:
`aeae5da4` (leak fix + dedupe + 0o700) →
`f92aee60` (try/catch + microtask hop parity) →
`56998885`/`fcd83737` → `2874ad8c` → `e41db794` → `2c02f945`.

The point is to confirm none of those churns regressed disk behavior, plus
to lock in coverage for edge cases the original verification didn't drive
explicitly.

## Standalone disk-tier suite — 37 / 37 PASS

Built `@percy/core` (`yarn build`) and ran a self-contained Node script
that imports the **real** `ByteLRU`, `DiskSpillStore`, and
`lookupCacheResource` from `dist/` and exercises every code path against
the real filesystem (no mocks, no Chromium).

### V1 ByteLRU bounded mode + onEvict (5/5)
- ✓ cap not exceeded → no eviction
- ✓ LRU eviction fires on overflow
- ✓ oversize set returns false
- ✓ oversize fires `too-big` eviction reason
- ✓ cache still contains b/c after oversize (rejected entry doesn't displace existing)

### V2 DiskSpillStore — 0o700 perms + counter filenames (7/7)
- ✓ `createSpillDir()` returns path under `os.tmpdir()`
- ✓ `createSpillDir()` uses `percy-cache-` prefix
- ✓ store ready after construction
- ✓ **dir mode is `0o700`** — verified via `fs.statSync(dir).mode & 0o777 === 0o700`
- ✓ counter-based filenames (1, 2, … — no URL data)
- ✓ destroy removes the spill dir

### V3 Disk content byte-for-byte (3/3)
- ✓ disk file SHA-256 matches input content (1 MB random bytes)
- ✓ restored content SHA matches input
- ✓ restored carries metadata (url, mimetype)

### V4 Replace + best-effort unlink + bytes accounting (7/7)
- ✓ size=1 after replace
- ✓ bytes correctly reclaimed on overwrite
- ✓ only one file remains on disk after replace
- ✓ **5× same-URL saves → size === 1** (regression guard)
- ✓ 5× same-URL saves → bytes === 800 (no doubling)
- ✓ 5× same-URL saves → spilled counter === 5 (counter advanced)
- ✓ 5× same-URL saves → only one file on disk (prior files unlinked)

### V5 Failure paths — write/read/init (8/8)
- ✓ init failure (mkdirSync throws) → `ready=false`
- ✓ init failure → `set()` returns false (short-circuit)
- ✓ init failure → `spillFailures` stays 0 (no write attempted)
- ✓ write failure (writeFileSync throws) → `set()` returns false
- ✓ write failure → `spillFailures` incremented
- ✓ read failure (file unlinked behind index) → `get()` returns undefined
- ✓ read failure → `readFailures` incremented
- ✓ read failure → index entry self-healed (`has()` returns false after)

### V6 lookupCacheResource — RAM → disk fallthrough (3/3)
- ✓ RAM miss → disk hit returns spilled resource
- ✓ `cache disk-hit:` debug log fires
- ✓ snapshot-local resource short-circuits both cache and disk

### V7 ByteLRU + DiskSpillStore wired together — full eviction roundtrip (4/4)
- ✓ ByteLRU evicts oldest when over cap
- ✓ onEvict callback writes to disk
- ✓ `lookupCacheResource` restores from disk
- ✓ restored stat counter advances

## CI on `2c02f945`

| Job | Result |
|---|---|
| Lint, Typecheck, Build, Semgrep, CodeQL | ✓ pass |
| Linux all 17 packages | ✓ pass |
| **Test @percy/core (Linux)** | **✓ 764/764 specs PASS, 100% coverage** |
| `byte-lru.js` line / branch / function / statement | 100 / 100 / 100 / 100 |
| `discovery.js` line / branch / function / statement | 100 / 100 / 100 / 100 |
| `percy.js` line / branch / function / statement | 100 / 100 / 100 / 100 |

## Real-build evidence (carried forward from 2026-04-24)

The April 24 run on `6a1a3d0e` already proved the integrated disk-spill
flow against live Percy. None of the post-review commits touched the
read/write paths — only the `'end'` cleanup ordering, telemetry egress
helper, and tests — so those builds remain authoritative for the e2e
behavior:

| TC | Build | Evidence |
|---|---|---|
| Baseline | [#392](https://percy.io/9560f98d/web/test-pranav-8a4f5725/builds/49052568) | no spill dir created |
| `--max-cache-ram 1` (clamped) | [#394](https://percy.io/9560f98d/web/test-pranav-8a4f5725/builds/49052667) | 7 spills, `Cache eviction active` info log |
| Live spill-dir inspection | [#397](https://percy.io/9560f98d/web/test-pranav-8a4f5725/builds/49052891) | mid-flight `ls` showed counter files 1–7 with byte sizes matching the spilled CSS resources; peaked at 37 files / 35 MB; **dir gone after stop** |
| Disk restore (2 snapshots, concurrency:1) | [#401](https://percy.io/9560f98d/web/test-pranav-8a4f5725/builds/49053065) | 97 spills + **96 `cache disk-hit` log lines** |
| Telemetry payload | [#403](https://percy.io/9560f98d/web/test-pranav-8a4f5725/builds/49053289) | All 8 disk fields populated: `spilled=97, restored=96, peak_bytes=36 003 060` |
| 40 MB cap | [#405](https://percy.io/9560f98d/web/test-pranav-8a4f5725/builds/49061405) | 82 spills, 81 disk-hits, peaked at 22 entries / 21 MB |
| Build comparisons | all of the above | every completed comparison has diff_ratio `0.0000` vs baseline |

## What changed since 2026-04-24 — and why those changes don't risk this evidence

| Commit | Change | Impact on disk path |
|---|---|---|
| `aeae5da4` | `try/finally` around `browser.close()` in 'end' handler | Cleanup runs even when browser teardown throws — strict improvement; tested via existing "calls diskStore.destroy" spec that now also asserts `stats.finalDiskStats` is captured |
| `aeae5da4` | `mkdirSync` mode 0o700 | Multi-tenant safety; verified above (dir mode `700`) |
| `aeae5da4` | `ByteLRU.values()` removed | Was unreferenced anywhere in src/ |
| `aeae5da4` | `Percy.sendCacheTelemetry()` extraction | Single egress for both cache_summary (awaited) and cache_eviction_started (fire-and-forget); test parity preserved |
| `f92aee60` | sendCacheSummary outer try/catch + fireCacheEventSafe `Promise.resolve().then(...)` microtask hop | Strict pre-refactor parity — payload-build never fails stop, eviction-loop hot path never blocks on payload serialization |
| `e41db794` | `.catch(() => {})` on the microtask chain | Defensive against Node 14's unhandled-rejection-as-fatal mode if the inner catch arm itself throws |

None of these commits touched `DiskSpillStore.set/get/destroy`,
`createSpillDir`, the `onEvict` callback wiring, or the cache lookup
fallthrough logic, so the April 24 e2e evidence remains valid.

## Confidence

The combination of:

1. **CI green** (764 specs, 100% coverage) on the head commit;
2. **37/37 standalone assertions** against the real built dist, including
   SHA roundtrip on real fs;
3. **Six real Percy builds** previously finalized with all-zero diff
   ratios;
4. **No source change to disk-spill primitives** since that real-build
   evidence was captured;

…means the disk-spill feature is verified working end-to-end on the
current head. The failing local Mac Chromium runs we saw today are an
unrelated browser-launch infra issue on this machine — they don't
implicate the cache code.

packages/core/src/discovery.js
```diff
@@ -391,30 +392,174 @@ export async function* discoverSnapshotResources(queue, options, callback) {
...
+// Cache lookup shared by the network intercept path. RAM miss falls through
+// to the disk tier; read failures return undefined so the browser refetches.
+// Also resolves the array-valued root-resource shape used for multi-width
+// DOM snapshots, regardless of which tier returned it.
+export function lookupCacheResource(percy, snapshotResources, cache, url, width) {
+  let resource = snapshotResources.get(url) || cache.get(url);
+  const disk = percy[DISK_SPILL_KEY];
+  if (!resource && disk) {
+    resource = disk.get(url);
+    if (resource) {
```
amandeepsingh333, line 437: **HIGH: Disk-hit resources are never re-promoted to RAM**

When `disk.get(url)` returns a resource, it's served to the caller but never re-inserted into the RAM ByteLRU. This means every subsequent access to the same spilled resource hits `readFileSync` again.

For resources evicted early but still frequently needed (e.g., a shared `styles.css` used across 50 snapshots), every access is a synchronous disk read. The typical LRU disk-spill pattern promotes disk-tier hits back to RAM:

```js
if (!resource && disk) {
  resource = disk.get(url);
  if (resource) {
    cache.set(url, resource, entrySize(resource));
    percy.log.debug(`cache disk-hit (promoted): ${url} ...`);
  }
}
```

Without this, the disk tier degrades from an overflow cache to a read-every-time-from-disk pattern for warm entries.

packages/core/src/cache/byte-lru.js
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  // Returns true on success; false on any failure so caller falls back to drop.
+  // Overwrites prior spill for the same URL — a fresh discovery write wins.
+  set(url, resource) {
+    if (!this.#ready) return false;
+
+    let content = resource?.content;
+    if (content == null) return false;
+    if (!Buffer.isBuffer(content)) {
```
amandeepsingh333: **HIGH: Array-valued root resources (multi-width snapshots) silently fail to spill**

The cache can hold array-valued entries (array of root resources for different widths). When evicted from ByteLRU, `onEvict` passes the array to `DiskSpillStore.set(key, value)`. Here, `resource?.content` on an array returns `undefined`, so `content == null` triggers and the method returns `false`. The root resource is silently dropped.

Multi-width root snapshots can never be spilled to disk — they're always lost on eviction. Either:
1. Handle arrays here (serialize elements separately or as JSON), or
2. Add an explicit check with a log so the silent failure is visible:
```js
if (Array.isArray(resource)) {
  this.log?.debug?.(`disk-spill skip (array resource): ${url}`);
  return false;
}
```

packages/core/src/cache/byte-lru.js
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// Handles the two Percy cache-entry shapes: single resource, or array of
+// roots captured at multiple widths (see discovery.js parseDomResources).
+export function entrySize(resource, overhead = DEFAULT_PER_ENTRY_OVERHEAD) {
+  if (Array.isArray(resource)) {
+    return resource.reduce((n, r) => n + (r?.content?.length ?? 0) + overhead, 0);
+  }
+  return (resource?.content?.length ?? 0) + overhead;
+}
```
amandeepsingh333, line 126: **MEDIUM: `entrySize` uses `content.length` which undercounts bytes for multi-byte string content**

`content.length` on a string returns character count, not byte count. For multi-byte UTF-8 content (CJK, emoji), `'\u{1F600}'.length === 2` but `Buffer.byteLength('\u{1F600}') === 4`.

If `content` arrives as a string (which is possible — `DiskSpillStore.set` explicitly handles `!Buffer.isBuffer(content)`), byte budget accounting will undercount, potentially allowing the cache to exceed the RAM cap.

Consider `Buffer.byteLength(content)` or assert that content is always a Buffer at this point.

packages/core/src/discovery.js
```diff
@@ -476,8 +616,32 @@ export function createDiscoveryQueue(percy) {
...
+                  // ByteLRU fires onEvict('too-big') for oversize entries;
+                  // the oversize_skipped stat + debug log live there.
+                  cache.set(r.url, r, entrySize(r));
+                } else {
                   cache.set(r.url, r);
+                  // Track bytes unconditionally so peak is accurate;
+                  // one-shot warn emission is gated by warningFired.
+                  const stats = percy[CACHE_STATS_KEY];
+                  stats.unsetModeBytes += entrySize(r);
+                  if (!stats.warningFired && stats.unsetModeBytes >= warnThreshold) {
```
amandeepsingh333, line 653: **MEDIUM: `unsetModeBytes` over-counts when the same URL is saved multiple times**

When `maxCacheRam` is unset and the plain `Map` is used, `cache.set(r.url, r)` overwrites any prior value, but `stats.unsetModeBytes` unconditionally adds the new entry size without subtracting the old one. If the same URL is saved across multiple snapshots (common for shared CSS/JS), the counter inflates and may trigger the 500MB warning prematurely.

```js
if (cache.has(r.url)) {
  stats.unsetModeBytes -= entrySize(cache.get(r.url));
}
cache.set(r.url, r);
stats.unsetModeBytes += entrySize(r);
```

packages/core/src/config.js
```diff
@@ -369,6 +369,10 @@ export const configSchema = {
       disableCache: {
         type: 'boolean'
       },
+      maxCacheRam: {
+        type: ['integer', 'null'],
+        minimum: 0
+      },
```
amandeepsingh333, line 378: **LOW: Schema allows `maxCacheRam: 0`**

`minimum: 0` means `maxCacheRam: 0` passes validation. The clamp in discovery.js catches it (bumps to 25MB), but users might expect `0` to mean "unbounded" or "disabled" rather than "25MB cache." Consider `minimum: 1` or explicitly treating `0` as null/unbounded.

packages/core/src/cache/byte-lru.js
```diff
@@ -0,0 +1,216 @@
...
+      try { content = Buffer.from(content); } catch { return false; }
+    }
+
+    // Counter-based filename keeps URL-derived data out of path.join —
+    // avoids any path-traversal surface even though sha256 would be safe.
+    const filepath = path.join(this.dir, String(++this.#counter));
+
+    try {
+      fs.writeFileSync(filepath, content);
+    } catch (err) {
```
amandeepsingh333, line 185: **NIT: Sync `writeFileSync` on the eviction hot path**

When the cache is full, every `ByteLRU.set()` in the eviction `while` loop triggers a synchronous disk write here. For builds with thousands of resources and a tight cap, this serializes disk I/O on the Node event loop.

The 25MB per-entry ceiling bounds the worst case for a single write, but aggregate across a batch of evictions can add up. Worth monitoring post-deploy via `cache_summary` telemetry — if `disk_spilled_count` is high and builds are slow, async writes with a queue would be the natural next step.

amandeepsingh333 (changes requested):
## Review: `--max-cache-ram` bounded asset-discovery cache with disk-spill tier\n\nWell-structured PR — clean separation between ByteLRU (RAM tier), DiskSpillStore (disk tier), and the discovery wiring. Error handling is thorough with graceful fallbacks at every I/O boundary. Commit history is bisectable. Test coverage is extensive (44 unit + 8 integration specs).\n\nHowever, there are two high-severity issues that undermine the disk-spill tier's value, plus a few medium findings.\n\n### Summary\n\n| Severity | Finding |\n|----------|----------|\n| HIGH | Disk-hit resources never re-promoted to RAM — repeated disk reads on warm entries |\n| HIGH | Array-valued root resources (multi-width snapshots) silently fail to spill |\n| MEDIUM | `entrySize` uses `content.length` — undercounts bytes for multi-byte strings |\n| MEDIUM | `unsetModeBytes` over-counts on duplicate URL saves — premature 500MB warning |\n| LOW | Config schema allows `maxCacheRam: 0` which silently clamps to 25 |"

packages/core/src/discovery.js
pranavz28: Fixed in `9ed10d0d` — `lookupCacheResource` now promotes disk hits back to RAM and deletes the disk entry, so a hot URL pays the `readFileSync` cost once. Test at `byte-lru.test.js` ("promotes a disk hit back to RAM and frees the disk slot") asserts the second lookup is a RAM hit (no `readFileSync` calls).

packages/core/src/cache/byte-lru.js
pranavz28: Fixed in `9ed10d0d` — `DiskSpillStore.set` now serializes multi-width root arrays via `JSON.stringify` with binary content base64-encoded, and `get` decodes them back. Tests cover the byte-for-byte roundtrip, array-decode self-heal, and `JSON.stringify`-throws refusal (circular ref).

packages/core/src/cache/byte-lru.js
pranavz28: Fixed in `9ed10d0d` — `entrySize` now uses `Buffer.byteLength(content)` for string content via a shared `contentBytes` helper. Test asserts `entrySize({ content: "\u{1F600}" }, 0) === 4` (4 UTF-8 bytes, not 2 string units) and the same for CJK.

packages/core/src/discovery.js
pranavz28: Fixed in `9ed10d0d` — `saveResource` now subtracts `entrySize(prior)` before adding `entrySize(r)` in the unset-cap branch. Regression guard at `discovery.test.js` asserts `unsetModeBytes` is unchanged after a same-URL re-save.

packages/core/src/config.js
pranavz28: Fixed in `9ed10d0d` — schema `minimum` bumped from 0 to 1. Zero has no useful semantics (null = unbounded, `--disable-cache` = off), and rejecting at schema time avoids the silent clamp to 25MB.

packages/core/src/cache/byte-lru.js
pranavz28: Acknowledged — tracked in the PR rollout plan. `disk_spilled_count` and `disk_peak_bytes` already ride the `cache_summary` event; if the post-deploy dashboard shows high spill counts on slow builds, the natural next step is an async write queue. Keeping sync for now to bound semantics.

packages/core/src/cache/disk-spill.js
```diff
@@ -0,0 +1,164 @@
...
+      ...this.#stats,
+      currentBytes: this.#bytes,
+      peakBytes: this.#peakBytes,
+      entries: this.#index.size
+    };
+  }
+
+  #pathFor(url) {
+    const hex = crypto.createHash('sha256').update(url).digest('hex').slice(0, FILENAME_HEX_LEN);
+    return path.join(this.dir, hex);
```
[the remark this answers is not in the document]
pranavz28: Stale finding — `packages/core/src/cache/disk-spill.js` no longer exists. The file was consolidated into `packages/core/src/cache/byte-lru.js` in `8e289f52`, and the path-traversal concern itself was eliminated by switching to monotonic counter-based filenames (no URL-derived data in `path.join`) in commit `34*`. Closing as resolved.

packages/core/src/cache/disk-spill.js
```diff
@@ -0,0 +1,164 @@
...
+    this.#index.delete(url);
+    try { fs.unlinkSync(entry.path); } catch { /* best-effort */ }
+  }
+}
+
+// Build a unique per-run temp directory path. Not created here — the store's
+// constructor performs the mkdir so failures funnel through the same handler.
+export function createSpillDir(prefix = 'percy-cache') {
+  const suffix = `${process.pid}-${crypto.randomBytes(4).toString('hex')}`;
+  return path.join(os.tmpdir(), `${prefix}-${suffix}`);
```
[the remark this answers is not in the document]
pranavz28: Stale finding — `packages/core/src/cache/disk-spill.js` no longer exists. The file was consolidated into `packages/core/src/cache/byte-lru.js` in `8e289f52`, and the path-traversal concern itself was eliminated by switching to monotonic counter-based filenames (no URL-derived data in `path.join`) in commit `34*`. Closing as resolved.

amandeepsingh333 (approved):

PR #2210, opened 2026-05-04
feat(core): direct gRPC client for Maestro view-hierarchy resolver (Phase 2.2)

Sriram567:
## Summary

**Phase 2.2 perf optimization for the Maestro Android element-region resolver.** Replaces the JVM-cold-start `maestro --udid <serial> hierarchy` shell-out (~9s p50) with a direct gRPC call to `dev.mobile.maestro` on device port 6790. Maestro CLI shell-out is preserved as a graceful fallback for local dev, schema drift, and connection-class errors.

**Customer-visible impact:** element-region resolution latency drops from ~9s to <100ms per snapshot. A 10-snapshot flow with element regions now adds ~1s instead of ~90s.

### What changed

- Vendored `maestro_android.proto` from upstream (`mobile-dev-inc/Maestro` @ `bc8bde1b`, CLI release `cli-2.5.1`) at `packages/core/src/proto/`.
- Added `@grpc/grpc-js@^1.14.3` + `@grpc/proto-loader@^0.8.0` (both Node-14-safe, used on BS hosts).
- New `runGrpcDump` + `classifyGrpcFailure` in `maestro-hierarchy.js` with:
  - Module-scope `Client` cache per `(host, port)`; eager close + evict on connection-class failure.
  - Two-tier deadline: 250ms healthy-call (`CallOptions.deadline`) + 2s circuit-breaker (`Promise.race`) — bounds the blast radius of [grpc-node#2620](https://github.com/grpc/grpc-node/issues/2620) "stuck CONNECTING" + [#2285](https://github.com/grpc/grpc-node/issues/2285) "lying READY state."
  - 14-row error-classification table (schema-class vs connection-class). Schema-class skips fallback per R3.
  - Reuses existing `sliceXmlEnvelope` + `flattenNodes` — Maestro's `ViewHierarchyResponse.hierarchy` is the same UiAutomator XML format (verified at source level via `ViewHierarchy.kt`).
- New `discoverGrpcPort`: `MAESTRO_GRPC_PORT` env preferred, falls back to parsing `adb -s <serial> forward --list` for `tcp:6790`.
- New `PERCY_MAESTRO_GRPC=0` kill switch — short-circuits to maestro CLI fallback, logged at WARN on every dump call so rollback state is observable.
- Healthcheck dirty bit: `/percy/healthcheck` exposes `maestroHierarchyDrift: { code, reason, firstSeenAt }` on first schema-class failure (closes the silent-drift gap that bit `PERCY_LABELS`).
[...]

PR #2217, opened 2026-05-07
feat(core): cross-platform Maestro view-hierarchy resolver + screenshot filePath relay + fallback observability

Sriram567:
> **Update (2026-05-12):** This PR now also bundles a relay-side change for the Maestro screenshot path coupling. See [§ Also bundled below](#also-bundled). Companion SDK PR: [percy/percy-maestro-app#3](https://github.com/percy/percy-maestro-app/pull/3).

## Summary

Replaces the per-snapshot ~9s JVM-cold-start `maestro hierarchy` CLI shell-out with **direct transport to Maestro's view-hierarchy services**, on both platforms. Element-region resolution now runs as a stateless RPC against Maestro's existing channel rather than spawning a second Maestro flow context — fixing the production gRPC-session-collision failure mode that drops snapshots whenever a Maestro flow is in progress (which is always, during element-region resolution).

| Platform | Primary (env-conditional) | Fallback chain |
|---|---|---|
| **Android** | gRPC POST `MaestroDriver/viewHierarchy` on `127.0.0.1:$PERCY_ANDROID_GRPC_PORT` | maestro CLI shell-out → adb `uiautomator dump` |
| **iOS** | HTTP POST `/viewHierarchy` on `127.0.0.1:$PERCY_IOS_DRIVER_HOST_PORT` | maestro CLI shell-out (`--driver-host-port`) |

Both transports drop in alongside the same two-slot `maestroHierarchyDrift` envelope on `/percy/healthcheck` (per-platform schema-drift surface) and follow the same three-class error taxonomy: schema-class → no fallback + drift bit; channel-broken → fallback + cache eviction; contention-class → fallback (skipping CLI) + cache PRESERVED.

Self-hosted customers without the env var injection see **zero behavior change** — the env var presence is the deployment-shape signal; absence routes to the existing maestro CLI primary + adb fallback.

This PR consolidates work originally scoped across three branches (`feat/ios-element-regions-maestro-hierarchy` PR #2202 closed, `feat/grpc-element-region-resolver` PR #2210 to be closed, and the iOS HTTP work on this branch).

### iOS HTTP path (already shipped on the prior 14 commits of this branch)

`runIosHttpDump` POSTs `{appIds: [], excludeKeyboardElements: false}` to Maestro's iOS XCTestRunner `/viewHierarchy` endpoint. At cli-2.0.7 the runner detects the AUT itself via `RunningApp.getForegroundApp()` (Maestro PR #2365) — no bundleId discovery, no SDK changes, no realmobile control-plane changes. SpringBoard-only responses (older Maestro) route to maestro-CLI fallback.

### Android gRPC path (newly absorbed from PR #2210)

`runAndroidGrpcDump` calls `MaestroDriver/viewHierarchy` directly via `@grpc/grpc-js` over the same transport Maestro CLI uses internally. Vendored proto at `packages/core/src/proto/maestro_android.proto` (upstream SHA `bc8bde1b`, cli-2.5.1).

**Three-class error taxonomy** (refined from PR #2210's two-class scheme during deepen-plan):

| Class | gRPC status codes | Fallback | Cache eviction | Drift bit |
|---|---|---|---|---|
| Schema-class | INVALID_ARGUMENT, FAILED_PRECONDITION, OUT_OF_RANGE, UNIMPLEMENTED, DATA_LOSS, decoder-failure | None | No | Yes |
| Channel-broken | UNAVAILABLE, INTERNAL, CANCELLED | maestro CLI → adb | Yes | No |
| Contention-class | DEADLINE_EXCEEDED, RESOURCE_EXHAUSTED, ABORTED | adb only (skip CLI — would queue behind same flow) | **No** | No |

The contention-class refinement is the key correctness win: timeout under live Maestro flow load is *backpressure evidence*, not channel-breakage. Evicting the channel on every timeout would force a TCP+HTTP/2+TLS reconnect (~50-200ms cost) that buys nothing, because the underlying agent is still busy. Keeping the channel cached lets the next call reuse the queue position.

**Symmetric timeouts**: `GRPC_HEALTHY_DEADLINE_MS = 1500` + `GRPC_CIRCUIT_BREAKER_MS = 5000`, parity with the iOS HTTP path's existing values. Outer `Promise.race` is defense-in-depth against historical `grpc-node#2620` (closed in 1.9.11).

**Per-`Percy` cache scope**: the `grpcClientCache` Map is constructed on the `Percy` instance and disposed in `stop()` — matches `@percy/core`'s established ownership pattern for every other long-lived resource (server, browser, queues, client). Module-global state would leak channels between concurrent `Percy` instances and create shutdown races.

**Shutdown race handling**: `percy.stop()` sets `cache.shutdownInProgress = true` before closing channels. Any in-flight `runAndroidGrpcDump` that hits `CANCELLED` returns `{kind:'unavailable', reason:'shutdown'}` — no fallback chain on a tearing-down process.

### Two rollback knobs (deliberate)

- **Slow rollback** (deploy-shape change): remove `PERCY_ANDROID_GRPC_PORT` (or `PERCY_IOS_DRIVER_HOST_PORT`) injection in mobile/realmobile.
- **Fast rollback** (in-process kill switch): set `PERCY_MAESTRO_GRPC=0` in the BS appPercy env. Skips gRPC on next CLI restart without coordinated mobile-side deploy. The 3am-page response.

## What's in this bundle

The iOS HTTP path commits already on this branch plus 5 new commits absorbing PR #2210:

| Unit / origin | Commit | Description |
|---|---|---|
| 2026-05-07-002 plan Unit 1 | `68e67db5` | vendor maestro_android.proto + add @grpc/grpc-js deps |
| 2026-05-07-002 plan Units 2/3/7 | `73459be6` | direct gRPC Android resolver with three-class taxonomy |
| 2026-05-07-002 plan Unit 5 | `135f0ea3` | per-Percy gRPC client cache lifecycle + shutdown race |
| 2026-05-07-002 plan Units 2/3/4/5 | `bf5d1f55` | 28 new test specs covering all dispatch + classification paths |
| 2026-05-07-002 plan Unit 6 | `e8583bdf` | concurrent-access merge gate harness with p95<1200ms / p99<2000ms gate |

(Commits `a1bd69da` and earlier are the iOS HTTP path documented in the prior PR description history.)

## Testing

- **807 of 807 specs run, 28 added by this absorption**. All 28 new specs pass:
  - `Unit / maestro-hierarchy / Android gRPC primary path` — 28 specs across `classifyGrpcFailure`, `runAndroidGrpcDump`, cache reuse + per-instance isolation, `closeGrpcClientCache`, and `dump({platform:'android'})` dispatch.
- **27 pre-existing failures unchanged from master** (Install Chromium spy setup × 21, runDoctorOnFailure × 5, IPv4/IPv6 AggregateError flake on api.test.js:655 — none related to this branch).
- **All 80+ existing maestro-hierarchy specs pass** (Android adb path, iOS HTTP path, parity tests, two-slot drift envelope tests).

## BrowserStack App Automate validation (2026-05-07)

End-to-end validation of the iOS HTTP path on real BS realmobile/mobile hosts (full plan + results in local `docs/plans/2026-05-07-001-feat-bs-validation-maestro-ios-http-resolver-plan.md`):

| Platform | BS build | Percy build | Snapshots | Resolver outcome |
|---|---|---|---|---|
| Android | `5439a79e` (passed) | [#49501406](https://percy.io/9560f98d/app/MaestroPercyApp-99cd1aa2/builds/49501406) | 2 | `unavailable / multi-device-no-serial` (resolved by mobile PR #13206 commit `ddac377`) |
| Android (post-mobile-fix) | `8ed1a6c8` (passed) | [#49502511](https://percy.io/9560f98d/app/MaestroPercyApp-99cd1aa2/builds/49502511) | 2 | `dump-error / fallback-dump-exit-137` (the gRPC-contention root cause this PR's gRPC-direct path fixes) |
| iOS | `41ebf750` (passed) | [#49501844](https://percy.io/9560f98d/app/MaestroPercyApp-99cd1aa2/builds/49501844) | 2 | `dump-error / http-non-json-content-type` — **iOS HTTP primary path was exercised in production**; schema-drift envelope correctly classified the response. |

The validation surfaced the BS-side env-injection gaps (mobile and realmobile didn't pass `ANDROID_SERIAL` / `PERCY_IOS_DRIVER_HOST_PORT` to the Percy CLI process). Companion commits landed on the BS-side PRs:
- [browserstack/mobile#13206](https://github.com/browserstack/mobile/pull/13206) commit `ddac377` — `ANDROID_SERIAL` injection in `cli_manager.rb`.
- [browserstack/realmobile#9840](https://github.com/browserstack/realmobile/pull/9840) commit `62a0f7e` — `PERCY_IOS_DRIVER_HOST_PORT` + `PERCY_IOS_DEVICE_UDID` injection in `maestro_session.rb start_app_percy`.

- **What to monitor/search**
  - Logs (Percy CLI per-session): `[percy] iOS HTTP schema-drift: ...` / `[percy] gRPC viewHierarchy schema-class failure (...)` — proto drift signals.
  - Metrics: `/percy/healthcheck` `maestroHierarchyDrift` envelope. Alert when `.android` or `.ios` slot populates.
- **Validation checks (queries/commands)**
  - `curl http://127.0.0.1:<cli_port>/percy/healthcheck` mid-session → expect `maestroHierarchyDrift: { android: null, ios: null }`.
  - Run the merge-gate harness against a real Android device after BS-side `PERCY_ANDROID_GRPC_PORT` injection lands: `MAESTRO_ANDROID_TEST_DEVICE=<serial> PERCY_ANDROID_GRPC_PORT=<port> node packages/core/test/integration/maestro-hierarchy-concurrent.harness.js` — gate is p95 < 1200ms AND p99 < 2000ms across 100 iterations.
- **Expected healthy signal(s)**
  - `[percy] dump took Nms via grpc (M nodes)` (Android, env-set) or `[percy] dump took Nms via maestro-http` (iOS, env-set).
  - Both drift slots stay `null`.
  - Either drift slot populates → ops investigates Maestro version + considers proto refresh PR.
  - Sustained Android contention (p99 > 2000ms across N sessions) → file open-circuit backoff follow-up; meanwhile flip `PERCY_MAESTRO_GRPC=0` for fast rollback.
  - Window: 7 days post-merge of this PR + the BS-side env-injection PRs.
  - Owner: percy-cli oncall.

## Pending follow-ups (out of scope for this PR)

- Companion `browserstack/mobile` PR injecting `PERCY_ANDROID_GRPC_PORT` (analog of `realmobile` commit `62a0f7e` for iOS). Without this, the Android gRPC primary stays dormant in production.
- Maestro version compatibility: iOS validation surfaced `http-non-json-content-type` from this realmobile deployment's Maestro version. Needs follow-up to determine missing content-type header vs endpoint shape vs version mismatch.
- Open-circuit backoff if sustained contention shows p99 > 2000ms in production (deferred per plan; not needed unless the harness/production reveals it).
- Close [percy/cli#2210](https://github.com/percy/cli/pull/2210) and delete `feat/grpc-element-region-resolver` after this PR merges to master.

---

[Compound Engineering v2.54.0](https://github.com/EveryInc/compound-engineering-plugin)
🤖 Generated with Claude Opus 4.7 (1M context, ultrathink) via [Claude Code](https://claude.com/claude-code)

---

## Also bundled: screenshot relay `filePath` field (commit `36f9c56c`)

The `/percy/maestro-screenshot` relay now accepts an optional `filePath` field on the request body. When present, the relay reads the file at that path directly (after the same `realpath` + per-platform session-root `startsWith` security check used by the legacy glob path) and skips the glob. When absent, the legacy glob runs unchanged.

This eliminates a hidden coupling between BrowserStack-infra `SCREENSHOTS_DIR` conventions and the relay's hardcoded glob pattern — surfaced by a recent regression (BS build `0444158…`) where a BS-infra patch put screenshot files one directory level too shallow and snapshots silently failed with `"Snapshot command was not called"`.

### Version-skew matrix

The companion SDK change lives in [`percy/percy-maestro-app#3`](https://github.com/percy/percy-maestro-app/pull/3) (`1.0.0-beta.2`). The two PRs are independently deployable:

| CLI version | SDK version | Behaviour |
|---|---|---|
| New CLI (≥ 1.31.11-beta.1) | Old SDK | Old SDK doesn't post `filePath`; legacy glob runs; **no change**. |
| Old CLI | New SDK (≥ 1.0.0-beta.2) | New SDK's version-gate detects old CLI via `x-percy-core-version`; SDK falls back to relative `SCREENSHOT_NAME`; **no change**. |
| New CLI | New SDK | SDK posts absolute path under BS session root; relay reads directly; **path-coupling eliminated**. |

### Tests

9 new jasmine specs inside the existing `describe('/percy/maestro-screenshot')` block in `packages/core/test/api.test.js` cover the contract:

- Happy paths (Android + iOS) — file under per-platform session root, base64 content matches the on-disk bytes.
- Shape rejection (400) — non-string, non-absolute, length > 1024.
- Missing file (404) — `realpath` fails; same error shape as the legacy glob's 404.
- Path-traversal / outside-session-root (404) — `realpath` succeeds but `startsWith(sessionRoot)` fails. Same code path defeats symlink-escape — the `realpath` of a symlink target gets the prefix check.
- Cross-sessionId subtree (404) — file exists, but at a different sessionId's session root.
- Empty `filePath` string — treated as absent; falls through to legacy glob. Old SDKs that emit the field unconditionally are safe.

All 22 maestro-screenshot specs pass (13 existing + 9 new). 28 unrelated pre-existing failures unchanged from master.

### Operational note

The BS-infra v4 `SCREENSHOTS_DIR=…/screenshots` patch in `mobile-pr/android/maestro/scripts/maestro_runner.rb` becomes a back-compat safety net once both PRs deploy. It is still required for any customer running an older SDK against this CLI. Removal can be tracked separately once older SDK versions are retired from customer test suites.

### Plan

- Plan: `docs/plans/2026-05-11-001-feat-sdk-cli-screenshot-path-decoupling-plan.md` (Unit 1 only — Units 2/3 land in the SDK PR).
- Origin: `docs/brainstorms/2026-05-11-maestro-screenshot-path-and-grpc-observability-requirements.md` (R2/R3/R4/R6 from Issue 1).

---

## Also bundled: hierarchy fallback observability (commit `ccd96f56`)

Surfaces every primary→fallback transition in the resolver cascade at the default log level, and extends the `/percy/healthcheck` `maestroHierarchyDrift` envelope with three resolver activity counters. **Observability only** — the cascade behaviour is unchanged. R7/R8 from the 2026-05-11 origin doc `docs/brainstorms/2026-05-11-maestro-screenshot-path-and-grpc-observability-requirements.md` (local only); R10 (decide whether to skip gRPC on BS, share channel, or accept the cascade) explicitly deferred until production data flows through this surface.

### Envelope shape (additive — existing fields preserved)

Slots stay `null` until any resolver activity. When populated:

```
maestroHierarchyDrift: {
  android: {
    // Activity counters — always present when slot is non-null.
    lastFailureClass: 'schema-class' | 'channel-broken' | 'contention-class' | 'other' | null,
    fallbackCount:    <cumulative primary→fallback transitions this process>,
    succeededVia:     'grpc' | 'maestro-cli' | 'adb' | 'none' | null,

    // Schema-class drift bit — only after a schema-class failure on this platform.
    code?, reason?, firstSeenAt?
  },
  ios: { ... succeededVia: 'maestro-http' | 'maestro-cli-fallback' | 'none' | null, ... }
}
```

The existing `{code, reason, firstSeenAt}` triple keeps its set-once-per-platform semantics. Ops dashboards reading those fields keep working unchanged.

### Info-level transition log shape

```
[percy] hierarchy: <primary> failed (<class>: <reason>) → falling back to <next>
```

Five primary→fallback edges, all bumped from `log.debug` → `log.info`:

```
[percy] hierarchy: grpc failed (contention-class: grpc-contention-deadline_exceeded) → falling back to adb
[percy] hierarchy: grpc failed (channel-broken: grpc-channel-broken-unavailable) → falling back to maestro-cli
[percy] hierarchy: maestro-cli failed (other: maestro-exit-1) → falling back to adb
[percy] hierarchy: maestro-http failed (channel-broken: http-econnrefused) → falling back to maestro-cli-fallback
[percy] hierarchy: maestro-http failed (other: out-of-range-port-99999) → falling back to maestro-cli-fallback
```

Internal classification details inside `runAndroidGrpcDump` stay at `log.debug` — the transition signal lives at the `dump()` cascade boundary, one info line per transition. Success-path lines (`dump took Nms via grpc`) stay debug; warn-level drift-fires lines (`gRPC viewHierarchy schema-class failure`) stay warn.

### Why this is worth shipping now (and refutes the "always fails" theory)

A recent BS validation observed `PERCY_ANDROID_GRPC_PORT` injected correctly, `adb forward` succeeded, yet the cascade fell back to adb. The session's working theory was *channel-broken because `dev.mobile.maestro` is single-client and the running flow holds it*. 

Investigation of upstream Maestro source ([`MaestroDriverService.kt`](https://github.com/mobile-dev-inc/maestro/blob/main/maestro-android/src/androidTest/java/dev/mobile/maestro/MaestroDriverService.kt)) refutes that:

```kotlin
NettyServerBuilder.forPort(port)
    .addService(Service(uiDevice, uiAutomation))
    ...
    .build().start()
```

No custom executor → default Netty worker pool. `viewHierarchy()` has no `synchronized`, no mutex, no single-flight pattern. The gRPC service accepts concurrent clients by construction. So the single observed fallback was most likely `DEADLINE_EXCEEDED` (contention-class, surfacing as `UiAutomation` backpressure under the running flow's call rate) or a transient connection-level event — *not* an architectural problem with the gRPC-direct design.

This observability surface is what tells contention-class apart from channel-broken apart from schema-class without another speculation pass. One BS build with this commit on the host produces hard data for R10.

### Tests

24 new specs in `packages/core/test/unit/maestro-hierarchy.test.js` inside a new `resolver activity counters + transition logs (R7/R8)` describe block. Covers:

- Initial state (no resolver activity → slots null, back-compat with existing assertions).
- Android success: gRPC primary, maestro-cli primary (no gRPC env), adb fallback after CLI failure.
- Android gRPC cascade: contention-class → adb, channel-broken → CLI, channel-broken → CLI failure → adb, schema-class no-fallback.
- Cumulative counters: two consecutive contention failures → `fallbackCount: 2`. Sticky lastFailureClass when later call succeeds via gRPC.
- iOS: HTTP success, HTTP schema-class, HTTP connection-fail → CLI, no-aut-tree → CLI, env-missing.
- Adb-unavailable on Android (`other` class, `succeededVia: 'none'`).
- Cross-platform isolation (iOS activity doesn't touch android slot and vice versa).
- Reset helper clears activity counters too.
- All 6 R7 info-level transition log lines.

Two existing tests (`connection-class failure does NOT flip the drift bit` and `SpringBoard-only response does NOT flip the drift bit`) updated to acknowledge the slot now populates with activity counters on those paths while drift-bit fields stay absent.

All 22 existing `/percy/maestro-screenshot` relay tests, 80+ existing maestro-hierarchy specs, and the cross-platform parity harness pass unchanged. Pre-existing failures (`Install Chromium` × 21, `runDoctorOnFailure` × 5, server-disabled × 1) unchanged from master.

### Plan

- Plan: `docs/plans/2026-05-12-001-feat-grpc-hierarchy-fallback-observability-plan.md` (local only) (R7/R8 from Issue 2 — observability only; R10 deferred).
- Origin: `docs/brainstorms/2026-05-11-maestro-screenshot-path-and-grpc-observability-requirements.md` (local only).

### Post-deploy validation (deferred per `feedback_defer_real_device_testing.md`)

One BS Android build + one BS iOS build exercising element-based regions, each verified against `[percy] hierarchy:` info-line in `/var/log/browserstack/percy_cli.<sid>.log` and the populated `maestroHierarchyDrift` slot via `curl http://127.0.0.1:<cli_port>/percy/healthcheck` mid-session. Recipe lives in the `percy-maestro` project memory.

Once that data lands, R10's follow-up plan can pick the right architectural fix (skip gRPC on BS / share channel / accept cascade) with evidence rather than speculation.

---

## Also bundled: Maestro tag-dim authority via PNG header (commit `9960741a`)

Tightens the `/percy/maestro-screenshot` relay so the screenshot PNG bytes are the single authoritative source for `tag.width` / `tag.height`. Eliminates the dependency on host-side OS tooling (`adb shell wm size` on Android, `xcrun devicectl` on iOS 17+) that the prior auto-injection plan needed. R5 from origin doc `docs/brainstorms/2026-05-22-maestro-auto-inject-device-metadata-requirements.md`.

### What changes in the CLI

`parsePngDimensions(buffer)` is added at module scope in `packages/core/src/api.js` — a 25-line helper that validates the PNG signature (`89 50 4E 47 0D 0A 1A 0A`) and reads `IHDR` width/height as big-endian uint32 at offsets 16/20. Wired into the `/percy/maestro-screenshot` handler immediately after the file buffer is read:

```js
let pngDims = parsePngDimensions(fileContent);
let tag = req.body.tag || { name: 'Unknown Device', osName: 'Android' };
if (!tag.name) tag.name = 'Unknown Device';
if (pngDims) {
  if (typeof tag.width !== 'number' || tag.width <= 0 || isNaN(tag.width)) tag.width = pngDims.width;
  if (typeof tag.height !== 'number' || tag.height <= 0 || isNaN(tag.height)) tag.height = pngDims.height;
}
```

Non-destructive fill — a customer who provides explicit `tag.width` / `tag.height` (e.g., a self-hosted Maestro user manually setting `PERCY_SCREEN_WIDTH` / `PERCY_SCREEN_HEIGHT`) wins. The PNG-derived value fills only when the customer omits the field, leaves it zero, or sends NaN.

### Why this matters

The prior plan attempted to derive width/height host-side (`adb shell wm size` on Android, `xcrun devicectl device info displays` on iOS 17+) and inject them as `PERCY_SCREEN_WIDTH` / `PERCY_SCREEN_HEIGHT` Maestro `-e` flags. Two operational gaps surfaced:

1. **iOS host coverage gap.** `xcrun devicectl` is iOS 17+ only; the BS iOS host pool serves iOS 14/15/16 devices where the tool simply does not exist. The prior plan's iOS path was demonstrably broken on host `185.255.127.52` (`iPhone 14 16.4`, no `devicectl` binary). Customers on those devices would have hit an unworkable carve-out.
2. **Customer-experience purity.** The auto-injection meant we'd still need `PERCY_SCREEN_WIDTH` / `PERCY_SCREEN_HEIGHT` in the customer's YAML for self-hosted Maestro (Maestro Cloud, local CI, customer device labs). The PNG-header derivation works for any environment that produces a PNG — universal.

### Validation (Unit 5) — proven on BOTH platforms

**iOS**: BS Maestro v2 build `44b3fa29d08fbfea95901589be2577fd203578f2` on host `185.255.127.52` (iPhone 14, iOS 16.4, **no `xcrun devicectl`**) → Percy build [#50073878](https://percy.io/9560f98d/app/MaestroPercyApp-99cd1aa2/builds/50073878). Customer YAML had **zero `PERCY_SCREEN_*` env vars**. Comparison tag:

```
name:        iPhone 14-16.4
width:       1170
height:      2532
os-name:     iOS
os-version:  16.4
```

Exact iPhone 14 PNG dimensions, derived purely from the relay's PNG-header parse. Proves the architectural pivot end-to-end on the host that was the original failure case — the prior plan's iOS path required `xcrun devicectl` which doesn't exist on this host.

**Android**: BS Maestro v2 build `24780eb0463b7c8b0f0f7b6dd5029011167cb681` on host `31.6.63.67` (Pixel 8, Android 14) → Percy build [#50077082](https://percy.io/9560f98d/app/MaestroPercyApp-99cd1aa2/builds/50077082). Customer YAML had **zero `PERCY_SCREEN_*` env vars**. Comparison tag:

```
name:        Google Pixel 8
width:       1080
height:      2400
os-name:     Android
os-version:  14
```

Exact Pixel 8 native resolution from PNG header. Confirms the same relay-side derivation works on Android with no host-side `adb shell wm size` shell-out (Unit 2 removed that).

**Smoke-test note on host overlay**: validation required bumping `@percy/core/package.json` on the host from `1.30.0` to `1.31.15-beta.0` so the SDK's `coreSupportsFilePath` version gate passes and the `filePath` field is forwarded with each upload. This is purely a manual-overlay artifact of testing pre-merge against a Nix-pinned `@percy/cli@1.30.0`; once this PR ships and the BS Nix derivation is bumped to a beta that includes these changes, the version gate passes automatically and no manual edit is needed.

### Tests

7 new specs in `packages/core/test/api.test.js` under the `/percy/maestro-screenshot` describe — `makePngHeader(width, height)` helper constructs minimal 24-byte PNG buffers. Coverage: happy-path fill, customer-pinned override, partial fill, non-PNG signature skip, truncated file skip, width=0 defense, `filePath` field parity with multipart upload.

### Companion changes (host-side reverts, separate PRs)

- **mobile#13206** (`feat/maestro-percy-integration`, commit `5cccff4fcf`): Android `setup_percy_device_metadata` helper no longer shells out to `adb wm size`. Returns only `{device_name, os_version}`. `adb` shell-out, `Timeout.timeout(3)` wrap, and `require "timeout"` deleted.
- **realmobile#9840** (`feat/maestro-percy-ios-integration`, commit `cee25cd051`): iOS `percy_device_metadata` helper no longer parses `device_config['device_resolution']` or version-gates iOS 17+. Returns only `{device_name, os_version}`. iOS 14/15/16 carve-out removed.
- **percy-maestro-app `main`** (commit `3b32225`): SDK no longer emits an iOS pre-CoreDevice WARN log. README "Device metadata auto-detection" section simplified — the 3-row coverage matrix + iOS 17+ carve-out collapses into one clean "What's auto-derived / What you still set" section.

### Plan

- Plan: `docs/plans/2026-05-23-001-refactor-maestro-screen-dims-via-png-header-plan.md` (local only — supersedes the dim-injection part of the 2026-05-22 plan).
- Origin: `docs/brainstorms/2026-05-22-maestro-auto-inject-device-metadata-requirements.md` (local only).

### Smoke-test pickup notes

- BS Android hosts run a Nomad `code_update` job that auto-reverts the mobile clone to the latest release tag every ~50 min. To iterate on the feat branch within that window, redeploy and trigger the BS build immediately. Captured in local memory `project_nomad_update_revert_android_hosts.md`.
- The host's Nix-pinned `@percy/cli@1.30.0` lacks transitive deps that master added (`@percy/monitoring`, `@grpc/grpc-js`, `busboy`, `fast-xml-parser`, etc.). For pre-merge smoke testing, manually `scp` these from a local cli build into the host's nix store `node_modules/` (and chmod a+rX). After this PR ships and the BS Nix derivation is bumped to a beta off master, the deps arrive automatically. Captured in `project_host_deps_solved_by_cli_bump.md`.


packages/core/src/maestro-hierarchy.js
```diff
@@ -0,0 +1,1260 @@
+// Cross-platform view-hierarchy resolver for /percy/maestro-screenshot element regions.
```
Shivanshu-07, line 1: **🟡 P2 — Verify the drift-bit clearing semantics and `PERCY_MAESTRO_GRPC=0` kill switch coverage.**

Two requests on the 1,260-LOC core resolver:

### 1. Drift-bit life cycle on Maestro version upgrade

The `maestroHierarchyDrift` envelope has set-once-per-platform `{code, reason, firstSeenAt}` semantics for schema-class failures. Two questions:

- **Does the bit ever clear?** If a customer hits a schema-class failure on Maestro v1.39.13, then upgrades to v2.5.1 mid-process (rolling deploys, manual restarts), does the drift bit clear or does it haunt the slot until process restart? The PR description says set-once but doesn't specify lifecycle.
- **Is the bit per-`Percy` instance or per-process?** If per-process module-global state, this contradicts the per-instance discipline applied to `grpcClientCache`. If per-instance, it clears on `stop()` — fine, but worth a spec assertion.

A 1-line spec like *"upgrading Maestro version between snapshots → next failure populates a fresh slot"* would lock this in.

### 2. `PERCY_MAESTRO_GRPC=0` kill switch placement

The PR advertises this as the 3am-fast-rollback path. Want to verify three things:

- The check happens **early** in `runAndroidGrpcDump` (before any gRPC channel work), so `=0` is genuinely cheap on the cascade.
- The same env var also gates the **iOS HTTP path** (`runIosHttpDump`) — readers of the PR description might expect a separate `PERCY_MAESTRO_HTTP=0`, but the docstring says one kill switch covers both. Confirm in code.
- The env var is read on **every dump call**, not cached at `Percy` construction. Otherwise toggling it during an incident requires a process restart, which contradicts the "fast rollback" framing in mobile #13206 and realmobile #9840's docs.

If any of those don't hold, the rollback story is weaker than the BS-side PRs assume. Worth a 3-line spec covering each.

### 3. Optional refactor: split the file

`maestro-hierarchy.js` is 1,260 LOC after this PR. Inside it: gRPC client + gRPC classifier + iOS HTTP client + iOS classifier + drift-envelope writer + cascade orchestrator + cache lifecycle. Each is independently testable. After this PR ships, consider:

```
maestro-hierarchy/
  index.js              ← public dump() + cascade
  android-grpc.js       ← gRPC client + classifier
  ios-http.js           ← HTTP client + classifier
  drift.js              ← envelope + counters
  cache.js              ← per-Percy lifecycle
```

Not a blocker — the test suite is thorough enough that the bundling is acceptable today. Just a "next time someone touches this" pointer.

Shivanshu-07 (commented):
## Cross-Repo Maestro+Percy Stack Review (3 of 4)

Reviewed alongside companions: browserstack/mobile#13206, browserstack/realmobile#9840, percy/percy-maestro-app#3.

**2 inline file-level comments below + cross-repo coordination notes.** No P1 blockers — the design is genuinely well-thought-through. Suggestions cluster around verifying invariants the PR description promises and tightening the rollback story so it's actionable in practice.

### Top findings

1. **Drift-bit lifecycle on Maestro version upgrade** — set-once semantics need a spec assertion that the bit clears appropriately (or doesn't, by design — but pin it).
2. **`PERCY_MAESTRO_GRPC=0` kill switch wiring** — verify it's read on every dump call (not cached), gates both Android gRPC AND iOS HTTP paths, and is checked early in `runAndroidGrpcDump`. This is the entire "fast rollback" story across the BS-side PRs.
3. **Multi-instance isolation of `grpcClientCache`** — the docstring promises per-`Percy` scope; one spec assertion locks in the invariant.

### Cross-PR coordination notes

This PR is the **center of a 4-PR stack** (mobile #13206, realmobile #9840, percy-maestro-app #3). Some observations from reviewing them together:

- **Deploy order:** This PR should land FIRST. The BS-side env vars (`PERCY_ANDROID_GRPC_PORT`, `PERCY_IOS_DRIVER_HOST_PORT`, `ANDROID_SERIAL`, `MAESTRO_BIN`) are inert against the old CLI. The 2026-05-07 production validation confirms what happens when the BS side lands without the corresponding env injection (`multi-device-no-serial`, `fallback-dump-exit-137`).
- **Kill switch wiring on BS side:** Neither mobile #13206 nor realmobile #9840 explicitly wires `PERCY_MAESTRO_GRPC=0` into `cli_env`. During a real incident, the on-call needs to know exactly where to set this — host-level env? per-session config? puma worker restart required? A short runbook in this PR's description (or the percy oncall handbook) closes the gap.
- **`maestroHierarchyDrift` envelope dashboards:** The envelope is great but no Honeycomb saved query / Grafana panel lands with this PR. Worth pre-creating the alert *"drift slot populates → page percy-cli oncall"* so the 7-day validation window is observable, not just measurable.

### What's exceptional about this PR

The PR is unusually well-engineered for a 5,768-LOC change:

- **Three-class error taxonomy** (schema / channel-broken / contention) is correct. Contention-class preserving the channel cache is the key correctness win — the alternative (reconnect on every timeout) burns ~50-200ms for nothing because the underlying agent is still busy.
- **Per-`Percy` cache scope** for the gRPC client + explicit `shutdownInProgress` flag — correct concurrency isolation.
- **Symmetric Android/iOS timing** (`GRPC_HEALTHY_DEADLINE_MS = 1500` matches existing iOS HTTP values) — eliminates a class of "why is platform X timing out earlier" questions.
- **Two-slot drift envelope** preserves existing `{code, reason, firstSeenAt}` semantics so existing dashboards keep working.
- **Two rollback knobs** (slow deploy-shape vs fast in-process kill switch) — this is exactly the operational design real systems need. (Just needs the kill-switch verification noted above.)
- **28 new specs + p95<1200ms/p99<2000ms concurrent-access merge gate** — performance assertions baked into CI is the right move.
- **Vendored proto** with SHA + cli version reference so upstream Maestro proto drift surfaces in a code review rather than a 3am page.
- **Bundled scope is justifiable** — the PR description's "Also bundled" sections (screenshot relay + observability) are individually shippable but consolidated because each is small AND the test suite covers them in isolation. Not the pattern for every PR, but defensible here.

### Process feedback (low priority)

The bundling of PRs #2202 (closed) + #2210 (to-be-closed) + iOS HTTP + screenshot relay + observability into one is reviewer-heavy (5,768 LOC across 33 files). For the next refactor of comparable size, consider:

1. Sub-PRs per **unit of rollback** — if you can roll back screenshot relay independently from gRPC, they probably warrant separate PRs.
2. A **PR-stack tool** (graphite, ghstack) — gives reviewers small commits to review against each other, while still landing as one merge.

Not a blocker for THIS PR; the test suite is thorough enough that the bundling is acceptable.

🤖 Multi-agent code review via [Claude Code](https://claude.com/claude-code)

packages/core/src/maestro-hierarchy.js
Sriram567: Thanks for catching this — answering each in turn:

**Does the bit clear?** No, by design. It's set-once-per-process for schema-class failures only (`first-seen wins`, see `setMaestroHierarchyDrift` at `maestro-hierarchy.js:506`). A Maestro version upgrade mid-process won't clear it — but it also won't be *overwritten*, so the operational signal "this process observed a schema-class drift" stays accurate across an in-process upgrade. The reset function at line 749 exists for test isolation only, not production rotation.

**Per-`Percy` instance or per-process?** Intentionally **module-scoped** (`maestro-hierarchy.js:187`, design rationale at top-of-file comment lines 54–60), diverging from `grpcClientCache` which is per-`Percy`. The reasoning: `grpcClientCache` holds a *resource* that must be torn down with its owning `Percy` (otherwise channels leak); drift is *observability state* that should be visible across any `Percy` constructed in this process. Two `Percy` instances in one CLI process sharing one drift bit is the correct behavior — schema drift is a property of the Maestro install, not of a snapshot session.

The set-once-per-process lifecycle is the property the existing two-slot spec coverage exercises. We'll add the explicit "version-change does not clear" assertion in the next refactor of this file.

On the file split (`android-grpc.js` / `ios-http.js` / `drift.js` / `cache.js` / `index.js`): 100% agree the file is at the upper end of "one module" — recording it as the natural next refactor. Holding for now because (a) the public surface is a single `dump()` + a small cluster of healthcheck accessors, so the bundled file is still navigable, and (b) the 28 new specs cover each concern in isolation — splitting won't change test discipline, only file-organization ergonomics.

Will pull both the lifecycle spec assertion and the file split in the same iteration, once the gRPC and iOS HTTP paths have a quarter of production data behind them and the right module boundaries are obvious from real call-graph weight.

Sriram567:
On the cross-repo coordination notes — two follow-up artifacts landing before canary, deliberately not in PR code:

1. **Kill-switch incident runbook** — adding a `PERCY_MAESTRO_GRPC=0` rollback section to the percy on-call handbook (host-level env, picked up on next CLI session start, no process restart). Will link the runbook entry back from this PR description before we open canary.
2. **`maestroHierarchyDrift` saved query / Grafana panel** — pre-creating the "drift slot populates → page percy-cli oncall" alert against the canary BS fleet so the 7-day validation window is observable from day one rather than reconstructable after the fact.

Both are doc/dashboard work, decoupled from the PR diff so they can iterate without re-reviewing this stack.

prklm10 (approved):

PR #2248, opened 2026-05-28
feat(core): self-hosted (non-BrowserStack) Maestro relay support

Sriram567:
## Summary

Adds non-BrowserStack ("self-hosted") support to the Maestro screenshot relay. Customers running `percy app:exec -- maestro test ...` on their own devices, CI, or device farms can now produce Percy builds without BS host injection.

**Architecture, in one sentence:** `sessionId` absent on `/percy/maestro-screenshot` is the self-hosted detection signal (it's a BrowserStack host-injected marker that lives nowhere else in the CLI); when absent, the relay resolves a file-find scope root from `PERCY_MAESTRO_SCREENSHOT_DIR` (process env, required, absolute existing directory — typically the customer's `maestro test --test-output-dir <DIR>` path), globs `<root>/**/<name>.png`, and applies the same realpath + trailing-separator prefix check that protects the BS path today. **The BrowserStack path is byte-identical** — the existing `/tmp/{sessionId}{_test_suite}` glob, manual-walker fallback, and security check all run exactly as before for any request with `sessionId`.

### What changed

- `packages/core/src/api.js` — `sessionId` is optional; `selfHosted = !sessionId` derives the mode. New scope-root resolution block validates `PERCY_MAESTRO_SCREENSHOT_DIR` (absolute + existing dir, 400 with actionable message otherwise, emitted before realpath so it isn't masked as a 404). File-find branches on `selfHosted` for the glob pattern; the realpath/prefix check is now rooted at the resolved `scopeRoot`. `filePath` is rejected in self-hosted mode (the SDK never emits it; honoring it against a caller-influenceable root would re-open arbitrary in-root reads).

- `packages/core/test/api.test.js` — new `self-hosted (sessionId absent)` describe locks the new branches: happy path (recursive glob find + payload omits sessionId + PNG-fill works), env-var validation (unset/relative/non-existent/file-not-dir → 400), `filePath` rejected, file-missing → 404, `name`-traversal rejected, coordinate region pass-through. The existing "rejects missing sessionId" test is updated to reflect the new behavior (missing sessionId + missing env var → 400 with the new actionable message).

### Security invariants (relocated, not removed)

- `SAFE_ID` on `name` (`^[a-zA-Z0-9_-]+$`) is load-bearing for the recursive glob — separators and traversal characters rejected.
- Trailing-separator guard preserved on the prefix check (prevents sibling-prefix bypass like `/x/.maestro` vs `/x/.maestro-secrets`).
[...]

PR #2254, opened 2026-05-29
feat(cli): auto-inject PERCY_SERVER for maestro test + PERCY_CLI_API alias

Sriram567:
## Summary

Two small UX-completion changes on top of #2248. Both remove the port-pairing footgun every native-SDK customer hits self-hosted, with **no** behavior change on the BrowserStack path.

### Unit 1 — `PERCY_CLI_API` alias in `@percy/cli-exec`
`percy app:exec` already exports `PERCY_SERVER_ADDRESS` into the child process. The Appium Python SDK reads `PERCY_CLI_API` instead. Today this works only by coincidence — both default to port 5338. With `--port` overrides (multi-device parallelism), the SDK silently points at the wrong CLI. This adds `env.PERCY_CLI_API = percy?.address()` alongside the existing exports so any SDK reading `PERCY_CLI_API` auto-aligns to whatever port `app:exec` picked.

### Unit 2 — auto-inject `-e PERCY_SERVER` for `maestro test` in `@percy/cli-app`
Maestro's GraalJS sandbox does not inherit the parent process's env, so `PERCY_SERVER_ADDRESS` is invisible to the SDK. Customer's only channel is Maestro `-e` flags. Today every self-hosted customer threads `-e PERCY_SERVER=http://localhost:<port>` through every Maestro invocation themselves, pairing ports manually for multi-device.

`app:exec` already knows the resolved port via `percy.address()`. The detector is conservative — basename match on argv[0], `test` subcommand, and a scan for any pre-existing `-e PERCY_SERVER=…` (customer override wins per R3). `npx maestro` and other shim invocations fall through to the explicit-`-e` pattern.

### Stacks on #2248

Base branch: `feat/self-hosted-maestro-percy`. Rebase target after #2248 merges: `master`. Both units are functionally independent of #2248's content but ship in the same release window as the customer-facing completion of self-hosted Maestro support.

## Why this is safe on BrowserStack

- BS doesn't wrap `maestro test` with `app:exec`. `maestro_runner.rb` runs maestro directly while the CLI runs in `app exec:start` mode separately, so cli-app's wrapped callback is unreachable on BS.
- Even if it ever did fire on BS, Maestro's documented precedence is `runFlow.env > -e > environment`, so BS's `runFlow.env` setting would still win.
- The `PERCY_CLI_API` export is unconditional but harmless: BS doesn't run the Appium Python SDK against an `app:exec`-spawned child either.

## Env-var conflict audit

Both names verified against the cli monorepo + every other Percy repo on disk:

- **`PERCY_CLI_API`** — exactly one existing reader (`percy-appium-python/percy/lib/cli_wrapper.py:10`), zero existing writers. No CLI internal use. Our write semantically matches the reader's existing default. cli-doctor's `env-audit.js` filters with `k.startsWith('PERCY_')` so the new var auto-surfaces; only the test fixture (`CLEAN_PERCY_ENV`) needs a one-line addition for hermeticity.
- **`PERCY_SERVER`** — read only inside Maestro's GraalJS, never by the CLI process itself. Four writer layers after this PR (`runFlow.env` > customer `-e` > our auto-inject > SDK default `http://percy.cli:5338`) layer cleanly with no overlap. R3 detector skip prevents double-injection.

## Testing

- **cli-app:** 16/16 specs pass, including 14 new scenarios for `maybeInjectMaestroServer` (happy path, customer override at multiple positions, basename match on absolute paths, `npx`/`python`/`appium` skip, `maestro hierarchy`/`list-devices` skip, percy-disabled skip, multi-device port isolation).
- **cli-exec:** new `PERCY_CLI_API` env-export test mirrors the existing `PERCY_SERVER_ADDRESS` and `PERCY_BUILD_URL` test patterns. Asserts the child sees `PERCY_CLI_API === PERCY_SERVER_ADDRESS === \"http://localhost:4567\"` when `--port=4567`.
- **cli-doctor:** `CLEAN_PERCY_ENV` fixture updated; 507/508 specs pass (the single failure is a pre-existing proxy-credentials network test in `connectivity.test.js`, unrelated).
- **Lint:** clean across all touched files.

- **What to monitor**
  - CLI relay logs (`/percy/maestro-screenshot` POSTs in self-hosted runs) — should land without `sessionId` in the request body for self-hosted, and continue to land WITH `sessionId` for BS App Automate.
  - Maestro flow logs — `[percy] Percy CLI healthcheck passed.` line should now appear without the customer setting `PERCY_SERVER`.
- **Validation checks**
  - Re-run the validation runbook at `percy-maestro/docs/solutions/best-practices/2026-05-27-self-hosted-maestro-validation.md` with the explicit `-e PERCY_SERVER=…` line **removed** from the `maestro test` command. Build #14-equivalent (`9×9=81`) should still produce 3 Unchanged snapshots against an approved baseline.
  - Smoke test a BS App Automate Maestro build on the same `bs-nixpkgs` cli pin; confirm snapshots produce identically to pre-deploy and that `sessionId=…` is present in every CLI relay log line.
  - Self-hosted: customer can run `percy app:exec -- maestro test flow.yaml` with no `-e PERCY_SERVER`, no manual env aliasing; SDK healthcheck passes; snapshots upload.
  - BrowserStack: byte-identical to pre-deploy — auto-inject never fires there.
- **Failure signal / rollback trigger**
  - Any BS App Automate Maestro build that previously passed reports `Snapshot command was not called` post-deploy → revert. (Should not be possible given the unreachable-on-BS analysis, but watch the first BS run.)
  - Self-hosted multi-device runs where two `app:exec --port` invocations show the same injected `PERCY_SERVER=` in their maestro logs → investigate multi-device isolation.
  - Window: 24h post-merge. Owner: same as #2248.

🤖 Generated with Claude Opus 4.7 via [Claude Code](https://claude.com/claude-code) + Compound Engineering v2.54.0

### DIFF
--- a/packages/cli-app/src/exec.js
+++ b/packages/cli-app/src/exec.js
@@ -1,3 +1,4 @@
+import path from 'path';
 import command from '@percy/cli-command';
 import * as ExecPlugin from '@percy/cli-exec';
 
@@ -15,6 +16,31 @@ export const start = command('start', {
   }
 }, ExecPlugin.start.callback);
 
+function hasExistingPercyServerFlag(args) {
+  for (let i = 2; i < args.length - 1; i++) {
+    if (args[i] === '-e' && /^PERCY_SERVER=/.test(args[i + 1])) return true;
+  }
+  return false;
+}
+
+// Maestro's GraalJS sandbox does NOT inherit the parent process's env,
+// so `PERCY_SERVER_ADDRESS` exported by app:exec is invisible to the
+// SDK. When wrapping `maestro test`, surface the CLI address through
+// Maestro's only env channel — `-e KEY=VALUE` flags — so the SDK
+// healthcheck can find the local CLI without the customer having to
+// pair ports manually. No-op when the customer already supplied their
+// own `-e PERCY_SERVER=...`.
+export function maybeInjectMaestroServer(ctx) {
+  const args = ctx?.argv;
+  if (!Array.isArray(args) || args.length < 2) return;
+  if (path.basename(args[0]) !== 'maestro') return;
+  if (args[1] !== 'test') return;
+  if (hasExistingPercyServerFlag(args)) return;
+  const addr = ctx.percy?.address();
+  if (!addr) return;
+  args.splice(2, 0, '-e', `PERCY_SERVER=${addr}`);
+}
+
 export const exec = command('exec', {
   description: 'Start and stop Percy around a supplied command for native apps',
   usage: '[options] -- <command>',
@@ -29,6 +55,9 @@ export const exec = command('exec', {
     projectType: 'app',
     skipDiscovery: true
   }
-}, ExecPlugin.default.callback);
+}, async function*(ctx) {
+  maybeInjectMaestroServer(ctx);
+  yield* ExecPlugin.default.callback(ctx);
+});
 
 export default exec;
--- a/packages/cli-app/src/index.js
+++ b/packages/cli-app/src/index.js
@@ -1,2 +1,2 @@
 export { default, app } from './app.js';
-export { exec, start, stop, ping } from './exec.js';
+export { exec, start, stop, ping, maybeInjectMaestroServer } from './exec.js';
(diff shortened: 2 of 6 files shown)


Sriram567:
Consolidated into #2261 (single self-hosted Maestro+Percy V1 PR — all 6 commits, same content, same UX outcomes). Closing this stacked PR in favor of the unified review surface.

PR #2262, opened 2026-06-02
fix(cli-command): exit non-zero on any signal-aborted run (PER-8678)

### LINKED ISSUE #2255: `percy exec` exits 0 during startup before launching wrapped command, causing false-positive CI success`
## The problem

`percy exec` exits successfully before launching the wrapped Cypress command, causing CI to report a passing E2E run even though no tests were executed.

We intermittently observe GitHub Actions builds reporting successful E2E execution while Cypress never starts running. The last Percy startup message is typically:

```text
[percy:core:install] Downloading Chromium 1300309...
```

After several minutes, the process exits successfully and the CI step is marked as passed.

Example:

```text
$ percy exec --parallel -- cypress run --config-file ./cypress/config.ts --browser chrome ${CY_RUN_PARAMS}

[percy:core:install] Downloading Chromium 1300309...

Done in 380.53s.
```

### Expected behavior

If Percy cannot complete startup and the wrapped command has not been launched, `percy exec` should exit non-zero.

At minimum, CI should fail when Cypress never starts running.

## Environment

* Node version: v24.16.0
* npm version: 11.13.0
* yarn version: 1.22.22
* `@percy/cli` version: 1.31.14
* Percy SDK: `@percy/cypress` 3.1.8
* OS version: Ubuntu 24.04.4 LTS
* CI: GitHub Actions
* Type of shell command-line interface: bash

## Details

The issue appears to occur during Percy startup, before the wrapped Cypress command is launched.

Observed behavior:

* Cypress never starts.
* The usual Percy log indicating execution of the wrapped command is never emitted.
* Percy logs `Downloading Chromium 1300309...`.
* After that, no further Chromium download progress is logged.
* Percy remains alive for several minutes while only emitting monitoring metrics.
* CPU utilization remains near zero for most of the runtime.
* Percy eventually logs:

  ```text
  [percy:monitoring] Stopped monitoring system metrics
  ```
* The process exits with status code 0.
* CI therefore reports a successful E2E run even though no tests were executed.

The relevant portion of the debug log looks like:

```text
[percy:core:install] Downloading Chromium 1300309...

... several minutes of monitoring output ...

[percy:monitoring] Stopped monitoring system metrics
Done in 380.53s.
```

One additional line appears immediately before process exit in CI jobs where the app under test is running locally using `webpack-dev-server`:

```text
[webpack-dev-server] Gracefully shutting down. To force exit, press ^C again. Please wait...
```

This line comes from our test environment rather than Percy itself, but it suggests that a graceful shutdown sequence is being triggered rather than an explicit failure path.

While investigating, we noticed that Percy CLI recently changed signal/shutdown handling in PR #2199, including explicit SIGINT/SIGTERM exit code mapping.

We have not identified the root cause, but the observed behavior seems consistent with Percy entering a shutdown path during startup and exiting successfully before the wrapped command is launched.

Regardless of the underlying cause, the externally observable behavior is that `percy exec` can report success even when the wrapped command never starts.

## Debug logs

Relevant excerpt:

```text
[percy:core:install] Downloading Chromium 1300309...
[percy:monitoring] cpuInfo: {"cores":4,"currentUsagePercent":1.0000000000000009,"cgroupExists":false} (5606ms)
[percy:monitoring] memoryInfo: {"currentUsagePercent":22.290926902438457,"totalMemory":16766423040} (2ms)

... monitoring output continues for several minutes ...

[percy:monitoring] Stopped monitoring system metrics (3798ms)
[webpack-dev-server] Gracefully shutting down. To force exit, press ^C again. Please wait...
Done in 380.53s.
```

Full debug logs are available if needed.

## Code to reproduce issue

We do not currently have a minimal reproduction.

The issue started recently without any relevant code changes on our side and is observed in GitHub Actions using:

```bash
percy exec --parallel -- cypress run --config-file ./cypress/config.ts --browser chrome .....
```

The key symptom is that Percy exits successfully during startup and Cypress is never launched.

pranavz28:
Fixes PER-8678 / closes percy/cli#2255

## Problem
`percy exec -- <cmd>` could exit **0 during startup before launching the wrapped command**, producing a false-positive CI success. Reported symptom: GitHub Actions reports a passing E2E run while Cypress never starts — last log is `Downloading Chromium 1300309...`, then minutes of `[percy:monitoring]` metrics, then exit 0.

## Root cause
`packages/cli-command/src/command.js`

The per-signal handler is registered for `SIGUSR1, SIGUSR2, SIGTERM, SIGINT, SIGHUP` and aborts with `new AbortError(signal, { signal, exitCode: 0 })`. But `beginShutdown` only records **SIGINT/SIGTERM** into `shutdownState.signal`. The exit-code gate in the runner's catch was keyed on `shutdownState.signal`:

```js
if (shutdownState.signal && err.signal && definition.exitOnError) { ... 130/143 ... }
```

So a **SIGHUP/SIGUSR1/SIGUSR2** abort during startup:
- skips that gate (`shutdownState.signal` is null), and
- skips the `if (err.exitCode !== 0)` gate (the AbortError carries `exitCode: 0`),

…so the runner returns normally and the process **exits 0** — even though `percy.yield.start()` was interrupted and the wrapped command was never spawned. In CI a stalled startup is typically torn down by SIGHUP; the sibling `[webpack-dev-server] Gracefully shutting down` line in the report corroborates an OS signal hitting the process group. PR #2199 fixed SIGINT/SIGTERM but left every other signal on the exit-0 path.

## Fix
- Extend `SIGNAL_EXIT_CODES` with `SIGHUP→129`, `SIGUSR1→138`, `SIGUSR2→140` (POSIX 128+signum).
- Key the exit-code gate off `err.signal` (the signal that actually aborted the run) instead of `shutdownState.signal`, with a non-zero `?? 1` fallback for any unmapped signal.

Any signal-aborted run now surfaces a non-zero exit code in production, so a startup interrupted before the wrapped command launches can never report CI success. Preserved: SIGINT→130 / SIGTERM→143, `PERCY_EXIT_WITH_ZERO_ON_ERROR=true` override, and `exitOnError:false` clean-unwind for programmatic/test usage.

## Testing
- Added a regression spec in `packages/cli-command/test/shutdown.test.js`: emits `SIGHUP` mid-run, asserts `process.exitCode === 129` (was `undefined`/0).
- `yarn workspace @percy/cli-command test` → **65/65 specs pass**.

### DIFF
--- a/packages/cli-command/src/command.js
+++ b/packages/cli-command/src/command.js
@@ -29,10 +29,19 @@ let activeContext = null;
 
 const DEFAULT_DRAIN_MS = 30_000;
 const HARD_EXIT_AFTER_FORCE_MS = 5_000;
-// POSIX-conventional signal exit codes: 128 + signal number.
-const EXIT_CODE_SIGINT = 130;
-const EXIT_CODE_SIGTERM = 143;
-const SIGNAL_EXIT_CODES = { SIGINT: EXIT_CODE_SIGINT, SIGTERM: EXIT_CODE_SIGTERM };
+// POSIX-conventional signal exit codes: 128 + signal number. SIGINT and
+// SIGTERM additionally engage drain semantics in beginShutdown; the
+// remaining signals only need an exit-code mapping so a signal-aborted
+// run never resolves as a false-positive success (PER-8678).
+const EXIT_CODE_SIGINT = 130; // 128 + 2
+const EXIT_CODE_SIGTERM = 143; // 128 + 15
+const SIGNAL_EXIT_CODES = {
+  SIGINT: EXIT_CODE_SIGINT,
+  SIGTERM: EXIT_CODE_SIGTERM,
+  SIGHUP: 129, // 128 + 1
+  SIGUSR1: 138, // 128 + 10
+  SIGUSR2: 140 // 128 + 12
+};
 
 // Begin or escalate drain. Idempotent on the same signal.
 function beginShutdown(signal) {
@@ -326,18 +335,26 @@ export function command(name, definition, callback) {
         else log.error(err);
       }
 
-      // Signal-driven shutdown — when SIGINT/SIGTERM was received
-      // during this run, set the signal-derived exit code
-      // (130 SIGINT / 143 SIGTERM) and return. We deliberately set
-      // `process.exitCode` and unwind cleanly rather than calling
+      // Signal-driven shutdown — when ANY process signal aborted this
+      // run, set the signal-derived exit code (130 SIGINT / 143 SIGTERM
+      // / 129 SIGHUP / 138 SIGUSR1 / 140 SIGUSR2, defaulting to 1 for
+      // any other signal) and return. Keying off `err.signal` (set on
+      // the AbortError by the per-signal handler) rather than
+      // `shutdownState.signal` is deliberate: beginShutdown only records
+      // SIGINT/SIGTERM into shutdownState for drain semantics, so the
+      // old `shutdownState.signal` gate let SIGHUP/SIGUSR1/SIGUSR2 fall
+      // through to the exitCode-0 AbortError path and exit 0 — a
+      // false-positive CI success when a run was aborted during startup
+      // before the wrapped command launched (PER-8678). We deliberately
+      // set `process.exitCode` and unwind cleanly rather than calling
       // `process.exit()`, so the surrounding catch's finally block (and
       // the lockfile's `process.on('exit')` handler) still run. The
       // event loop drains naturally because the unref'd drain/hard-exit
       // timers don't keep it alive. Tests with `exitOnError: false`
       // preserve the legacy clean-resolution behavior because
-      // AbortError carries exitCode:0 and the gate below is skipped.
-      if (shutdownState.signal && err.signal && definition.exitOnError) {
-        let signalCode = SIGNAL_EXIT_CODES[shutdownState.signal];
+      // AbortError carries exitCode:0 and this gate is skipped.
+      if (err.signal && definition.exitOnError) {
+        let signalCode = SIGNAL_EXIT_CODES[err.signal] ?? 1;
         let percyExitWithZeroOnError = process.env.PERCY_EXIT_WITH_ZERO_ON_ERROR === 'true';
         process.exitCode = percyExitWithZeroOnError ? 0 : signalCode;
         return;
--- a/packages/cli-command/src/command.js
+++ b/packages/cli-command/src/command.js
@@ -354,6 +354,12 @@ export function command(name, definition, callback) {
       // preserve the legacy clean-resolution behavior because
       // AbortError carries exitCode:0 and this gate is skipped.
       if (err.signal && definition.exitOnError) {
+        // Defensive non-zero fallback: every signal in the registered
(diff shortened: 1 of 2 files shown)


pranavz28:
## Code Review — PER-8678

Reviewed the diff (`command.js` +34/−11, `shutdown.test.js` +17). Traced the abort→catch path and the `percy exec` child-process flow to check for regressions.

### Summary
Fixes a false-positive CI success: `percy exec` could exit `0` when aborted by `SIGHUP`/`SIGUSR1`/`SIGUSR2` during startup. Adds exit-code mappings for those signals and re-keys the exit-code gate off `err.signal` (the signal that actually aborted) instead of `shutdownState.signal` (which only ever records SIGINT/SIGTERM).

### Review Table

| High | Security | Secrets / authz / input / IDOR / SQLi | N/A | No security surface touched |
| High | Correctness | Logic correct, edge cases | ✅ Pass | Bug analysis accurate; confirmed SIGHUP/USR1/USR2 previously hit the `exitCode:0` AbortError path |
| High | Correctness | Explicit error handling | ✅ Pass | `?? 1` fallback keeps unmapped signals non-zero; `PERCY_EXIT_WITH_ZERO_ON_ERROR` override preserved |
| High | Correctness | No race/concurrency issues | ✅ Pass | `AbortController.abort` idempotent (first reason wins); `err.signal` reflects the aborting signal |
| Medium | Testing | New code has tests | ✅ Pass | SIGHUP→129 regression spec added |
| Medium | Testing | Error/edge paths tested | ⚠️ Partial | Only SIGHUP tested; USR1/USR2 + `?? 1` fallback untested (fallback is `istanbul ignore`d). Acceptable — same code path |
| Medium | Testing | Existing tests pass | ✅ Pass | SIGINT→130 / SIGTERM→143 unaffected |
| Medium | Quality | Follows patterns / focused | ✅ Pass | Mirrors existing signal-test pattern + `process.exitCode`-then-`return` convention; tightly scoped |
| Low | Quality | Names / comments / deps | ✅ Pass | Comments explain *why*; no new deps |

### Notes (no blockers)

**A — Linux-specific signal numbers (cosmetic).** `SIGUSR1: 138` / `SIGUSR2: 140` use Linux signal numbers (10, 12); on macOS these are 30/31 → 158/159. Correct for Linux CI (the target environment), and `SIGHUP→129` + SIGINT/SIGTERM are universal. A one-line "Linux-conventional" comment would help, but not blocking.

**B — Broadened gate now intercepts any `.signal`-carrying error (verified-safe).** Changing `if (shutdownState.signal && err.signal ...)` to `if (err.signal && ...)` routes *any* error with a `.signal` property through the signal-exit branch. Traced `cli-exec/src/exec.js`: a normally-failing child exits via `exit(status, error, false)` where `error.signal` is undefined, so it still falls through to the regular exit-code path. The dominant `.signal` carrier is the parent's own `AbortError` — exactly what we want to catch. Every plausible case still yields **non-zero**, matching the PR's goal. No regression; flagging so it doesn't have to be re-derived.

### Verdict
**Approve.** Correct root-cause analysis, minimal well-targeted fix, comment-documented rationale, and a real regression test. The two notes are cosmetic/informational.

rishigupta1599 (approved):

PR #2300, opened 2026-06-17
ci: set ignore-scripts=false to fix executable build

pranavz28:
## Problem

The **Build Executables** workflow has failed for the last 3 releases (since `v1.32.0-beta.8`), most recently on [v1.32.0](https://github.com/percy/cli/actions/runs/27682535456/job/81873257738). The job dies in `scripts/executable.sh` at:

```
cp: ./build/*: No such file or directory
```

## Root cause

The release-automation PR added a root `.npmrc` with `ignore-scripts=true`. The Build Executables runner uses **Node 14 → npm 6**, and **npm 6 honors `ignore-scripts` even for explicit `npm run <script>`** (npm 7+ later relaxed this). So `npm run build_cjs` (`BABEL_ENV=dev babel packages -d build`) became a silent no-op — `build/` was never created — and the following `cp -R ./build/* packages/` failed.

`yarn build` two lines earlier is unaffected because it runs via `lerna` + Nx, whose task runner executes package scripts directly rather than through `npm run`.

Verified by elimination: same runner image, same Node version, and the exact `v1.32.0` source tree all build `build_cjs` fine; the only delta at the first failing tag was this `.npmrc`.

## Fix

Set `ignore-scripts=false` so `npm run build_cjs` executes on the npm-6 runner.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,4 +1,4 @@
-ignore-scripts=true
+ignore-scripts=false
 strict-ssl=true
 save-exact=true
 audit-level=high


rishigupta1599 (approved):

pranavz28:
## 🤖 Claude Code Review — `stack:pr-review`

**PR:** https://github.com/percy/cli/pull/2300 • **Head:** e4672cc5 • **Scope:** single-file CI config change (`.npmrc`)

**Summary:** Flips `ignore-scripts` from `true` → `false` in the repo-root `.npmrc` to fix the Build Executables release job, which has failed for the last 3 releases (since `v1.32.0-beta.8`).

### Review Table

| High | Security | No hardcoded secrets or credentials | Pass | Config-only |
| High | Security | Auth / input / IDOR / SQLi | N/A | No such surface in this change |
| High | Security | Supply-chain: install-script execution | Pass (note) | Re-enables lifecycle scripts on install — see Findings. Restores pre-existing behavior; repo-local & CI-only |
| High | Correctness | Logic correct, handles edge cases | Pass | On npm 6 (Node 14 / Build Executables), `ignore-scripts=true` suppresses even explicit `npm run`, so `npm run build_cjs` no-ops → empty `build/` → `cp` fails. `false` restores it |
| Medium | Testing | Existing tests still pass | Pass | 45/45 gating checks green at time of review |
| Medium | Quality | Follows existing patterns | Pass | `false` is npm's default and the repo's effective behavior before the `.npmrc` was added |
| Medium | Quality | Focused (single concern) | Pass | One line, one purpose |
| Low | Quality | Rationale documented | Pass | Captured in commit message + PR body |

### Findings

**1. Supply-chain hardening trade-off (informational, not blocking)**
- **File:** `.npmrc:1`
- **Issue:** `ignore-scripts=true` blocks dependency `preinstall`/`install`/`postinstall` scripts during `npm install`/`yarn install`. Setting it to `false` re-enables them.
- **Why it's acceptable here:**
  - This `.npmrc` is **repo-local** and governs only this monorepo's own dev/CI installs. npm excludes `.npmrc` from published tarballs, so **consumers of `@percy/*` are unaffected**.
  - `false` is npm's **default** and the repo's effective behavior for years before the `.npmrc` was introduced by the release-automation PR — this restores known-good state rather than introducing new risk.
  - As committed, the hardening flag **broke the executable release build**: npm 6 (Node 14 on the Build Executables runner) honors `ignore-scripts` even for explicit `npm run`, silently no-op'ing `npm run build_cjs` so `build/` was never produced and `cp -R ./build/* packages/` failed.
- **Alternative (if keeping the hardening is desired):** keep `ignore-scripts=true` and instead bypass npm's script runner in `scripts/executable.sh` (`npm run build_cjs` → `BABEL_ENV=dev ./node_modules/.bin/babel packages -d build`). Either approach is reasonable.

### Overall: ✅ Pass

**Verdict:** Approve — minimal, correct fix that unblocks the release executable build and restores the repository's long-standing default behavior. The only consideration is the supply-chain trade-off, documented above, which has no impact on published packages.

PR #2342, opened 2026-07-17
fix(dom): stop rewriting :checked/:disabled rules — cascade-order corruption recolors Angular Material buttons (PER-10077)

prklm10:
Fixes [PER-10077](https://browserstack.atlassian.net/browse/PER-10077) — AON customer, 70+ visual diffs.

## Root cause

`packages/dom/src/serialize-pseudo-classes.js` (interactive-states auto-detect, shipped in the 1.31.8 → 1.31.14 window via PER-7292 / #2177):

1. `markInteractiveStates` stamped every live `:disabled`/`:checked` element with `data-percy-disabled`/`data-percy-checked`.
2. `extractPseudoClassRules` copied every CSS rule whose selector contains `:disabled`/`:checked`, rewrote the pseudo to the attribute selector, and appended the copy in a `<style data-percy-interactive-states>` at the **end of `<head>`**.

The copy flips cascade ties. In the customer's app:

- sheet 1 (`styles-*.css`): `.mat-mdc-raised-button:not(:disabled) { background-color: var(--mat-button-protected-container-color, white) !important }` → resolves `#36495a` on `mat-primary` buttons
- sheet 2 (`rpcc-styles-*.css`, later): `.mat-mdc-raised-button.mat-primary { background-color: #1b98e0 !important }` → blue

Both are `!important` with equal specificity (0,2,0), so **later source order wins** → blue (matches live Chrome and the 1.31.8 baseline). Percy copied rule 1 (contains `:disabled`) to the end of head — rule 2 has no pseudo-class so it was not copied — making the copy the last equal-specificity `!important` declaration → buttons rendered `#36495a` instead of `#1b98e0`.

**Reproduced** on snapshot 2792517378 / 2792516808 of build 51680903: loading the serialized DOM in headless Chrome and deleting only the injected `<style data-percy-interactive-states>` flips the Execute buttons from `rgb(54,73,90)` (broken head screenshot) back to `rgb(27,152,224)` (baseline screenshot).

## Fix

Drop `:checked` and `:disabled` from the auto-detect stamp + rewrite entirely. Both states serialize natively, so the rewrite added no coverage:

- `disabled` is a reflected content attribute — always present in the serialized HTML (including `<fieldset disabled>` ancestors), so `:disabled` matches in the renderer as-is.
- `checked`/`selected` properties are synced to attributes on the clone by `serialize-inputs.js`, so `:checked` matches as-is.

`:focus`/`:focus-within` (genuinely not serializable) and opt-in `:hover`/`:active` are unchanged. Their rewritten copies only ever *match* explicitly stamped elements, but note the same end-of-head reordering hazard exists in principle for `:not(:focus)`-style rules — preserving cascade position for injected copies would be the complete fix and can be a follow-up.

## Testing

- `yarn workspace @percy/dom test` — 900 tests pass
- New regression tests:
  - `serialize-pseudo-classes.test.js`: 'cascade safety for :checked / :disabled rules (PER-10077)' — two-sheet equal-specificity `!important` scenario asserts no `:not(:disabled)` copy is injected; rewritePseudoSelector expectations flipped to assert `:checked`/`:disabled` pass through unchanged
  - `serialize-dom.test.js`: end-to-end asserts no `data-percy-checked`/`data-percy-disabled` stamping or rule rewriting
- Mechanics tests that used `:checked` as a vehicle (@media wrapper preservation, multi-sheet owner grouping, createElement/head fallbacks) switched to `:hover`, which still exercises the same paths

[PER-10077]: https://browserstack.atlassian.net/browse/PER-10077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


prklm10:
**Minimal repro + end-to-end verification** (also saved in percy-ops `tickets/PER-10077/repro/`):

3-file page mirroring the customer's cascade: sheet 1 has `.mat-btn:not(:disabled) { background-color: var(--btn-container-color, white) !important }`, sheet 2 (later) sets the token to `#36495a` and overrides `.mat-btn.primary { background-color: #1b98e0 !important }`. Equal specificity, both `!important` → later sheet wins → button is blue in any real browser. A `disabled` button is also on the page (required — the serializer only copies `:disabled` rules when a `:disabled` element exists).

**Local A/B** (serialize with a given @percy/dom bundle, reload serialized HTML with page JS disabled, measure computed style):

| bundle | live button | serialized button |
|---|---|---|
| @percy/dom 1.32.4 (published) | rgb(27,152,224) | **rgb(54,73,90) — bug** |
| this PR's build | rgb(27,152,224) | rgb(27,152,224) — correct |

**Real Percy builds** — same page, same CLI, only the dom bundle swapped:
- broken (master bundle): https://percy.io/fac6cb3e/web/pradum-bug-fixes-1-497164e6/builds/51881131 → Execute button renders dark
- fixed (this PR's bundle): https://percy.io/fac6cb3e/web/pradum-bug-fixes-1-497164e6/builds/51881134 → Execute button renders blue, matching the live browser

Shivanshu-07 (approved):

aryanku-dev (commented):
Claude Code Review (automated) — 3 inline finding(s). Full report in the PR comment below. Verdict: Passed.

packages/dom/src/serialize-pseudo-classes.js
```diff
@@ -484,20 +461,36 @@ function extractPseudoClassRules(ctx) {
...
+// Insert an interactive-states <style> immediately AFTER its source sheet's
+// clone element so the copy keeps the sheet's original source-order rank —
+// later stylesheets still win equal-specificity ties, exactly as on the live
+// page (PER-10077). The sheet's ownerNode (<style> / <link rel=stylesheet>) is
+// stamped with a data-percy-element-id, which serialize-cssom preserves on any
+// node it rebuilds in place. Falls back to appending at the scope's end when
+// the sheet has no locatable clone anchor.
+function injectAtSheetPosition(scopeRoot, sheet, styleElement, fallbackTarget) {
+  const anchorId = sheet.ownerNode.getAttribute('data-percy-element-id');
+  const anchor = anchorId ? scopeRoot.querySelector(`[data-percy-element-id="${anchorId}"]`) : null;
```
aryanku-dev, line 483: **[Medium] Document-level anchor lookup re-scans the whole clone per sheet**

For shadow-scoped sheets the code builds `cloneByPercyId` once (a single `querySelectorAll` pass) and does O(1) `Map.get` lookups. For document-level sheets (`owner === null`), this line instead runs a fresh full-tree `querySelector` once per sheet that produced a rewritten rule — O(sheets × DOM size) on a CSS-heavy page, where the pre-PR append-to-head was O(1) regardless of sheet count.

**Suggestion:** Build `cloneByPercyId` unconditionally (drop the `owner !== null` gate above) and pass it into `injectAtSheetPosition` for both branches instead of doing a per-sheet `querySelector`.

packages/dom/src/serialize-pseudo-classes.js
```diff
@@ -484,20 +461,36 @@ function extractPseudoClassRules(ctx) {
...
 
+// Insert an interactive-states <style> immediately AFTER its source sheet's
+// clone element so the copy keeps the sheet's original source-order rank —
+// later stylesheets still win equal-specificity ties, exactly as on the live
+// page (PER-10077). The sheet's ownerNode (<style> / <link rel=stylesheet>) is
+// stamped with a data-percy-element-id, which serialize-cssom preserves on any
+// node it rebuilds in place. Falls back to appending at the scope's end when
+// the sheet has no locatable clone anchor.
+function injectAtSheetPosition(scopeRoot, sheet, styleElement, fallbackTarget) {
+  const anchorId = sheet.ownerNode.getAttribute('data-percy-element-id');
```
aryanku-dev, line 482: **[Medium] Unguarded `sheet.ownerNode` dereference in an uncaught region**

Every entry reachable through `document.styleSheets`/`shadowRoot.styleSheets` — the only source feeding `perSheet` — has a non-null `ownerNode` today, so this isn't reachable in the tested matrix. But it's an unguarded access to browser-CSSOM state that isn't this code's own invariant, it has no test coverage, and unlike the earlier `sheet.cssRules` access (try/catch with a documented cross-origin fallback) this injection loop has no try/catch at all. A throw propagates out of `serializePseudoClasses` uncaught — the try/finally in `serializeDOM` only wraps `_liveMutations` cleanup — aborting the whole snapshot, on a function that runs on every snapshot.

**Suggestion:** Use optional chaining plus a short comment recording why it's believed unreachable today:

```suggestion
  const anchorId = sheet.ownerNode?.getAttribute('data-percy-element-id');
```

packages/dom/src/prepare-dom.js
```diff
@@ -10,8 +10,16 @@ export function markElement(domElement, disableShadowDOM, forceShadowAsLightDOM)
   // Mark elements that are to be serialized later with a data attribute.
   // Custom elements with ElementInternals or closed shadow roots also get
   // stamped so the post-clone state-fallback can locate their clones.
+  let tagName = domElement.tagName?.toLowerCase();
+  // Stylesheet <link>s are stamped so the pseudo-class serializer can locate
+  // the sheet's clone element and anchor its rewritten interactive-state rules
+  // immediately after it — preserving the sheet's original cascade position
+  // instead of appending at the end of <head> (PER-10077).
+  let isStylesheetLink = tagName === 'link' &&
+    /(^|\s)stylesheet(\s|$)/i.test(domElement.getAttribute('rel') || '');
```
aryanku-dev, line 19: **[Low] Hand-rolled `rel` token matching instead of `relList`**

JS `\s` also matches `\v` and Unicode whitespace, a superset of HTML's "ASCII whitespace" — a subtle mismatch with the platform's own DOMTokenList semantics. Not a bug in practice; style nit only.

**Suggestion:** `domElement.relList?.contains('stylesheet')` is simpler and matches the platform exactly.

aryanku-dev:
# Claude Code PR Review

**PR:** https://github.com/percy/cli/pull/2342  •  **Head:** cec61e1  •  **Reviewers:** stack-code-reviewer

## Summary

Fixes PER-10077: the interactive-states serializer copied every CSS rule containing `:disabled`/`:checked`, rewrote the pseudo to an attribute selector, and appended the copy at the **end of `<head>`** — flipping equal-specificity `!important` cascade ties and recoloring Angular Material buttons page-wide. The fix drops `:checked`/`:disabled` from the auto-detect stamp+rewrite entirely (both serialize natively), and anchors each remaining rewritten copy immediately after its source sheet's clone element via a new `injectAtSheetPosition` helper, with stylesheet `<link>`s now stamped with `data-percy-element-id` so they can be located.

## Review Table

| High | Security | No hardcoded secrets or credentials | Pass | No secrets in diff. |
| High | Security | Authentication/authorization checks present | N/A | DOM serialization library; no auth surface. |
| High | Security | Input validation and sanitization | Pass | The `[data-percy-element-id="${anchorId}"]` interpolation in `injectAtSheetPosition` was verified: `anchorId` is always produced by the codebase's own `uid()` (`_` + base-36, alphanumeric only), never derived from page content. No selector-injection risk. Matches the existing pattern in `serialize-cssom.js:72`. |
| High | Security | No IDOR — resource ownership validated | N/A | No resource access in this diff. |
| High | Correctness | Logic is correct, handles edge cases | Pass | Cascade-anchoring logic verified: shadow scoping is correct (`scopeRoot` is the host's `shadowRoot`, so lookups can't cross into sibling/nested shadow trees); the fallback path preserves the pre-PR append target for both doc-level and shadow-level scopes. Pipeline ordering verified — `serializeCSSOM` runs before `serializePseudoClasses` and preserves `data-percy-element-id` across its rebuild-in-place, so anchors still resolve. The `:checked`/`:disabled` removal was independently verified against `serialize-inputs.js` and `clone-dom.js` and loses no real coverage (see Findings note). |
| High | Correctness | Error handling is explicit, no swallowed exceptions | Fail | `injectAtSheetPosition` dereferences `sheet.ownerNode` with no null-guard and, unlike the earlier `sheet.cssRules` access, sits in a loop with no surrounding try/catch — see Finding 2. Not reachable in the current browser matrix, so this row does not gate (severity Medium, not High). |
| High | Correctness | No race conditions or concurrency issues | Pass | Synchronous single-pass serialization; no async work added. |
| Medium | Testing | New code has corresponding tests | Fail | The `<link>`-stamping → `<link>`-anchor path — the literal new mechanism this PR adds — is tested only for "is stamped", never for "rule lands in the right place". See Finding 3. |
| Medium | Testing | Error paths and edge cases tested | Pass | The no-anchor fallback path has a dedicated test (`falls back to end of head when a stamped sheet has no clone anchor`). The `sheet.ownerNode` branch in Finding 2 is untested but also currently unreachable. |
| Medium | Testing | Existing tests still pass (no regressions) | Pass | All GitHub CI checks green at `cec61e1` (Lint, Build, Regression, CodeQL, Semgrep, and the full dual-browser `Test @percy/*` matrix). Reviewer independently ran the `@percy/dom` Karma suite on ChromeHeadless: **495/495 passing**. |
| Medium | Performance | No N+1 queries or unbounded data fetching | Fail | Document-level sheets now do one full-tree `querySelector` each instead of reusing the `cloneByPercyId` map already built for the shadow case — O(sheets × DOM size) where the pre-PR append was O(1). See Finding 1. |
| Medium | Performance | Long-running tasks use background jobs | N/A | Synchronous serialization path by design. |
| Medium | Quality | Follows existing codebase patterns | Pass | The new `<link>` stamp reuses the established `data-percy-element-id` convention already applied to `<style>`, `<input>`, and custom elements — additive, and verified not to interact with `serializeCSSOM` (external `href` sheets fail `isCSSOM()` so are never rebuilt). |
| Medium | Quality | Changes are focused (single concern) | Pass | Two tightly-coupled halves of one root cause (stop rewriting `:checked`/`:disabled`; anchor what remains at source-sheet position), plus one small unrelated test-comment tidy in `packages/core/test/utils.test.js`. |
| Low | Quality | Meaningful names, no dead code | Pass | Removing the `ctx._stampedInteractive` short-circuit deletes genuine dead code — it only ever filtered `:checked`/`:disabled`, which no longer exist in the pseudo list. Not a perf regression. |
| Low | Quality | Comments explain why, not what | Pass | Comments are unusually strong here — the cascade-tie rationale and the "serializes natively" justification are both documented at the point of change with the ticket reference. |

## Findings

**1.**
- **File:** `packages/dom/src/serialize-pseudo-classes.js:483`
- **Severity:** Medium
- **Issue:** For shadow-scoped sheets the code builds `cloneByPercyId` once (a single `querySelectorAll` pass) and does O(1) `Map.get` lookups. For document-level sheets (`owner === null`), `injectAtSheetPosition` instead runs a fresh full-tree `scopeRoot.querySelector('[data-percy-element-id="…"]')` once per sheet that produced a rewritten rule. On a page with many `<link>`/`<style>` sheets each carrying at least one `:hover`/`:focus`/`:active` rule this is O(sheets × DOM size), where the pre-PR append-to-head was O(1) regardless of sheet count.
- **Suggestion:** Build `cloneByPercyId` unconditionally (drop the `owner !== null` gate at line 445) and pass it into `injectAtSheetPosition` for both branches instead of doing a per-sheet `querySelector`.

**2.**
- **File:** `packages/dom/src/serialize-pseudo-classes.js:482`
- **Severity:** Medium
- **Issue:** `sheet.ownerNode.getAttribute(...)` is an unguarded dereference. In current mainstream browsers every entry reachable through `document.styleSheets`/`shadowRoot.styleSheets` — the only source feeding `perSheet` via `collectStyleSheets` — has a non-null `ownerNode` (adopted/constructed stylesheets aren't exposed there and are handled separately in `serialize-cssom.js`), so this is not reachable in the tested matrix. But it is an unguarded access to browser-CSSOM state that isn't the code's own invariant, it has zero test coverage, and — unlike the earlier `sheet.cssRules` access, which is wrapped in try/catch with a documented cross-origin fallback — this injection loop has no try/catch at all. A throw here propagates out of `serializePseudoClasses` uncaught (the try/finally in `serializeDOM` only wraps `_liveMutations` cleanup), aborting the entire snapshot. This function runs unconditionally on every snapshot.
- **Suggestion:** `const anchorId = sheet.ownerNode?.getAttribute('data-percy-element-id');` plus a one-line comment recording why it is believed unreachable today.

**3.**
- **File:** `packages/dom/test/serialize-dom.test.js` (new `stylesheet <link> stamping` describe block) and `packages/dom/test/serialize-pseudo-classes.test.js` (`cascade position of injected rules`)
- **Severity:** Medium
- **Issue:** The `prepare-dom.js` stamping logic exists *only* to support anchoring against `<link>` clones, but no test closes that loop. The new `stamps only stylesheet <link>s with a percy element id` test asserts the attribute is present and never asserts a copy is anchored after that `<link>`. Both cascade-position tests use inline `<style>` elements — and the `serialize-pseudo-classes.test.js` pair even hand-fabricates `data-percy-element-id` on the source `<style>` tags, bypassing the real clone pipeline entirely. So the PR's core new mechanism is covered for "is stamped" but never for "rule lands in the right place."
- **Suggestion:** Add one test serving a real same-origin `.css` asset (like the existing `test/assets/*` fixtures) via `<link rel="stylesheet">` with a `:hover`/`:focus` rule, asserting the injected `<style data-percy-interactive-states>` sits immediately after the `<link>` clone and before a later sheet. Note a `data:text/css,…` href will not work — `sheet.cssRules` on a `data:`-sourced link is liable to hit the same opaque-origin restriction the code's own try/catch guards against, which is likely why the existing test only checks the stamp.

**4.**
- **File:** `packages/dom/src/prepare-dom.js:19`
- **Severity:** Low
- **Issue:** `/(^|\s)stylesheet(\s|$)/i.test(domElement.getAttribute('rel') || '')` hand-rolls `rel` token matching. JS `\s` also matches `\v` and Unicode whitespace, a superset of HTML's "ASCII whitespace" — a subtle semantic mismatch with the platform, though not a bug in practice.
- **Suggestion:** `domElement.relList?.contains('stylesheet')` — simpler, and exactly the DOMTokenList semantics the platform uses. Style nit only.

**5.**
- **File:** `.nycrc` / `packages/dom/src/serialize-pseudo-classes.js` (out of diff)
- **Severity:** Informational — not attributable to this PR
- **Issue:** A local Chrome-only coverage run showed `serialize-pseudo-classes.js` at 98.63% lines / 99.19% branches, against a repo-root `.nycrc` requiring 100% with no `per-file` override. The shortfall traces entirely to `markOpenPopovers`'s pre-existing `:popover-open`-unsupported catch branch, which this diff only touches in comments and which most likely only fires under Firefox in the real dual-browser CI job (no Firefox binary was available in the review sandbox to merge that coverage in).
- **Suggestion:** No action needed — GitHub CI is green at `cec61e1`, including both `Test @percy/dom` jobs, so the real dual-browser merge does satisfy the threshold. Recorded only because the reviewer could not reproduce the full-matrix figure locally.

---

**Verdict: PASS** — no High or Critical findings; the cascade fix and the `:checked`/`:disabled` removal both held up under source-level verification, with three Medium items (perf, a latent unguarded dereference, and a test gap on the new `<link>`-anchoring path) worth addressing before merge.

PR #2343, opened 2026-07-17
Fix: Percy CLI hangs on Node.js 26 during Chromium extraction (PER-10062)

prklm10:
Fixes PER-10062

## Root cause
`@percy/core` extracts the downloaded Chromium archive with `extract-zip@2.0.1` (`packages/core/src/install.js:149`), which depends on `yauzl@2`. yauzl ≤3.3.0 relies on undefined behavior of Node stream `destroy()`; Node.js 26 tightened this, so `zipfile.openReadStream()` callbacks never fire on deflate entries and extraction hangs forever (https://github.com/thejoshwolfe/yauzl/issues/176, fixed in yauzl 3.3.1). `extract-zip` is unmaintained and still pins yauzl 2, so every consumer hangs on Node 26 — Playwright hit the same bug and moved to yauzl 3 directly.

Because `percy exec` launches the discovery browser (and therefore installs Chromium) **before** spawning the wrapped command, the hang means customer test suites never start and builds finalize with `no_snapshots`. Setting `PERCY_BROWSER_EXECUTABLE` (skipping install) is the customer-reported workaround, which corroborates the diagnosis.

Reproduced locally: `install.chromium()` hangs mid-extraction on Node 26.5.0 and completes on Node 22.20.0; `extract-zip` alone hangs on the same archive under Node 26.

## Fix
Replaces `extract-zip` with a local `packages/core/src/unzip.js` built on `yauzl@^3.4.0` (engines node >=12, single dependency `pend`), preserving previous behavior: file modes (default 755/644), directory entries, symlinks, `__MACOSX/` skip, zip-slip guard, and the absolute-target-directory requirement.

Verified extraction of `chrome-mac.zip` (152 MB, deflate entries + symlinks) is byte-, mode-, and symlink-identical to extract-zip's output, completes on Node 14.21.3 / 22.20.0 / 26.5.0, and the extracted Chromium launches (`--version` → Chromium 143.0.7499.0).

## Testing
- New `packages/core/test/unit/unzip.test.js`: 7 specs against hand-crafted zip fixtures — stored + deflated entries, nested dirs, `__MACOSX/` skip, zip-slip rejection, relative-dir rejection, executable mode preservation, symlink restoration.
- `packages/core/test/unit/install.test.js` mocks updated to stub the new module; all 48 specs (install + unzip) pass under Node 14 (CI test env).

### DIFF
--- a/packages/core/src/unzip.js
+++ b/packages/core/src/unzip.js
@@ -59,9 +59,14 @@ export default async function unzip(archive, { dir }) {
             return zipfile.readEntry();
           }
 
-          let dest = path.join(dir, entry.fileName);
+          // strip any leading traversal segments before joining (zip-slip)
+          let fileName = entry.fileName.replace(/^(\.\.(\/|\\|$))+/, '');
+          // `dir` is the trusted install target from install.js, not user input,
+          // and `fileName` is sanitized above with the traversal guard below
+          // nosemgrep: javascript.lang.security.audit.path-traversal.path-join-resolve-traversal.path-join-resolve-traversal
+          let dest = path.join(dir, fileName);
 
-          // guard against zip-slip
+          // guard against zip-slip for any remaining traversal
           if (path.relative(dir, dest).split(path.sep).includes('..')) {
             throw new Error(`Out of bound path "${dest}" found while processing file ${entry.fileName}`);
           }
--- a/packages/core/src/unzip.js
+++ b/packages/core/src/unzip.js
@@ -1,0 +1,100 @@
+import fs from 'fs';
+import path from 'path';
+import stream from 'stream';
+import { promisify } from 'util';
+import yauzl from 'yauzl';
+
+const pipeline = promisify(stream.pipeline);
+const openZip = promisify(yauzl.open);
+
+// stat mode constants for entry external file attributes
+const IFMT = 0o170000;
+const IFDIR = 0o040000;
+const IFLNK = 0o120000;
+
+// Returns the file mode of an extracted entry, with defaults matching the
+// previous extract-zip behavior when the archive does not provide one
+function extractedMode(mode, isDir) {
+  return (mode || (isDir ? 0o755 : 0o644)) & 0o777;
+}
+
+// Collects a readable stream into a utf8 string (symlink targets)
+async function readAll(readable) {
+  let chunks = [];
+  for await (let chunk of readable) chunks.push(chunk);
+  return Buffer.concat(chunks).toString('utf8');
+}
+
+// Extracts a zip archive into a directory, preserving file modes, directory
+// entries, and symlinks. Replaces the unmaintained extract-zip package, whose
+// yauzl@2 dependency deadlocks on deflate entries under Node 26
+// (https://github.com/thejoshwolfe/yauzl/issues/176).
+export default async function unzip(archive, { dir }) {
+  if (!path.isAbsolute(dir)) {
+    throw new Error('Target directory is expected to be absolute');
+  }
+
+  await fs.promises.mkdir(dir, { recursive: true });
+  dir = await fs.promises.realpath(dir);
+
+  let zipfile = await openZip(archive, { lazyEntries: true, autoClose: false });
+  let openReadStream = promisify(zipfile.openReadStream.bind(zipfile));
+
+  try {
+    await new Promise((resolve, reject) => {
+      let done = false;
+      let finish = err => {
+        if (done) return;
+        done = true;
+        if (err) reject(err); else resolve();
+      };
+
+      zipfile.on('error', finish);
+      zipfile.on('end', () => finish());
+
+      zipfile.on('entry', async entry => {
+        try {
+          // skip macOS resource fork entries
+          if (entry.fileName.startsWith('__MACOSX/')) {
+            return zipfile.readEntry();
+          }
+
+          let dest = path.join(dir, entry.fileName);
+
+          // guard against zip-slip
+          if (path.relative(dir, dest).split(path.sep).includes('..')) {
+            throw new Error(`Out of bound path "${dest}" found while processing file ${entry.fileName}`);
+          }
+
+          let mode = (entry.externalFileAttributes >> 16) & 0xFFFF;
+          let isSymlink = (mode & IFMT) === IFLNK;
+          let isDir = (mode & IFMT) === IFDIR ||
+            entry.fileName.endsWith('/') ||
+            // directories from some windows archivers lack mode bits
+            ((entry.versionMadeBy >> 8) === 0 && entry.externalFileAttributes === 16);
+
+          if (isDir) {
+            await fs.promises.mkdir(dest, { recursive: true, mode: extractedMode(mode, true) });
+          } else {
+            await fs.promises.mkdir(path.dirname(dest), { recursive: true });
+            let contents = await openReadStream(entry);
+
+            if (isSymlink) {
+              await fs.promises.symlink(await readAll(contents), dest);
+            } else {
+              await pipeline(contents, fs.createWriteStream(dest, { mode: extractedMode(mode, false) }));
+            }
+          }
+
+          zipfile.readEntry();
+        } catch (error) {
+          finish(error);
+        }
+      });
+
+      zipfile.readEntry();
+    });
+  } finally {
+    zipfile.close();
+  }
+}
--- a/packages/core/src/unzip.js
+++ b/packages/core/src/unzip.js
@@ -44,6 +44,8 @@ export default async function unzip(archive, { dir }) {
     await new Promise((resolve, reject) => {
       let done = false;
       let finish = err => {
+        /* istanbul ignore if: defensive guard for stray zipfile events
+           after the promise has already settled */
         if (done) return;
         done = true;
         if (err) reject(err); else resolve();
@@ -67,6 +69,8 @@ export default async function unzip(archive, { dir }) {
           let dest = path.join(dir, fileName);
 
           // guard against zip-slip for any remaining traversal
+          /* istanbul ignore if: yauzl rejects traversal entry names at parse
+             time (validateFileName); kept as defense-in-depth */
           if (path.relative(dir, dest).split(path.sep).includes('..')) {
             throw new Error(`Out of bound path "${dest}" found while processing file ${entry.fileName}`);
           }
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -83,6 +83,9 @@ const METADATA_HOSTNAMES = new Set([
 // ::ffff:169.254.169.254 compares equal to the literal 169.254.169.254 — the
 // OS routes it to the IPv4 service, so it must not slip past the IP set).
 function canonicalHost(host) {
+  /* istanbul ignore next: every caller guards against falsy hosts
+     (matchMetadataHost early-returns; METADATA_IPS maps literals) —
+     kept as a defensive no-op for direct use */
   if (!host) return host;
   let h = String(host).toLowerCase().replace(/\.$/, '');
   let bare = h.replace(/^\[/, '').replace(/\]$/, '');
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -55,7 +55,6 @@
     "busboy": "^1.6.0",
     "content-disposition": "^0.5.4",
     "cross-spawn": "^7.0.3",
-    "extract-zip": "^2.0.1",
     "fast-glob": "^3.2.11",
     "fast-xml-parser": "^4.4.1",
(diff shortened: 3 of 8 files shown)


packages/core/src/unzip.js
```diff
@@ -0,0 +1,109 @@
...
+            ((entry.versionMadeBy >> 8) === 0 && entry.externalFileAttributes === 16);
+
+          if (isDir) {
+            await fs.promises.mkdir(dest, { recursive: true, mode: extractedMode(mode, true) });
+          } else {
+            await fs.promises.mkdir(path.dirname(dest), { recursive: true });
+            let contents = await openReadStream(entry);
+
+            if (isSymlink) {
+              await fs.promises.symlink(await readAll(contents), dest);
```
prklm10, line 92: **[Medium] Symlink target is not validated for containment**

The zip-slip guard only checks where the symlink is *created*, not where it *points* — an archive can plant a link inside `dir` targeting an absolute path (`/etc/passwd`) or a relative escape (`../../..`). Anything downstream that follows the link in the extracted tree can be led outside the install dir. This is parity with the old extract-zip, not a regression, but worth closing while rewriting this module.

**Suggestion:** resolve the target against `path.dirname(dest)` and reject when `path.relative(dir, resolved)` contains `..`, mirroring the existing zip-slip check:

```js
let linkTarget = await readAll(contents);
let resolved = path.resolve(path.dirname(dest), linkTarget);
if (path.relative(dir, resolved).split(path.sep).includes('..')) {
  throw new Error(`Out of bound symlink target "${linkTarget}" in ${entry.fileName}`);
}
await fs.promises.symlink(linkTarget, dest);
```

_Reviewer: stack:code-reviewer_

packages/core/src/unzip.js
```diff
@@ -0,0 +1,109 @@
...
+            // directories from some windows archivers lack mode bits
+            ((entry.versionMadeBy >> 8) === 0 && entry.externalFileAttributes === 16);
+
+          if (isDir) {
+            await fs.promises.mkdir(dest, { recursive: true, mode: extractedMode(mode, true) });
+          } else {
+            await fs.promises.mkdir(path.dirname(dest), { recursive: true });
+            let contents = await openReadStream(entry);
+
+            if (isSymlink) {
```
prklm10, line 91: **[Medium] Retry over a partial extraction throws `EEXIST` on symlinks**

`fs.createWriteStream` overwrites existing files and `mkdir({ recursive: true })` tolerates existing dirs, but `fs.promises.symlink` throws `EEXIST` if `dest` already exists. If a previous install was interrupted mid-extraction (crash, SIGKILL, disk full), the retry into the same `outdir` — which `download()` reuses unless the final executable exists — fails hard on the first symlink entry.

**Suggestion:** `await fs.promises.rm(dest, { force: true })` before the `symlink` call to make retries idempotent.

_Reviewer: stack:code-reviewer_

packages/core/src/unzip.js
prklm10, line 92: **[Low] Windows symlink `type` not specified**

Omitting the `type` argument defaults to `'file'` on Windows, which is wrong for directory symlinks (Chromium archives can contain them, e.g. framework links). The test suite skips symlink assertions on win32, so this path is unverified rather than proven broken.

**Suggestion:** pass an explicit `type` derived from the entry, or add a comment documenting that Windows symlink extraction is not verified.

_Reviewer: stack:code-reviewer_

packages/core/src/unzip.js
```diff
@@ -0,0 +1,109 @@
...
+export default async function unzip(archive, { dir }) {
+  if (!path.isAbsolute(dir)) {
+    throw new Error('Target directory is expected to be absolute');
+  }
+
+  await fs.promises.mkdir(dir, { recursive: true });
+  dir = await fs.promises.realpath(dir);
+
+  let zipfile = await openZip(archive, { lazyEntries: true, autoClose: false });
+  let openReadStream = promisify(zipfile.openReadStream.bind(zipfile));
```
prklm10, line 41: **[Low] yauzl@3 ships a native promise/async-iterator API**

`yauzl.openPromise`, `zipfile.openReadStreamPromise`, and `for await (const entry of zipfile.eachEntry())` exist in yauzl@3.4.0. The hand-rolled `done`-flag event state machine (plus its istanbul-ignore defensive guards) is exactly the class of event/stream lifecycle code that caused the original hang.

**Suggestion:** non-blocking follow-up — adopt `eachEntry()` to drop ~30 lines of custom event plumbing.

_Reviewer: stack:code-reviewer_

packages/core/src/unzip.js
```diff
@@ -0,0 +1,109 @@
...
+
+// stat mode constants for entry external file attributes
+const IFMT = 0o170000;
+const IFDIR = 0o040000;
+const IFLNK = 0o120000;
+
+// Returns the file mode of an extracted entry, with defaults matching the
+// previous extract-zip behavior when the archive does not provide one
+function extractedMode(mode, isDir) {
+  return (mode || (isDir ? 0o755 : 0o644)) & 0o777;
```
prklm10, line 18: **[Low] Explicit mode `0` silently becomes the default `644`/`755`**

The falsy check means an entry that intentionally sets permission bits to `0` gets the defaults instead. This mirrors extract-zip's behavior, so it's intentional parity — but a future reader may "fix" it and change behavior.

**Suggestion:** add a one-line comment noting the falsy check is intentional parity with extract-zip.

_Reviewer: stack:code-reviewer_

packages/core/src/utils.js
```diff
@@ -83,6 +83,9 @@ const METADATA_HOSTNAMES = new Set([
 // ::ffff:169.254.169.254 compares equal to the literal 169.254.169.254 — the
 // OS routes it to the IPv4 service, so it must not slip past the IP set).
 function canonicalHost(host) {
+  /* istanbul ignore next: every caller guards against falsy hosts
```
prklm10, line 86: **[Low] Unrelated coverage fix bundled into the Node 26 hang PR**

The `canonicalHost` istanbul-ignore and its companion test (commits 5752528, ed70ae7) fix a master coverage gate break, not the extraction hang. Bundling them muddies the diff's intent and future git blame for both the Node 26 fix and the metadata-SSRF guard.

**Suggestion:** split into a separate commit/PR titled around the coverage fix, unless CI required them to land together on this branch.

_Reviewer: stack:code-reviewer_

prklm10 (commented):

prklm10:
# Claude Code PR Review

**PR:** https://github.com/percy/cli/pull/2343  •  **Head:** ed70ae7  •  **Reviewers:** stack:code-reviewer

## Summary
Replaces `extract-zip` (whose `yauzl@2` dependency deadlocks on deflate entries under Node.js 26) with a local `unzip` module built on `yauzl@^3.4.0`, preserving extract-zip behavior (file modes, directory entries, symlinks, `__MACOSX/` skip, zip-slip guard, absolute-target requirement) with a from-scratch zip-fixture test suite; also bundles a small `canonicalHost` coverage fix.

## Review Table

| High | Security | No hardcoded secrets or credentials | Pass | No secrets introduced; dependency swap only |
| High | Security | Authentication/authorization checks present | N/A | No auth surface in this change |
| High | Security | Input validation and sanitization | Pass | Zip-slip guarded twice (yauzl@3 `validateFileName` at parse time + explicit pre-join sanitization and traversal check, verified against yauzl source); symlink *target* containment gap noted as Medium finding #1 (parity with old extract-zip, not a regression) |
| High | Security | No IDOR — resource ownership validated | N/A | No resource-ownership surface |
| High | Security | No SQL injection (parameterized queries) | N/A | No database code |
| High | Correctness | Logic is correct, handles edge cases | Pass | 10 unzip specs run and pass locally; covers stored/deflated entries, nested dirs, modes, windows dir attrs, symlinks, corruption |
| High | Correctness | Error handling is explicit, no swallowed exceptions | Pass | Entry-handler errors funnel through `finish()` and reject the extraction promise; corrupted data rejects; `zipfile.close()` in `finally` |
| High | Correctness | No race conditions or concurrency issues | Pass | `done` flag guards double-settle; `lazyEntries` serializes entry processing |
| Medium | Testing | New code has corresponding tests | Pass | New 210-line `unzip.test.js` with hand-built zip fixtures; 100% coverage on the module |
| Medium | Testing | Error paths and edge cases tested | Pass | Traversal entries, corrupt deflate data, relative target dir all covered; Windows symlink type is an untested gap (finding #3) |
| Medium | Testing | Existing tests still pass (no regressions) | Pass | install.test.js mocks updated from `extract-zip` to `./unzip.js`; unzip suite passes locally |
| Medium | Performance | No N+1 queries or unbounded data fetching | N/A | No query surface |
| Medium | Performance | Long-running tasks use background jobs | N/A | Extraction is inherently synchronous with install; streams via `pipeline` (no full-file buffering except symlink targets) |
| Medium | Quality | Follows existing codebase patterns | Pass | ESM module style, comment density, and test conventions match packages/core |
| Medium | Quality | Changes are focused (single concern) | Fail | Unrelated `canonicalHost` istanbul-ignore/coverage changes (`packages/core/src/utils.js`, `test/utils.test.js`) bundled into a Node-26-hang fix (finding #5) |
| Low | Quality | Meaningful names, no dead code | Pass | Clear naming; removes 6 transitive deps, adds 1 |
| Low | Quality | Comments explain why, not what | Pass | Comments document parity rationale, yauzl issue link, and defense-in-depth reasoning |
| Low | Quality | No unnecessary dependencies added | Pass | Net reduction: extract-zip + 6 transitive deps removed, yauzl@3 (+pend) added |

## Findings

- **File:** `packages/core/src/unzip.js:92`
- **Severity:** Medium
- **Issue:** Symlink extraction validates where the link is *created* but not where it *points*. An archive can plant a symlink whose target is absolute (`/etc/passwd`) or escapes the install dir (`../../..`); downstream consumers of the extracted tree could follow it outside the sandboxed install dir. Parity with old extract-zip, not a regression.
- **Suggestion:** Resolve the target against `path.dirname(dest)` and reject if `path.relative(dir, resolved)` contains `..`, mirroring the existing zip-slip guard.

- **File:** `packages/core/src/unzip.js:91`
- **Severity:** Medium
- **Issue:** `fs.promises.symlink` throws `EEXIST` if `dest` already exists. A retry after an interrupted extraction into the same `outdir` (which `download()` reuses unless the final executable exists) fails hard on any symlink entry, while regular files/dirs overwrite fine.
- **Suggestion:** `await fs.promises.rm(dest, { force: true })` before creating the symlink to make retries idempotent.

- **File:** `packages/core/src/unzip.js:92`
- **Severity:** Low
- **Issue:** No `type` argument to `fs.promises.symlink` — defaults to `'file'` on Windows, wrong for directory symlinks (which Chromium archives can contain). Tests skip symlink assertions on win32, so this path is unverified.
- **Suggestion:** Pass an explicit `type` based on the entry, or document that Windows symlink extraction is not verified.

- **File:** `packages/core/src/unzip.js:41`
- **Severity:** Low
- **Issue:** yauzl@3 ships a native promise/async-iterator API (`yauzl.openPromise`, `zipfile.openReadStreamPromise`, `zipfile.eachEntry()`), but the module hand-rolls a `done`-flag event state machine over `promisify` — the exact class of event/stream lifecycle code that caused the original hang.
- **Suggestion:** Follow-up refactor to `for await (const entry of zipfile.eachEntry())`, removing ~30 lines of custom event plumbing and the istanbul-ignore defensive guards.

- **File:** `packages/core/src/utils.js:86`
- **Severity:** Low
- **Issue:** The `canonicalHost` istanbul-ignore and its coverage test (commits 5752528, ed70ae7) are unrelated to the Node 26 extraction fix — scope creep that muddies the diff and future git blame for both the hang fix and the metadata-SSRF guard.
- **Suggestion:** Split into a separate commit/PR titled around the coverage fix, unless CI required them to land together.

- **File:** `packages/core/src/unzip.js:18`
- **Severity:** Low
- **Issue:** `extractedMode` treats an explicit mode of `0` as "no mode provided" and substitutes `644`/`755`. Intentional parity with extract-zip, but a future reader may "fix" it and change behavior.
- **Suggestion:** Add a one-line comment noting the falsy check is intentional parity.

- **File:** `packages/core/test/unit/unzip.test.js` (file-level)
- **Severity:** Low
- **Issue:** No regression coverage for the original failure mode: the hang was silent, and no Node 26 CI job exercises real extraction, so a future dependency bump could quietly reintroduce it.
- **Suggestion:** Add a Node 26 job to the CI matrix (in progress on PER-9999) so extraction runs under the affected runtime.

---

**Verdict: PASS** — core fix is correct, well-tested, and dependency-reducing; findings are non-blocking hardening follow-ups (symlink target containment and retry idempotency are worth a fast follow-up PR).

PR #2353, opened 2026-07-27
fix(core): follow BS App Automate tmp dir relocation for Maestro screenshots

prklm10:
## Summary
- BrowserStack hosts relocated the App Automate session directories away from `/tmp/{sessionId}`, breaking the Maestro screenshot relay's hardcoded `/tmp` globs and scope roots — screenshot lookups started 404'ing.
- All BrowserStack-mode paths (glob patterns, manual-walker fallback, and the `scopeRoot` used by the realpath containment check) now route through a new `appAutomateTmpDir()` helper.
- The helper reads a new `PERCY_APP_AUTOMATE_TMP_DIR` env var — injected by the host with the relocated root — and falls back to the legacy `/tmp` convention when unset, so no internal infra path is hardcoded in the CLI and hosts with older injection keep working during rollout. Trailing separators are trimmed, backslashes normalized for the glob, and non-absolute values fall back to `/tmp`.
- Self-hosted mode is untouched — it continues to scope via `PERCY_MAESTRO_SCREENSHOT_DIR`.

## Test plan
- Existing `/percy/maestro-screenshot` specs (Android + iOS globs, `filePath` acceptance/containment, PNG-fill, regions) — all passing against the default root.
- New `PERCY_APP_AUTOMATE_TMP_DIR override` specs: globbing under an overridden root (real-fs fixture), trailing-slash tolerance, non-absolute fallback to `/tmp`, direct trim/fallback assertions on the exported helper, and `filePath` containment re-scoping to the overridden root.
- Full `packages/core/test/api.test.js` run locally: all maestro specs pass; the single unrelated failure (`when the server is disabled…`, ECONNREFUSED vs AggregateError) also fails on a clean master checkout — pre-existing local Node environment issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2357, opened 2026-07-30
fix(core): let BS hosts inject a complete Maestro scope root

rishigupta1599:
## Summary

Follow-up to #2353. That PR assumed the App Automate tmp-dir relocation only moved the root while preserving the `{root}/{sessionId}{_test_suite}` layout beneath it. That holds on Android but not on iOS.

| | Relocated path | What the CLI composes | `PERCY_APP_AUTOMATE_TMP_DIR` enough? |
|---|---|---|---|
| Android | `<aa-tmp>/<device>/test-suites/<sid>_test_suite/logs/*/screenshots/` | `{TMP}/{sid}_test_suite/logs/*/screenshots/` | ✅ |
| iOS | `<aa-tmp>/<device>/logs/maestro_debug_*/` | `{TMP}/{sid}/*_maestro_debug_*/` | ❌ |

The iOS layout drops the `sessionId` path segment entirely — the CLI appends it, so no tmp-root value can absorb `<device>/logs` — and drops the `<device>_` prefix that the `*_maestro_debug_*` glob keys on. Lookups there 404 no matter what `PERCY_APP_AUTOMATE_TMP_DIR` is set to.

## What this adds

`PERCY_MAESTRO_BS_SCOPE_ROOT`, for layouts the convention can no longer express. When set, the value **is** the scope root:

- the relay globs `{root}/**/{name}.png` with no layout assumption
- the manual-walker fallback recurses the same root instead of the platform convention
- the realpath containment check anchors on it

The security boundary is relocated, never widened — an out-of-root `filePath` still 404s. Non-absolute values are ignored so a malformed injection falls back to the composed convention rather than producing a cwd-relative root. Android (`PERCY_APP_AUTOMATE_TMP_DIR`) and self-hosted (`PERCY_MAESTRO_SCREENSHOT_DIR`) are untouched.

No existence pre-check on the override, unlike self-hosted: this is host config, not customer config, so a stale root should surface as the same 404 the containment check emits rather than a 400 aimed at the customer.

## Host side

- `browserstack/realmobile` — injects `PERCY_MAESTRO_BS_SCOPE_ROOT` (needs this PR released)
- `browserstack/mobile` — injects `PERCY_APP_AUTOMATE_TMP_DIR`, fixed by #2353 alone

Both host PRs pin 1.32.5 today; realmobile needs a second bump once this ships.

## Test plan

8 new specs in `packages/core/test/api.test.js`: locating a screenshot the platform convention can't reach, android under the same override, override winning over `PERCY_APP_AUTOMATE_TMP_DIR`, trailing-slash tolerance, non-absolute fallback, `filePath` containment re-anchored on the override (in-root resolves, out-of-root 404s), missing root 404, plus direct trim/null assertions on the exported helper.

**Local run caveat, stated plainly:** every real-fs-fixture spec in `api.test.js` fails on my machine with `ENOTEMPTY` from the `afterEach` `rmSync`, including blocks this PR doesn't touch. Measured baseline on clean master: **15 failures, all `ENOTEMPTY`**. With this PR: **23** — a delta of exactly the 8 new specs, same environmental cause, no new failure mode. Those blocks need CI to actually be exercised.

Because of that, I verified the logic directly against the real filesystem using the exact relocated layouts — 7/7: helper trim/absolute semantics, #2353-alone still 404ing on the realmobile layout, explicit root locating the file, out-of-root `filePath` rejected, in-root accepted; plus a separate check that the Android glob still composes correctly from `PERCY_APP_AUTOMATE_TMP_DIR`.

ESLint clean on all three changed files. (`yarn lint` for the whole package fails environmentally here — 54 identical "No Babel config file detected" errors across every test file, untouched ones included.)

Ref: PER-10040, AAP-18965

### DIFF
--- a/packages/core/src/maestro-screenshot-file.js
+++ b/packages/core/src/maestro-screenshot-file.js
@@ -15,16 +15,10 @@ export function appAutomateTmpDir() {
   return path.isAbsolute(dir) ? dir : '/tmp';
 }
 
-// Complete session scope root for BS hosts whose layout the
-// {appAutomateTmpDir()}/{sessionId}{_test_suite} convention can't express:
-// realmobile's AAP-18965 iOS relocation drops both the sessionId segment and
-// the <device>_ prefix the glob keys on. Android kept its shape and stays on
-// PERCY_APP_AUTOMATE_TMP_DIR.
-//
-// When set this IS the scope root — the relay globs {root}/**/{name}.png and
-// the realpath containment check anchors on it, so the boundary is relocated,
-// never widened. Non-absolute values are ignored rather than resolved against
-// cwd, which also rejects '/' once the trailing separator is trimmed.
+// Complete scope root for hosts whose layout {appAutomateTmpDir()}/{sessionId}
+// can't express (realmobile's AAP-18965 iOS move). When set this IS the root:
+// globbed recursively, and the realpath containment check anchors on it — so
+// the boundary moves, never widens. Non-absolute is ignored, which drops '/'.
 export function bsScopeRootOverride() {
   let raw = process.env.PERCY_MAESTRO_BS_SCOPE_ROOT;
   if (!raw) return null;
@@ -40,9 +34,8 @@ export function bsScopeRootOverride() {
    rationale (upstream SAFE_ID validation, depth cap, exact filename match). */
 async function manualScreenshotWalk(platform, sessionId, name, scopeRoot) {
   const files = [];
-  // `accept` gates which matches count: the iOS convention keeps its
-  // `_maestro_debug_` guard, while an explicit scope root (already narrowed to
-  // one session by the host) takes anything.
+  // `accept` gates matches: iOS keeps its `_maestro_debug_` guard, an explicit
+  // root (already narrowed to one session) takes anything.
   const walkFrom = async (root, accept) => {
     const walk = async (dir, depth) => {
       if (depth > 15) return; // sanity cap
@@ -84,17 +77,14 @@ async function manualScreenshotWalk(platform, sessionId, name, scopeRoot) {
 //   1. `filePath` supplied (BrowserStack new SDK — absolute path under the BS
 //      session root; rejected upstream in self-hosted mode).
 //   2. BrowserStack glob (the BS-infra SCREENSHOTS_DIR layout).
-//   3. Recursive glob under scopeRoot — self-hosted
-//      (PERCY_MAESTRO_SCREENSHOT_DIR) or a BS explicit root
-//      (PERCY_MAESTRO_BS_SCOPE_ROOT); `recursiveScope` selects it.
+//   3. Recursive glob under scopeRoot — self-hosted, or a BS explicit root.
 // Either way, the shared realpath + scopeRoot prefix check below enforces the
 // security invariant. Returns the canonicalized absolute path, or throws
 // ServerError(404) when the file is missing or resolves outside scopeRoot.
 // Callers pass `filePath` already shape-validated, plus the resolved `scopeRoot`
 // and `selfHosted` flag.
 export async function locateScreenshot({ platform, sessionId, name, filePath, scopeRoot, selfHosted, recursiveScope }) {
-  // Derived here rather than trusted from the caller so `selfHosted` alone
-  // keeps its existing meaning.
+  // Derived, not trusted from the caller, so `selfHosted` keeps its meaning.
   let scopedGlob = selfHosted || !!recursiveScope;
   let chosenFile;
   if (filePath) {
@@ -109,8 +99,8 @@ export async function locateScreenshot({ platform, sessionId, name, filePath, sc
     //   Self-hosted: recursive glob under the customer's --test-output-dir
     //     (scopeRoot = PERCY_MAESTRO_SCREENSHOT_DIR). `name` is SAFE_ID-validated
     //     by the caller, so it cannot contain separators or traversal chars.
-    //   BS explicit scope root: same recursive glob — the host already narrowed
-    //     scopeRoot to this session, so there is no convention left to key on.
+    //   BS explicit root: same recursive glob — the host already narrowed
+    //     scopeRoot to this session, so no convention is left to key on.
     let searchPattern;
     if (scopedGlob) {
       // fast-glob requires forward-slashes in patterns on every platform; on
@@ -139,8 +129,7 @@ export async function locateScreenshot({ platform, sessionId, name, filePath, sc
       // Fast-glob import / glob call failed — fall back to manual walker (BS
       // only; self-hosted has no fixed-layout convention, so empty → 404 with
       // the actionable PERCY_MAESTRO_SCREENSHOT_DIR guidance from the caller).
-      // The walker mirrors the glob: an explicit scope root recurses that root,
-      // otherwise it follows the platform convention.
+      // The walker mirrors the glob: explicit root recurses it, else convention.
       // See manualScreenshotWalk() at file top + the file-level .semgrepignore.
       /* istanbul ignore next — only fires when fast-glob import throws
          (broken install / FS corruption); integration-test territory. */
--- a/packages/core/src/maestro-screenshot-file.js
+++ b/packages/core/src/maestro-screenshot-file.js
@@ -15,31 +15,64 @@ export function appAutomateTmpDir() {
   return path.isAbsolute(dir) ? dir : '/tmp';
 }
 
+// Complete session scope root, injected by hosts whose layout no longer fits
+// the {appAutomateTmpDir()}/{sessionId}{_test_suite} convention at all — where
+// relocating the root alone can't express the new shape.
+//
+// realmobile's AAP-18965 relocation is the motivating case: iOS Maestro debug
+// output moved to <aa-tmp>/<device>/logs/maestro_debug_*, which drops the
+// sessionId path segment entirely (the CLI appends it, so no tmp-root value can
+// absorb <device>/logs) and drops the <device>_ prefix the *_maestro_debug_*
+// glob keys on. Android kept its shape under the relocation and stays on
+// PERCY_APP_AUTOMATE_TMP_DIR.
+//
+// When set, this value IS the scope root: the relay globs {root}/**/{name}.png
+// with no layout assumption, and the realpath containment check anchors on it —
+// so the security boundary is relocated, never widened. Non-absolute values are
+// ignored (falling through to the composed convention) rather than producing a
+// cwd-relative root; trailing separators are trimmed for the same
+// prefix-composition reason as appAutomateTmpDir(). BrowserStack-mode only —
+// self-hosted scoping stays on PERCY_MAESTRO_SCREENSHOT_DIR.
+export function bsScopeRootOverride() {
+  let raw = process.env.PERCY_MAESTRO_BS_SCOPE_ROOT;
+  if (!raw) return null;
+  let dir = raw.replace(/[/\\]+$/, '');
+  return path.isAbsolute(dir) ? dir : null;
+}
+
 /* istanbul ignore next — defensive manual directory walker invoked only when
    fast-glob import fails (broken install / FS corruption). Unit tests
    exercise the primary glob path; integration tests on BS hosts exercise
    the walker against real session layouts. Path-traversal sinks inside this
    function are suppressed at file level in .semgrepignore with the same
    rationale (upstream SAFE_ID validation, depth cap, exact filename match). */
-async function manualScreenshotWalk(platform, sessionId, name) {
+async function manualScreenshotWalk(platform, sessionId, name, scopeRoot) {
   const files = [];
-  try {
-    if (platform === 'ios') {
-      const sessionDir = `${appAutomateTmpDir()}/${sessionId}`;
-      const walk = async (dir, depth) => {
-        if (depth > 15) return; // sanity cap
-        let entries;
-        try { entries = await fs.promises.readdir(dir, { withFileTypes: true }); } catch { return; }
-        for (const entry of entries) {
-          const full = path.join(dir, entry.name);
-          if (entry.isDirectory()) {
-            await walk(full, depth + 1);
-          } else if (entry.isFile() && entry.name === `${name}.png` && full.includes('_maestro_debug_')) {
-            files.push(full);
-          }
+  // Shared recursive walker for both root-relative layouts. `accept` gates
+  // which matching files count, so the iOS convention can keep its
+  // `_maestro_debug_` guard while an explicit scope root — where the host has
+  // already narrowed the tree to one session — takes any depth.
+  const walkFrom = async (root, accept) => {
+    const walk = async (dir, depth) => {
+      if (depth > 15) return; // sanity cap
+      let entries;
+      try { entries = await fs.promises.readdir(dir, { withFileTypes: true }); } catch { return; }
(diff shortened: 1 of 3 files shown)


packages/core/test/api.test.js
```diff
@@ -1915,6 +1915,142 @@ describe('API Server', () => {
...
+
+        await expectAsync(postMaestro({ name: SS_NAME, sessionId: SID, platform: 'android' }))
+          .toBeResolvedTo(jasmine.objectContaining({ success: true }));
+
+        let [payload] = percy.upload.calls.mostRecent().args;
+        // Fell through to the default /tmp glob, not a cwd-relative root
+        expect(payload.tiles[0].content).toBe(Buffer.from('PNGBYTES-ANDROID').toString('base64'));
+      });
+
+      it('re-anchors filePath containment on the overridden root', async () => {
```
aryanku-dev, line 2003: **[Medium] No symlink-escape test for the overridden scope root**

The self-hosted arm already has `404s when a globbed file resolves outside the root (symlink escape)`; this override block has no analogue. The containment code is shared, so present-day risk is low — but a future edit that shortcuts realpath for the "host-trusted" root would go undetected, and re-anchored containment is the property this PR explicitly claims.

**Suggestion:** Add a spec here that places a symlink inside `REALMOBILE_DIR` pointing outside `SCOPE_ROOT`, then asserts a 404 matching `/resolved outside session dir/`.

aryanku-dev (commented):
Claude Code Review (automated) — 5 inline finding(s). Full report in the PR comment below. Verdict: Passed.

packages/core/src/maestro-screenshot.js
```diff
@@ -124,10 +129,22 @@ export async function handleMaestroScreenshot(req, res, percy) {
...
+    // Host-injected complete scope root wins over the composed convention. No
+    // existence pre-check here (unlike self-hosted): this is host config, not
+    // customer config, so a stale/missing root should surface as the same 404
+    // the containment check emits rather than a 400 aimed at the customer.
+    let overrideRoot = bsScopeRootOverride();
+    if (overrideRoot) {
+      scopeRoot = overrideRoot;
+      recursiveScope = true;
+      percy.log.debug(`maestro screenshot scope root overridden: ${scopeRoot}`);
+    } else {
```
aryanku-dev, line 139: **[Low] A rejected override falls through silently**

When `PERCY_MAESTRO_BS_SCOPE_ROOT` is set but rejected by `bsScopeRootOverride()` (non-absolute, or reducing to `/`), this branch composes the convention root with no log line. The accepted path logs at debug; the rejected path is silent, so a host-side typo surfaces only as 404s indistinguishable from an ordinary missing file.

**Suggestion:** Emit `percy.log.warn` here when `process.env.PERCY_MAESTRO_BS_SCOPE_ROOT` is truthy but the helper returned null, mirroring the existing debug log.

packages/core/src/maestro-screenshot-file.js
```diff
@@ -15,31 +15,64 @@ export function appAutomateTmpDir() {
...
+// so the security boundary is relocated, never widened. Non-absolute values are
+// ignored (falling through to the composed convention) rather than producing a
+// cwd-relative root; trailing separators are trimmed for the same
+// prefix-composition reason as appAutomateTmpDir(). BrowserStack-mode only —
+// self-hosted scoping stays on PERCY_MAESTRO_SCREENSHOT_DIR.
+export function bsScopeRootOverride() {
+  let raw = process.env.PERCY_MAESTRO_BS_SCOPE_ROOT;
+  if (!raw) return null;
+  let dir = raw.replace(/[/\\]+$/, '');
+  return path.isAbsolute(dir) ? dir : null;
```
aryanku-dev, line 26: **[Low] Rejection of root `/` is incidental to the trim regex, not an explicit guard**

`'/'` is rejected only because the trailing-separator strip turns it into `''`, which then fails `path.isAbsolute`. The behaviour is pinned by a test, but the safety property — an override can never widen the root to the whole filesystem — is incidental in the code, so a future refactor of the trim could silently reintroduce it.

**Suggestion:** Record the intent at the return, e.g. `// '/' trims to '' and is rejected here — intentional; prevents an override that would recurse the whole filesystem.`

packages/core/test/api.test.js
```diff
@@ -1915,6 +1915,142 @@ describe('API Server', () => {
...
+
+        await expectAsync(postMaestro({
+          name: FILEPATH_NAME,
+          sessionId: SID,
+          platform: 'ios',
+          filePath: `${IOS_FILEPATH_DIR}/${FILEPATH_NAME}.png`
+        })).toBeRejectedWithError(/Screenshot not found/);
+      });
+
+      it('404s when the overridden root does not exist', async () => {
```
aryanku-dev, line 2027: **[Low] No multi-match mtime tie-break test for the override glob**

`recursiveScope` routes the override through the same `files.length > 1` mtime-descending sort, but no spec exercises it. This matters more here than for the convention glob precisely because the layout constraint was removed: a same-named PNG from a second flow anywhere under the session root can now match.

**Suggestion:** Add a spec with two `${SS_NAME}.png` fixtures at different depths and mtimes under `SCOPE_ROOT`, asserting the newer one is chosen.

packages/core/src/maestro-screenshot.js
```diff
@@ -124,10 +129,22 @@ export async function handleMaestroScreenshot(req, res, percy) {
...
   } else {
-    scopeRoot = platform === 'ios'
-      ? `${appAutomateTmpDir()}/${sessionId}`
-      : `${appAutomateTmpDir()}/${sessionId}_test_suite`;
+    // Host-injected complete scope root wins over the composed convention. No
+    // existence pre-check here (unlike self-hosted): this is host config, not
+    // customer config, so a stale/missing root should surface as the same 404
+    // the containment check emits rather than a 400 aimed at the customer.
+    let overrideRoot = bsScopeRootOverride();
+    if (overrideRoot) {
```
aryanku-dev, line 135: **[Low] Override applies to Android too — worth documenting on the host side**

Design note, not a defect. The override is applied regardless of `platform`, so an Android session under it also loses the `_test_suite/logs/*/screenshots` structural guard. That is intentional and tested (`applies to android too`), but the PR description frames the feature as iOS-only, and nothing in code prevents a host from setting the var too broadly and quietly loosening Android's guard.

**Suggestion:** State in the BS-host runbook that this is set per-invocation for iOS sessions, not host-wide, unless Android's layout has also changed.

aryanku-dev:
# Claude Code PR Review

**PR:** https://github.com/percy/cli/pull/2357  •  **Head:** 0b084ed  •  **Reviewers:** stack-code-reviewer

## Summary

Adds `PERCY_MAESTRO_BS_SCOPE_ROOT` so a BrowserStack host can inject a complete Maestro session scope root for layouts the `{appAutomateTmpDir()}/{sessionId}{_test_suite}` convention can no longer express (the iOS realmobile relocation); when set, the relay globs `{root}/**/{name}.png`, the manual-walker fallback recurses that root, and the existing realpath containment check re-anchors on it.

## Review Table

| High | Security | No hardcoded secrets or credentials | Pass | Config read from `process.env` only; nothing embedded. |
| High | Security | Authentication/authorization checks present | Pass | No auth surface added. The relay's authz-analog is the realpath containment check, which is preserved (see next rows). |
| High | Security | Input validation and sanitization | Pass | `bsScopeRootOverride()` rejects non-absolute values and (via the trailing-separator trim) `/`; `name` remains SAFE_ID-validated upstream. |
| High | Security | No IDOR — resource ownership validated | Pass | Containment logic at `maestro-screenshot-file.js:193-205` is untouched and generic over `scopeRoot`: both file and root are `realpath`'d, then compared with `startsWith(prefix + '/')`. The trailing slash defeats sibling-prefix collisions (`/root` vs `/root-evil`), and realpath defeats symlink swaps. Boundary is relocated, not widened — the PR's central claim holds on inspection. |
| High | Security | No SQL injection (parameterized queries) | N/A | No database access in this change. |
| High | Correctness | Logic is correct, handles edge cases | Pass | `recursiveScope` is derived in `locateScreenshot` rather than trusted from the caller, so `selfHosted` keeps its existing meaning; trailing-slash, non-absolute, and missing-root cases all behave as documented. |
| High | Correctness | Error handling is explicit, no swallowed exceptions | Pass | Missing/out-of-root still throws `ServerError(404)`. The deliberate omission of an existence pre-check (vs self-hosted's 400) is reasoned in-code: host config should not produce a customer-facing 400. |
| High | Correctness | No race conditions or concurrency issues | Pass | Multi-match tie-break (mtime-descending, `:177-181`) is pre-existing and unchanged; the override only widens what reaches it. |
| Medium | Testing | New code has corresponding tests | Pass | 8 new specs covering locate, Android parity, precedence over `PERCY_APP_AUTOMATE_TMP_DIR`, trailing slash, non-absolute fallback, `filePath` re-anchoring, missing root, and helper trim/null semantics. |
| Medium | Testing | Error paths and edge cases tested | Fail | Two gaps in exactly the risk-prone areas: no symlink-escape test for the override root (the self-hosted arm has one), and no multi-match tie-break test for the now-unbounded override glob. See Findings 1 and 4. |
| Medium | Testing | Existing tests still pass (no regressions) | Unverified | No clean local run was achievable. Two unrelated environmental failure classes hit the file: a Chromium snapshot download 404, and `ENOTEMPTY` from `rmSync` cleanup that fires identically in the **pre-existing, untouched** `PERCY_APP_AUTOMATE_TMP_DIR override` block. This corroborates the PR's own account, but it means CI is the only place the new assertions get exercised — **confirm this file is green in CI before merge.** |
| Medium | Performance | No N+1 queries or unbounded data fetching | Pass | The override glob is broader than the convention glob (no `*_maestro_debug_*` / `_test_suite/logs` segment requirement), but bounded by a host-narrowed per-session root, and the walker keeps its depth-15 cap. |
| Medium | Performance | Long-running tasks use background jobs | N/A | Single-request file lookup. |
| Medium | Quality | Follows existing codebase patterns | Pass | Mirrors `appAutomateTmpDir()`'s validate-and-trim shape and reuses the self-hosted recursive-glob path rather than inventing new containment logic. |
| Medium | Quality | Changes are focused (single concern) | Pass | Three files, one concern; the walker refactor into `walkFrom(root, accept)` is in service of it. |
| Low | Quality | Meaningful names, no dead code | Pass | `bsScopeRootOverride`, `recursiveScope`, `scopedGlob` all read clearly. |
| Low | Quality | Comments explain why, not what | Pass | Unusually thorough — the motivating layout change, the security rationale, and the 404-vs-400 choice are all recorded at the point of decision. |

## Findings

**1.**
- **File:** `packages/core/test/api.test.js:2009`
- **Severity:** Medium
- **Issue:** No symlink-escape test exists for the new override root. The self-hosted arm already has `404s when a globbed file resolves outside the root (symlink escape)`; the override block has no analogue. The containment code is shared so present-day risk is low, but a future edit that shortcuts realpath for the "host-trusted" root would go undetected — and this is the one property the PR explicitly claims.
- **Suggestion:** Add a spec in the `PERCY_MAESTRO_BS_SCOPE_ROOT override` block: place a symlink inside `REALMOBILE_DIR` pointing outside `SCOPE_ROOT`, then assert a 404 matching `/resolved outside session dir/`.

**2.**
- **File:** `packages/core/src/maestro-screenshot.js:143`
- **Severity:** Low
- **Issue:** When `PERCY_MAESTRO_BS_SCOPE_ROOT` is set but rejected (non-absolute, or reducing to `/`), the request falls through to the composed convention with no log line. The accepted path logs at debug; the rejected path is silent, so a host-side typo surfaces only as 404s indistinguishable from an ordinary missing file.
- **Suggestion:** In the fall-through `else`, emit `percy.log.warn` when `process.env.PERCY_MAESTRO_BS_SCOPE_ROOT` is truthy but `bsScopeRootOverride()` returned null, mirroring the existing debug log.

**3.**
- **File:** `packages/core/src/maestro-screenshot-file.js:40`
- **Severity:** Low
- **Issue:** Rejecting the literal value `/` is a side effect of the trailing-separator regex (`'/'` → `''`, which fails `path.isAbsolute`) rather than an explicit guard. The behaviour is pinned by a test, but the safety property — an override can never widen the root to the entire filesystem — is incidental in the code, so a future refactor of the trim could silently reintroduce it.
- **Suggestion:** Add a one-line comment at the return: `// '/' trims to '' and is rejected here — intentional; prevents an override that would recurse the whole filesystem.`

**4.**
- **File:** `packages/core/test/api.test.js:2034`
- **Severity:** Low
- **Issue:** No multi-match tie-break test for the override glob, though `recursiveScope` routes it through the same `files.length > 1` mtime sort. This matters more here than for the convention glob precisely because the layout constraint was removed: a same-named PNG from a second flow anywhere under the session root can now match.
- **Suggestion:** Add a spec with two `${SS_NAME}.png` fixtures at different depths and mtimes under `SCOPE_ROOT`, asserting the newer one is chosen.

**5.**
- **File:** `packages/core/src/maestro-screenshot.js:139`
- **Severity:** Low
- **Issue:** Design/documentation note, not a defect. The override is applied regardless of `platform`, so an Android session under it also loses the `_test_suite/logs/*/screenshots` structural guard — intentional and tested (`applies to android too`), but the PR description frames the feature as iOS-only ("Android kept its shape... stays on `PERCY_APP_AUTOMATE_TMP_DIR`"). Nothing in code prevents a host from setting it too broadly and quietly loosening Android's guard.
- **Suggestion:** State in the BS-host runbook that this must be set per-invocation for iOS sessions, not host-wide, unless Android's layout has also changed.

---

**Verdict: PASS** — the security boundary is genuinely relocated rather than widened, and the reasoning is well documented; the open items are test-coverage and observability gaps, none gating. Please confirm `packages/core/test/api.test.js` is green in CI before merge, since the new specs could not be exercised locally.

packages/core/src/maestro-screenshot-file.js
```diff
@@ -15,31 +15,64 @@ export function appAutomateTmpDir() {
...
 
+// Complete session scope root, injected by hosts whose layout no longer fits
+// the {appAutomateTmpDir()}/{sessionId}{_test_suite} convention at all — where
+// relocating the root alone can't express the new shape.
+//
+// realmobile's AAP-18965 relocation is the motivating case: iOS Maestro debug
+// output moved to <aa-tmp>/<device>/logs/maestro_debug_*, which drops the
+// sessionId path segment entirely (the CLI appends it, so no tmp-root value can
+// absorb <device>/logs) and drops the <device>_ prefix the *_maestro_debug_*
+// glob keys on. Android kept its shape under the relocation and stays on
```
Shivanshu-07: Too many comments

packages/core/src/maestro-screenshot-file.js
```diff
@@ -15,31 +15,55 @@ export function appAutomateTmpDir() {
   return path.isAbsolute(dir) ? dir : '/tmp';
 }
 
+// Complete session scope root for BS hosts whose layout the
```
aryanku-dev: trim further

aryanku-dev:
@Shivanshu-07 trimmed in `b49b123` — the comments this PR adds are down from 55 lines to 22, and `bsScopeRootOverride`'s doc block from 17 to 4.

Kept only what the code can't show for itself: why the `{tmpRoot}/{sessionId}` convention can't express the realmobile layout, and that the override relocates the realpath containment boundary rather than widening it. Everything that restated the line below it is gone.

Comments only — no code change, ESLint clean.

PR #2366, opened 2026-08-04
PER-8985 fix: ask for the full access token when the seed-build wait 403s

Shivanshu-07:
## What

When the drop-in's baseline seed wait polls `GET /builds/:id` with the project's default **write-only** token, the read 403s (build reads are master/read_only-gated server-side). Previously that surfaced as the generic, misleading timeout warning:

```
[percy] Baseline build did not finish processing in time (state: unknown) — snapshots in this run may show as new instead of diffing against the baseline
```

Now the auth failure is classified (`isAuthFailure`: response status 401/403 or message match), polling stops on the first failure (retries can never succeed), and the user gets targeted guidance **before their tests run**, while switching tokens can still save the first run:

```
[percy] Uploaded 2/2 baseline snapshot(s), but this token cannot read build status, so Percy cannot wait for your baseline to finish before tests start. Use your project's FULL ACCESS token as PERCY_TOKEN for this first run (Project settings → Tokens). If snapshots in this run appear as new instead of diffing, approve build #1 in the dashboard.
```

Observed live during beta validation (prod project `dropin-web-test-71650540`, `@percy/cli@1.32.6-beta.0`).

## Why this approach

Chosen over the API-side status-read carve-out (percy/percy-api#6648, now closed): keep the token model untouched and ask users to run the seeded first build with the full access token — this warning is the ask, fired exactly when it applies.

## Testing

- `waitForSeedBuild` returns `'unauthorized'` on 401/403 (response-status and message-only shapes) and stops after one poll
- `maybeSeedBaseline` prints the full-access-token guidance (and not the generic timeout warning); non-auth read errors keep the existing degrade path
- cli-exec suite: 72 specs, sole failure is the known pre-existing SIGTERM-interrupt timing flake (fails identically on pristine master)

### DIFF
--- a/packages/cli-exec/src/baseline.js
+++ b/packages/cli-exec/src/baseline.js
@@ -124,18 +124,35 @@ export async function findBaselineProvider({ cwd = process.cwd(), log } = {}) {
   return null;
 }
 
+// The seed-build status poll needs a READ-capable token, but `percy exec` normally runs with the
+// project's default write-only token — the read then 403s (build reads are master/read_only-gated
+// server-side). That's a token-choice problem the user can fix, not a transient failure, so it is
+// classified here and surfaced as its own sentinel instead of the generic wait-timeout warning.
+function isAuthFailure(error) {
+  let status = error?.response?.statusCode;
+  if (status === 401 || status === 403) return true;
+  return /\b(401|403|forbidden|unauthori[sz]ed)\b/i.test(error?.message || '');
+}
+
 // The seed build keeps processing (renders + auto-approval) after finalize. The head build must
 // not start until it reaches a terminal state — head snapshots select their baseline as they are
 // processed, and an unapproved seed means the whole first run shows as new instead of diffing.
 // The timeout matches the pipeline latency budget (~99% of builds finish under 5 minutes) —
 // a seed of committed screenshots still renders server-side, so first runs can hold for minutes.
+// Returns the terminal state, or 'unauthorized' when the token cannot read build status at all
+// (every retry would fail identically, so polling stops on the first auth failure).
 export async function waitForSeedBuild(client, buildId, { log, timeout = 600000, interval = 5000 }) {
   let deadline = Date.now() + timeout;
   let state = 'pending';
   let polls = 0;
 
   for (;;) {
-    ({ state } = (await client.getBuild(buildId)).data.attributes);
+    try {
+      ({ state } = (await client.getBuild(buildId)).data.attributes);
+    } catch (err) {
+      if (isAuthFailure(err)) return 'unauthorized';
+      throw err;
+    }
     if (state !== 'pending' && state !== 'processing') return state;
     if (Date.now() >= deadline) return state;
     // A visible heartbeat every ~30s so a multi-minute first-run hold doesn't look like a hang.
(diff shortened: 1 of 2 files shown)


rishigupta1599 (approved):

aryanku-dev (approved):

Shivanshu-07:
# Claude Code PR Review

**PR:** https://github.com/percy/cli/pull/2366  •  **Head:** 2d8129d  •  **Reviewers:** fallback inline checklist

## Summary
Classifies the 401/403 the seed-build status poll hits under `percy exec`'s default write-only token, stops polling on the first auth failure, and replaces the misleading `state: unknown` timeout warning with targeted guidance to use the project's full access token — emitted after the seed uploads and before the head build's tests run.

## Review Table

| High | Security | No hardcoded secrets or credentials | Pass | Message text only; test tokens are fakes |
| High | Security | Authentication/authorization checks present | N/A | Client-side handling of a server-side authz denial; no auth surface changed |
| High | Security | Input validation and sanitization | Pass | Classifier reads only the caught error's status/message; regex is anchored with word boundaries (`4030` does not match `\b403\b`) |
| High | Correctness | Logic is correct, handles edge cases | Pass | Auth failure returns the `'unauthorized'` sentinel on the FIRST poll (retrying an authz denial can never succeed); non-auth errors rethrow into the pre-existing degrade path unchanged; sentinel handled before the generic else so the two warnings are mutually exclusive. E2E-verified against a live 403 (local percy-api, write-only token) and the polling path (master token, heartbeats intact) |
| High | Correctness | Error handling is explicit, no swallowed exceptions | Pass | The try/catch narrows to exactly the classified case; everything else propagates as before. Pathological non-Error throws (e.g. `throw null`) would TypeError inside the classifier and land in the caller's existing catch → generic warn — degraded but safe |
| High | Correctness | No race conditions or concurrency issues | Pass | The head-build race on write-only tokens is pre-existing and now *documented to the user* at the moment it applies; no new concurrency introduced |
| Medium | Testing | New code has corresponding tests | Pass | 4 new specs: 403-with-response (asserts single poll + new message + generic message absent), message-only 401, message-less non-auth error (rethrow path), plus the sentinel wiring through `maybeSeedBaseline` |
| Medium | Testing | Error paths and edge cases tested | Pass | Both classifier branches (response status, message regex) and the empty-message fallback arm covered — this was the CI coverage-gate fix |
| Medium | Testing | Existing tests still pass (no regressions) | Pass | CI 46/46 green on head; locally the only failure is the long-known SIGTERM-interrupt timing flake (fails identically on pristine master) |
| Medium | Performance | No N+1 queries or unbounded data fetching | Pass | Strictly fewer requests than before: polling stops on the first auth failure instead of throwing per-attempt |
| Medium | Quality | Follows existing codebase patterns | Pass | Sentinel-return style matches the wait's existing state returns; message tone/shape matches the surrounding seeding notices; classifier mirrors the SDK sync classifier's auth detection |
| Medium | Quality | Changes are focused (single concern) | Pass | 2 files, +72/−1 |
| Low | Quality | Meaningful names, no dead code | Pass | `isAuthFailure` / `'unauthorized'` sentinel are self-describing |
| Low | Quality | Comments explain why, not what | Pass | Comments carry the token-model rationale and the fires-before-tests timing argument |
| Low | Quality | No unnecessary dependencies added | Pass | None |

## Findings

- **File:** `packages/cli-exec/src/baseline.js:132`
- **Severity:** Low
- **Reviewer:** fallback inline checklist
- **Issue:** The message-regex fallback can misclassify a non-auth error whose text merely contains "403"/"forbidden" (e.g. a proxy error page quoted into the message) as an auth failure, showing the full-access-token advice erroneously.
- **Suggestion:** Acceptable: the response-status check runs first (authoritative for real API errors), the consequence is a single advisory warning rather than lost functionality, and the same detection shape is already used by the SDK's sync classifier. No change requested.

---

**Verdict: PASS**

PR #2388, opened 2026-08-18
feat(webdriver-utils): add scaleToFit for full-page Automate captures (PER-10530)

rishigupta1599:
## Why

Full-page POA captures are bounded by a 50,000px stitched-image budget. mobile-common enforces it by truncating the DOM walk at `MAX_PIXEL_HEIGHT_LIMIT / dpr` — **16,666 CSS px at DPR 3** — so any page taller than that silently loses its tail. That is the PER-10298 / PPLT-5851 dealblocker (Adecco, Akkodis).

`scaleToFit` opts a capture into downscaling every tile by a fixed `1/dpr` instead, so a page of up to 50,000 **CSS** px lands exactly on the same budget: `page_height × dpr × (1/dpr) == page_height`.

## Why the factor is fixed, not height-derived

Percy diffs pixel to pixel, and a uniform resize moves **both** dimensions. A height-derived factor would make output **width** a function of page height — so a page drifting from 16,600 to 18,000 px would change width, leaving snapshot and baseline with *different dimensions*: unalignable, not merely different.

Constant width forces a constant factor, and the ceiling forces `f ≤ 50000/(50000 × 3) = 1/3`. The three constraints (constant width, ≤ 50,000px, coverage to 50,000 CSS px) are only jointly satisfiable at `f = 1/dpr`. It is also opt-in, so quality never changes for anyone who does not ask for it.

## Changes

**`packages/core/src/config.js`** — declare `scaleToFit` on the snapshot schema. That section sets `additionalProperties: false`, so without a declaration the key is **rejected before it ever reaches the Automate session**. Gated `onlyAutomate`, exactly like `fullPage`.

**`packages/webdriver-utils/src/providers/genericProvider.js`** — `PERCY_SCALE_TO_FIT` in `addDefaultOptions()`, so a whole run can opt in without editing per-snapshot config (how support would enable this for an affected customer). Coerced to a real boolean, because mobile-common compares with `== true` and a truthy string would silently no-op.

**No provider changes needed.** `automateProvider.getTiles()` already forwards `options` wholesale and SeleniumHub passes it through verbatim, so the camelCase spelling is the contract with mobile-common.

## Notes for review

Two things I got wrong first and corrected, worth knowing if you extend these tests:

- Automate options live on **`/config/snapshot`**, not `/snapshot` — the latter `$ref`s only a hand-picked subset and reports even `fullPage` as `unknown property`. My first specs validated against `/snapshot` and passed **vacuously**.
- `onlyAutomate` is evaluated when AJV **compiles** the schema, not per-validation, so flipping `PERCY_TOKEN` inside a spec cannot change the outcome. The specs now assert the declaration structurally *plus* the real validation error, and explicitly assert the message is not `unknown property` so they cannot pass vacuously again.

## Testing

- `packages/webdriver-utils`: **240/240**
- `packages/core` `test/unit/config.test.js`: **21/21**
- `eslint` clean on all four files

The full `packages/core` suite does not pass in my environment — 155 failures, all `Failed to launch browser / Timed out after 30000ms`, i.e. Chrome cannot start here. Unrelated to this change, but I could not use it as a signal, so I ran the config specs in isolation.

## Depends on

browserstack/mobile-common#1255 — the capture-side implementation. This PR only plumbs the option; without it the option is accepted and ignored.

### DIFF
--- a/packages/cli-doctor/src/checks/config.js
+++ b/packages/cli-doctor/src/checks/config.js
@@ -80,7 +80,7 @@ export async function checkConfig(options = {}) {
 
     // Keys that only work with automate tokens
     const automateOnlyKeys = ['fullPage', 'freezeAnimation', 'freezeAnimatedImage',
-      'freezeAnimatedImageOptions', 'ignoreRegions', 'considerRegions'];
+      'freezeAnimatedImageOptions', 'ignoreRegions', 'considerRegions', 'scaleToFit'];
     // Keys that only work with web tokens (not automate, not app)
     const webOnlyKeys = ['waitForTimeout', 'waitForSelector'];
 
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -202,10 +202,8 @@ export const configSchema = {
         type: 'boolean',
         onlyAutomate: true
       },
-      // Opts a full-page Automate capture into downscaling each tile by 1/dpr, so the
-      // stitched image stays inside the 50,000px comparison ceiling. Without it the
-      // walkable DOM height is capped at 50000/dpr CSS px and taller pages are truncated.
-      // Only honoured on the fullPage path -- mobile-common ignores it for singlepage.
+      // Downscales each tile by 1/dpr so the stitched image fits the 50,000px ceiling;
+      // without it the DOM walk is capped at 50000/dpr CSS px. fullPage path only.
       scaleToFit: {
         type: 'boolean',
         onlyAutomate: true
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -202,6 +202,14 @@ export const configSchema = {
         type: 'boolean',
         onlyAutomate: true
       },
+      // Opts a full-page Automate capture into downscaling each tile by 1/dpr, so the
+      // stitched image stays inside the 50,000px comparison ceiling. Without it the
+      // walkable DOM height is capped at 50000/dpr CSS px and taller pages are truncated.
+      // Only honoured on the fullPage path -- mobile-common ignores it for singlepage.
+      scaleToFit: {
+        type: 'boolean',
+        onlyAutomate: true
+      },
       freezeAnimation: { // for backward compatibility
         type: 'boolean',
         onlyAutomate: true
--- a/packages/core/src/config.js
+++ b/packages/core/src/config.js
@@ -933,7 +933,11 @@ export const comparisonSchema = {
         },
         cliScreenshotStartTime: { type: 'integer', default: 0 },
         cliScreenshotEndTime: { type: 'integer', default: 0 },
-        screenshotType: { type: 'string', default: 'singlepage' }
+        screenshotType: { type: 'string', default: 'singlepage' },
+        // additionalProperties is false here and PercyConfig.validate DELETES unknown keys
+        // from the object it validates, so an undeclared key is silently dropped pre-upload.
+        scaleToFit: { type: 'boolean' },
+        appliedScaleFactor: { type: 'number', exclusiveMinimum: 0, maximum: 1 }
       }
     },
     tag: {
--- a/packages/core/src/utils.js
+++ b/packages/core/src/utils.js
@@ -412,6 +412,9 @@ export function percyAutomateRequestHandler(req, percy) {
 
   req.body.options = merge([{
     fullPage: percy.config.snapshot.fullPage,
+    // Must be listed here or a .percy.yml `snapshot.scaleToFit` validates and is then
+    // silently dropped -- the page truncates as before while the build stays green.
+    scaleToFit: percy.config.snapshot.scaleToFit,
     percyCSS: percy.config.snapshot.percyCSS,
     freezeAnimatedImage: percy.config.snapshot.freezeAnimatedImage || percy.config.snapshot.freezeAnimation,
     freezeImageBySelectors: percy.config.snapshot.freezeAnimatedImageOptions?.freezeImageBySelectors,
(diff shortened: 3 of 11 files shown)


rishigupta1599:
# Claude Code PR Review

**PR:** https://github.com/percy/cli/pull/2388  •  **Head:** c3286cf  •  **Reviewers:** stack-code-reviewer

## Summary

Adds an opt-in `scaleToFit` snapshot option for full-page Automate/POA captures: declares the new keys on the snapshot and comparison schemas, gates the option to `fullPage` only on both the automate and playwright provider paths, and forwards the host-reported `appliedScaleFactor` to percy-api as comparison metadata.

## Review Table

| High | Security | No hardcoded secrets or credentials | Pass | — |
| High | Security | Authentication/authorization checks present | N/A | — |
| High | Security | Input validation and sanitization | Pass | Host-reported factor is range-checked `(0,1]` and `Number.isFinite`-guarded before use |
| High | Correctness | Logic is correct, handles edge cases | Pass | Both focus areas verified by runtime trace, not just structurally — see below |
| High | Correctness | Error handling is explicit, no swallowed exceptions | Pass | — |
| High | Correctness | No race conditions or concurrency issues | N/A | — |
| Medium | Testing | New code has corresponding tests | **Fail** | Finding 1: the one-line fix at the heart of this PR has no regression test |
| Medium | Testing | Error paths and edge cases tested | Pass | Four boolean-logic branches of the gate traced against the implementation; all discriminate |
| Medium | Testing | Existing tests still pass (no regressions) | Pass | 246 + 163 specs pass locally; CI green apart from the Percy visual gate |
| Medium | Performance | No N+1 queries or unbounded data fetching | N/A | — |
| Medium | Quality | Follows existing codebase patterns | Pass | — |
| Medium | Quality | Changes are focused (single concern) | Pass | — |
| Low | Quality | Meaningful names, no dead code | **Fail** | Finding 3: `scaleFactor` (DPR, outbound) vs `appliedScaleFactor` (downscale ratio, inbound) is a readability trap |
| Low | Quality | Comments explain why, not what | Pass | — |
| Low | Quality | No unnecessary dependencies added | Pass | — |

## Focus areas — both verified clean

**Schema declaration survives `PercyConfig.validate`.** Confirmed by tracing the real upload call site, not just the diff: `percy.js:711-719` shallow-destructures `let { clientInfo, environmentInfo, labels, ...comparison } = options`, so `comparison.metadata` is the *same object reference* as `options.metadata`, and the object actually pushed to the upload queue is `options` (`percy.js:749`). Since `validate` mutates in place and deletes undeclared keys from nested objects, an undeclared metadata key would be stripped from `options.metadata` too — silently, pre-upload. Both `scaleToFit` and `appliedScaleFactor` are declared on `comparisonSchema.properties.metadata.properties` (`config.js:939-940`, `additionalProperties: false`), and `scaleToFit` is additionally declared on `configSchema.snapshot.properties` (`config.js:207-210`, registered at runtime as `/config/snapshot`). `screenshotType` was already declared. No leak.

**The `fullPage` gate runs on both provider paths.** `AutomateProvider.screenshot()` calls `super.screenshot()`, which invokes `addDefaultOptions()` at `genericProvider.js:183`. `PlaywrightProvider.screenshot()` never calls `super`, and this PR adds an explicit `this.addDefaultOptions()` at `playwrightProvider.js:48` — which does close the gap. See finding 1 for the coverage caveat.

## Findings

- **File:** `packages/webdriver-utils/test/providers/playwrightProvider.test.js:70-169`
- **Severity:** Medium
- **Issue:** the fix this PR exists to make on the playwright path is the single added line `this.addDefaultOptions()` in `screenshot()`, and nothing would notice if it were deleted. The `describe('screenshot')` tests stub `provider.getTiles` wholesale and never inspect `provider.options` afterwards; the new `getTiles()`-focused specs construct the provider with already-gated options (`{ fullPage: true, scaleToFit: true }`) and call `getTiles()` directly, bypassing `screenshot()`; and the `addDefaultOptions` unit tests in `genericProvider.test.js` exercise the method in isolation, not its invocation from `PlaywrightProvider.screenshot()`.
- **Suggestion:** one test that goes through `screenshot()` with `{ fullPage: false, scaleToFit: true }` and asserts `provider.options.scaleToFit` is false afterwards. That fails today if the line is removed.

- **File:** `packages/webdriver-utils/src/providers/playwrightProvider.js:48`
- **Severity:** Low
- **Issue:** calling `addDefaultOptions()` also newly applies the `freezeAnimation` coercion to the playwright path, which previously never ran that line at all. Very likely a desirable alignment with the automate path, but it is an unflagged behaviour change riding along inside a `scaleToFit`-titled fix.
- **Suggestion:** call it out in the PR description so downstream consumers know `freezeAnimation` defaulting now applies to playwright captures too.

- **File:** `packages/webdriver-utils/src/providers/genericProvider.js:55`
- **Severity:** Low
- **Issue:** a non-boolean `scaleToFit` (e.g. the string `'true'` from a hand-edited `.percy.yml`) is deliberately treated as *disabling* the feature. That is the intended design given mobile-common compares with `== true`, but the user-visible result is "feature silently off, no error".
- **Suggestion:** `log.warn` when `scaleToFit` is defined and not a boolean, so a misconfiguration is not mistaken for "scaling wasn't needed".

### Low — naming

`scaleFactor` (device pixel ratio, sent **to** the host) and `appliedScaleFactor` (downscale ratio, received **from** the host) sit adjacent in the same metadata block in both providers. Functionally correct and non-overlapping; worth an explicit comment that the two are unrelated.

## Test quality

No vacuous tests found. Each new test was checked against the implementation and would fail if the guard it covers were deleted — including the `merge()` trace confirming an explicit `scaleToFit: false` correctly overrides a `true` default rather than being skipped as falsy.

## Note for the author (not a finding)

An untracked `cli/.percy.yml` in the local working tree is picked up by cosmiconfig walking upward from `packages/core` and causes several hundred spurious local spec failures. It does not exist in CI.

---

**Verdict: PASS** — no High or Critical findings. Finding 1 is worth taking before merge: it is the only regression guard missing on the line this PR turns on.

ninadbstack (approved):

PR #2418, opened 2026-09-07
PPLT-6073: native PDF visual testing via POST /percy/pdf/snapshot

RaghavsBrowserStack:
## What

Adds first-class PDF visual testing to the CLI. An SDK hands over PDF bytes; the CLI rasterizes the document and creates **one Percy snapshot per page**, optionally blocking on synchronous comparison results.

This is the in-CLI replacement for the external [`percy-pdf`](https://github.com/BrowserStackCE/percy-pdf) solution, which can be retired once this ships.

```js
const { postPdfSnapshot } = require('@percy/sdk-utils');

const result = await postPdfSnapshot({
  name: 'Burglary Insurance Policy',
  sync: true,
  pdf: { content: fs.readFileSync('policy.pdf').toString('base64') },
  excludePages: [4, 7]
});
// result.body.data.pages[] -> one entry per page, each with its own diff-info
```

## Why not the old approach

`percy-pdf` wrapped the CLI from the outside: it unzipped a 5.8 MB pdf.js viewer, served it over `http-server`, generated a `percy snapshot` YAML config per document, and captured pages by having Percy's **renderer** run the viewer and deleting already-captured `div.page` children from the DOM between snapshots so `scope: .canvasWrapper` would land on the next page.

That design carried three bugs that are unrepresentable in this one:

1. **Page filtering mislabelled snapshots.** The per-page `execute` script was assigned inside a `forEach` over every previously-pushed snapshot, so all snapshots ended up with the *last* page's script. Any gap from `excludePages` meant every later page captured the wrong page under the right name.
2. **Page 1 and the second-to-last page could not be excluded** — page 1 was structurally the base snapshot, and the second-to-last was entangled with the viewer's page-prefetch step. `excludePages: [1]` now just works.
3. **Page counts were read from the baseline directory even when comparing a release**, so a release with extra pages silently lost them.

### Rendering happens in the discovery browser

The CLI already launches Chromium eagerly — `percy.start()` runs `if (!this.skipDiscovery) yield this.#discovery.start()`, whose `start` handler calls `percy.browser.launch()`. During a PDF-only run that browser sits **completely idle**: every `discovery queueInfo` line reports `{"queued":0,"pending":0,"total":0}`, because `percy.upload()` bypasses `discoverSnapshotResources`.

So pdf.js runs in a browser page rather than in Node. An earlier revision of this PR rasterized with `@napi-rs/canvas` (Skia); that shipped ~25MB of platform-specific native prebuilds to duplicate a renderer already present and running, and has been removed. A second benefit: rendering is now the CLI's **pinned** Chromium revision, so page rasters are as reproducible as the rest of Percy's pipeline instead of tracking a separately versioned native dependency.

### `@percy/cli-pdf` (new package) — helpers and pdf.js assets

No reference to `@percy/core`, which is what lets core list it as an `optionalDependency` without a cycle. Its only heavyweight cost is pdfjs-dist's ~34MB of library, fonts and cmaps, which nobody installs unless they snapshot a PDF.

| Module | Responsibility |
|---|---|
| `pages.js` | `resolvePages()` — parses `3`, `[1,2,5]`, `"1-5"`, `"1,3,8"`, `"2-"`; applies `excludePages`; validates against the real page count |
| `browser-scripts.js` | The functions that execute **inside the page** (`openDocument`, `measurePages`, `renderPage`, `destroyDocument`) plus the scale/dimension limits |
| `assets.js` | `pdfjsAssets()` — the installed pdfjs-dist paths the asset server and page injection need |

The wrapper DOM is **not** in this package — it is shared with `@percy/cli-upload`, see below.

### `@percy/core/src/pdf-rasterize.js` — the browser work

`rasterizePdf(percy, buffer, options)` stands up a throwaway loopback origin (`Server.serve`) exposing pdf.js, its worker, `standard_fonts`, `cmaps` and the document itself; opens an isolated page on it; injects pdf.js; then measures and renders each selected page to a PNG data URL.

**Serving the assets over a real origin is what makes standard fonts work.** pdf.js fetches `standardFontDataUrl` / `cMapUrl` over HTTP at render time, and the base-14 fonts (Helvetica, Times…) aren't embedded in most documents — without a reachable origin pdf.js renders wrong metrics or drops glyphs and only *warns*. `isEvalSupported: false` is still passed, because the PDF itself is untrusted input.

It calls `percy.browser.launch()` explicitly. That's idempotent (`if (this.readyState != null) return`), and it makes PDF snapshots work under `skipDiscovery`, where the eager launch doesn't happen.

### `@percy/core` — the endpoint

`POST /percy/pdf/snapshot` → `src/pdf-snapshot.js`, which validates the request, decodes the base64 document (50 MB cap and `%PDF-` magic-byte check, matching `/percy/comparison/upload`), lazily imports `@percy/cli-pdf`, and fans out.

Each page carries `resources` and **no `tag`**, so `createSnapshotsQueue`'s task handler picks `client.sendSnapshot` over `sendComparison`: these are real web snapshots, not comparisons. `resources` is passed as a function (the lazy-resource pattern `cli-upload` uses) so the root-DOM and resource-object construction is deferred into the queue task.

### Concurrency, precisely

| Stage | Behaviour |
|---|---|
| Rendering pages → PNG | **Sequential**, one `page.eval` per page against a single browser page. |
| Queuing the N snapshots | **All at once, non-blocking.** All pages are pushed via `percy.upload()` in a single synchronous pass. |
| Uploading | **Concurrent**, up to the snapshots queue's `concurrency` (`discovery.concurrency`, default 10). |
| Polling for completion | **One shared batched poll.** `WaitForJob` collects every pending job id and issues a single `getStatus` (`job_status?...&id=a,b,c`) per interval, so a 10-page PDF costs one request per poll, not ten. |
| Fetching final details | **Parallel** — `Promise.all` over `handleSyncJob`. |

Only rasterization is serial; nothing blocks page-by-page.

Sync mode attaches `{ resolve, reject }` per page (mirroring the `/percy/comparison` route), then aggregates each page's `handleSyncJob` result. `handleSyncJob` converts failures into `{ error }` rather than rejecting, so one bad page yields a partial result instead of losing every other page's.

### `@percy/sdk-utils` — the shared seam

`postPdfSnapshot()`, exported alongside `postSnapshot`, so all SDKs can re-export it.

### Pages are extracted, not rendered

percy-api can skip the renderer entirely for image-backed snapshots. `Comparison#upload_snapshot?` (`app/models/percy/comparison.rb`) gates on exactly two things:

```ruby
unless user_agent&.include?('@percy/cli-upload')   # substring match
unless resource_url&.include?('http://local/')     # the ROOT resource URL
```

When both hold, `start_comparison_job.rb` calls `extract_and_process_upload_snapshot`, which recovers the image by matching the root resource against `/` **and** `@percy/cli-upload/<version>`. The gate is a substring match, so naming both unlocks extraction while keeping the UA honest about which code actually ran, rather than impersonating the upload command.

### One wrapper, shared with cli-upload

That wrapper HTML is a **contract with percy-api**, not cosmetics — and a mismatch is not an error: extraction raises, percy-api rescues it, and the snapshot silently falls back to being rendered at ~10x the processing cost, with nothing surfaced to the user.

It therefore has exactly one definition, `buildImageSnapshotHtml` / `createImageSnapshotResources` in `@percy/core`'s `utils.js`. `cli-upload`'s `getImageResources` delegates to it, and the PDF path uses it. core cannot import `cli-upload` (`cli-upload` -> `cli-command` -> `core` would cycle), but `cli-upload` already reaches core's utils through `@percy/cli-command/utils`, so this needs no new dependency. `cli-upload`'s existing suite passes unchanged, confirming byte-identical output.

`core/test/image-snapshot-resources.test.js` pins percy-api's regex verbatim so drift fails CI instead of degrading silently in production.

**Operator note:** `upload_extraction_allowed?` only short-circuits on a project's default base branch. Elsewhere it mirrors the base comparison's `upload_snapshot_extracted` flag, so baselines taken before this change must be regenerated before comparison builds will extract.

## Two contract decisions worth calling out

**The document travels as base64 in an ordinary JSON body**, and **the sync response is always a JSON object, never a bare array.** Both exist so every SDK can call this with the HTTP client it already has. Concretely, for the .NET wrapper: the request goes through its existing `Dictionary<string, object>` → `JsonSerializer` helper, and the response through its existing `JObject.Parse` — no multipart, no streaming, no new HTTP machinery. The cost is base64's 33% inflation, which is the right trade for a one-line integration across ~26 SDKs.

<details>
<summary>Sync response shape</summary>

```json
{
  "success": true,
  "data": {
    "pdf-name": "Burglary Insurance Policy",
    "page-count": 3,
    "pages-snapshotted": 3,
    "status": "success",
    "pages": [
      { "page": 1, "snapshot-name": "Burglary Insurance Policy | Page 1",
        "screenshots": [ { "diff-info": { "diff-ratio": 0 } } ] },
      { "page": 2, "snapshot-name": "Burglary Insurance Policy | Page 2",
        "screenshots": [ { "diff-info": { "diff-ratio": 0.00085774 } } ] }
    ]
  }
}
```

`page` and `snapshot-name` are set **after** the API payload is spread in, so the values we submitted always win over whatever the API echoes back — otherwise a change in the API's naming would silently break the caller's page mapping.
</details>

## Notes for reviewers

- **Snapshot names match `percy-pdf` exactly** (`<name> | Page N`) so teams migrating keep their approved baselines instead of orphaning every one.
- **No `percy pdf <dir>` command, deliberately.** `percy.syncMode()` force-disables sync under `skipUploads`/`deferUploads`/`delayUploads`, which is exactly what the `snapshot` and `upload` commands set. A command could therefore never return comparison results, so PDF support is reachable only through this endpoint under `percy exec`.
- **Oversized pages are fitted, not rejected.** Legal (1224×2016 at scale 2) and A3 exceed Percy's 2000 px cap and are precisely the documents this targets, so `fitScale()` reduces the scale deterministically from the page's own dimensions and warns. Deterministic from page geometry means the same document always rasterizes identically, which is what a stable baseline needs.
- **`pdfjs-dist` pinned to 4.x, not 6.x** — 6.x requires Node >= 22.13, while 4.8.69 needs only Node >= 18, matching the CLI's engine range.
- **`standardFontDataUrl` and `cMapUrl` are configured explicitly.** Without them pdf.js renders standard fonts with wrong metrics or drops CJK glyphs and only *warns*, which would surface as a mysterious visual diff rather than an error.
- **`isEvalSupported: false`** — PDFs are untrusted input arriving over the local API, and this is the one pdf.js switch that permits code execution.
- **`standardFontDataUrl` / `cMapUrl` are served, not configured away.** Without them pdf.js renders standard fonts with wrong metrics or drops CJK glyphs, and only *warns* — which would surface as a mysterious visual diff rather than an error.
- **`new Function(pdfjsSource)`** mirrors `page.insertPercyDom()` (`page.js:239`); the input is pdfjs-dist's own vendored file, never user data.
- **Baselines from the earlier revision of this PR must be regenerated**, since the rasterizer changed from Skia to Chromium.

## A bug caught during development

The first end-to-end run reported `max diff-ratio=0` for a document that genuinely differed — **both the passing and failing functional tests passed vacuously.**

The wrapper (then a PDF-local `buildPageHtml`, since replaced by the shared one) ran `encodeURI` over an already percent-encoded image URL, turning `%20` into `%2520`. The `` then matched no registered resource, so every page rendered as the same blank sheet and all pages collapsed to a single image hash. It surfaced only by noticing that page 2 and page 3 shared a `current-image` hash in the raw sync payload.

Fixed to HTML-escape only. The guards now live in `packages/core/test/image-snapshot-resources.test.js` ("does not re-encode an already-encoded URL", "keeps the img src and the image resource URL identical") and `packages/core/test/pdf-snapshot.test.js` ("attaches a root DOM whose img src matches the image resource", "gives each page a distinct image resource"). This is a failure mode that reports green, so it is worth guarding directly.

## Node compatibility

`pdfjs-dist` is pinned to **2.16.105**, the last line that declares no `engines` constraint. 3.x and later declare `node: ">=18"`, and since yarn enforces `engines` across the whole tree, anything newer breaks `yarn install` on the repo's Node 14 CI — not just for PDF users, but for everyone. `@percy/cli-pdf` therefore declares `>=14`, matching its sibling packages.

## Testing

| Suite | Result |
|---|---|
| `@percy/cli-pdf` | 32/32 pass |
| `@percy/core` (`test/pdf-snapshot.test.js`, 19 specs) | 19/19 pass |
| `@percy/sdk-utils` (`postPdfSnapshot`, 5 specs) | 5/5 pass |
| `eslint` on all changed files | clean |

Verified end to end against a real Percy project using a Selenium + Mocha harness that drives a browser, downloads a PDF via a download button, and hands the bytes to this endpoint:

- unchanged document → all 3 pages `diff-ratio 0`, **exit 0**
- document changed on page 2 only → **exit 1**, `page 2 … diff-ratio 0.00085774`, while **pages 1 and 3 report 0**

That last line is the point of per-page snapshots: a one-page change in a multi-page document is localised rather than flagging the whole file.

## Follow-ups

- SDK wrappers exposing `postPdfSnapshot`. A `Percy.PdfSnapshot(string name, byte[] pdf, Dictionary<string, object>? options)` implementation for `percy-selenium-dotnet` is written against this contract and reuses its existing `Request()` helper unchanged; it will be raised separately.
- Customer-facing docs, and a deprecation notice on `percy-pdf` pointing here.

## Summary by CodeRabbit

- **New Features**
  - Added PDF snapshot support, converting PDF pages into individual visual snapshots.
  - Added page selection and exclusion options, including page ranges.
  - Added configurable rasterization scale with automatic sizing limits.
  - Added SDK support for submitting PDF snapshots with optional request parameters.
  - Added validation for PDF content, page selections, snapshot names, and rendering options.
  - Added configurable server host settings.
- **Bug Fixes**
  - Improved error messages for browser and rendering failures.
  - Improved image snapshot URL handling and resource matching.
- **Regression Testing**
  - Added cross-platform PDF rendering comparisons for consistent results.


RaghavsBrowserStack:
# Claude Code PR Review

**PR:** https://github.com/percy/cli/pull/2418  •  **Head:** 1ad51f7  •  **Reviewers:** stack:code-reviewer

## Summary
Adds native PDF visual testing: a new `@percy/cli-pdf` package (pdf.js assets + page-selection parsing + page-context scripts), `POST /percy/pdf/snapshot` in `@percy/core` backed by `pdf-rasterize.js` rendering in the already-idle discovery browser, a `postPdfSnapshot` SDK helper, and a shared image-snapshot resource/HTML builder extracted for reuse by `@percy/cli-upload`.

## Review Table

| High | Security | No hardcoded secrets or credentials | Pass | No credentials, tokens, or endpoints introduced by the diff. |
| High | Security | Authentication/authorization checks present | N/A | Local loopback CLI API; consistent with the existing unauthenticated `/percy/*` surface. |
| High | Security | Input validation and sanitization | Pass | `decodePdf` enforces base64 decode, `%PDF-` magic bytes and a 50 MB cap (matching `/percy/comparison/upload`); `buildImageSnapshotHtml` escapes `name`/`imageUrl`; pdf.js is injected with `isEvalSupported: false`. |
| High | Security | No IDOR — resource ownership validated | N/A | No multi-tenant resource lookup in this change. |
| High | Security | No SQL injection (parameterized queries) | N/A | No SQL in this codebase path. |
| High | Correctness | Logic is correct, handles edge cases | Pass | `resolvePages` covers single pages, arrays, closed/open ranges and exclusions, validated against the real page count. One schema/parser drift (Finding 1) is warn-only and does not change behaviour. |
| High | Correctness | Error handling is explicit, no swallowed exceptions | Pass | `rasterizePdf` uses `try/finally` to close both the isolated page and the ephemeral asset server on every exit path; request errors surface as `ServerError(400)`. One misleading message (Finding 4). |
| High | Correctness | No race conditions or concurrency issues | Pass | `percy.browser.launch()` is idempotent (`if (this.readyState != null) return`); pages render sequentially; the sync path uses the `generatePromise`-wrapped `percy.upload` with the existing `{resolve, reject}` job pattern. |
| Medium | Testing | New code has corresponding tests | Pass | `core/test/pdf-snapshot.test.js` (439 lines), plus `cli-pdf` tests for `pages`, `browser-scripts` and `assets`, `image-snapshot-resources.test.js`, and `sdk-utils` coverage. |
| Medium | Testing | Error paths and edge cases tested | Pass | Dedicated commits (`12cce155`, `858d617f`, `1cd17634`) close the error-path and page-context coverage gaps to the 100% threshold. |
| Medium | Testing | Existing tests still pass (no regressions) | Pass | All checks green except `Test @percy/cli-exec`, confirmed by the requester as unrelated to this PR and excluded from this review. One `Test @percy/core` job was still pending at review time. |
| Medium | Performance | No N+1 queries or unbounded data fetching | Pass | 50 MB document cap; `scale` capped at 5 and auto-reduced when a page would exceed 2000px. |
| Medium | Performance | Long-running tasks use background jobs | Pass | Reuses the discovery browser that is otherwise idle during a PDF run rather than launching a second one; async is the default, `sync` is opt-in. |
| Medium | Quality | Follows existing codebase patterns | Pass | Mirrors `Page.insertPercyDom()` for browser injection, `maestro-screenshot.js` for the sync-job pattern, and preserves `cli-upload`'s exact resource/HTML shape through the extraction. |
| Medium | Quality | Changes are focused (single concern) | Pass | Single feature; the `.github/workflows/test.yml` edit is the one line adding `@percy/cli-pdf` to the test matrix, required by the new package. |
| Low | Quality | Meaningful names, no dead code | Pass | `@napi-rs/canvas` and its ~25 MB of native prebuilds were removed in `45ceaf2a` rather than left behind. |
| Low | Quality | Comments explain why, not what | Pass | Non-obvious choices (serving assets over a real origin so pdf.js can fetch `standardFontDataUrl`/`cMapUrl`) are explained. |
| Low | Quality | No unnecessary dependencies added | Pass | `pdfjs-dist` pinned to 2.x and carried as an `optionalDependency` so its ~34 MB is only installed by users who snapshot PDFs; no `@percy/core` reference from `cli-pdf`, avoiding a cycle. |

## Findings

- **File:** `packages/core/src/config.js:1062` (and `:1071` for `excludePages`)
- **Severity:** Medium
- **Issue:** The JSON-schema `pattern` for `pages`/`excludePages` is stricter than the parser it describes. `parseSelection` in `packages/cli-pdf/src/pages.js` deliberately skips empty segments, and `packages/cli-pdf/test/pages.test.js:34-35` pins that behaviour (`'1,,3'` → `[1,3]`, `'2,'` → `[2]`). Verified against the live regex: both of those inputs fail to match, while `'1,3-5,8'`, `'3-'` and `'2-4'` pass. `validatePdfSnapshotOptions` only logs (it does not reject), so this is not a functional break — but it emits a misleading `Invalid PDF snapshot options: - pages: ...` warning for inputs the package's own tests treat as valid, which erodes the warning's signal.
- **Suggestion:** Allow empty segments in the pattern, e.g. `'^\s*(\d+\s*(-\s*\d*\s*)?)?(\s*,\s*(\d+\s*(-\s*\d*\s*)?)?)*$'`, or drop the regex and rely on `resolvePages`, which already raises specific runtime errors.

- **File:** `packages/core/src/config.js` (`scale`) vs `packages/cli-pdf/src/browser-scripts.js` (`DEFAULT_SCALE`, `MAX_SCALE`)
- **Severity:** Low
- **Issue:** `maximum: 5` / `default: 2` are hardcoded literals in the schema while `pdf-rasterize.js` enforces the exported `MAX_SCALE`/`DEFAULT_SCALE` constants. Two sources of truth that can silently diverge. The optional-dependency boundary is presumably why `config.js` can't import them.
- **Suggestion:** Add a comment in `config.js` naming `@percy/cli-pdf`'s constants as canonical, or add a test asserting the two stay in sync.

- **File:** `packages/core/src/pdf-snapshot.js:103-104`
- **Severity:** Low
- **Issue:** `widths: snapshotOptions.widths || [width]` and `minHeight: snapshotOptions.minHeight || height` discard explicit falsy overrides (`widths: []`, `minHeight: 0`) in favour of the derived default.
- **Suggestion:** Use `??` or an explicit key check if "unset" and "falsy" should be distinguished. Low impact — neither falsy value is meaningful input.

- **File:** `packages/core/src/pdf-snapshot.js:53-54`
- **Severity:** Low
- **Issue:** The short-buffer branch throws ``` `pdf.content` is not valid base64-encoded data ```, but it fires whenever the *decoded* buffer is under 5 bytes — which valid base64 of a tiny payload will also do. It conflates "too short to be a PDF" with "not valid base64".
- **Suggestion:** Reuse the adjacent, more accurate wording ("does not decode to a PDF (missing `%PDF-` header)"), since both branches mean the same thing to the caller.

### Verified non-issues
The reviewer explicitly checked and cleared: `percy.upload(...).catch(reject)` (valid — `upload` is `generatePromise`-wrapped in the `Percy` constructor); `new Function(pdfjsSource)` injection (identical to the existing `insertPercyDom()` pattern); `widths`/`minHeight`/`sync` all declared in the shared `/snapshot#/$defs/common` schema; asset-server and page cleanup on every exit path; and `cli-upload`'s refactor preserving the exact prior HTML/resource shape.

> Note: the security reviewers in this repo (`stack:security-review`, `stack:security-auditor`) are disabled for this orchestrator run; the Security rows above reflect the general reviewer's incidental observations, not a dedicated security pass.

---

**Verdict: PASS**

packages/core/src/pdf-rasterize.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  let assets = pdfjsAssets();
+  let server = await createAssetServer(pdfBuffer, assets);
+  let origin = server.address();
+  let page;
+
+  try {
+    await percy.browser.launch();
+    page = await percy.browser.page({ meta: { snapshot: { name: 'pdf' } } });
```
rishigupta1599, line 80: do we need browser launch, was assuming page would be converted to image and directly upload

packages/core/src/utils.js
```diff
@@ -488,6 +488,67 @@ export function createRootResource(url, content, attrs = {}) {
...
+// which recovers the image by matching this HTML against
+//
+//   /<img\s+src="([^"]+)"\s+width="(\d+)px"\s+height="(\d+)px"/
+//
+// and skips the renderer entirely -- measured at ~1s per page versus ~9-19s
+// rendered. A mismatch is not an error: extraction raises, percy-api rescues,
+// and the snapshot silently falls back to being rendered. So the attribute
+// order (src, width, height), the `px` suffixes and the integer dimensions all
+// matter, and there must only ever be one definition of them.
+export function buildImageSnapshotHtml({ name, imageUrl, width, height }) {
```
rishigupta1599, line 542: why do we need this? is it inline with upload?

packages/core/src/pdf-rasterize.js
RaghavsBrowserStack: it is possible and for some reason that is the approach claude went with initially, there's this library called @napi-rs/canvas which can simulate a canvas without a browser but it works by downloading a system compatible rendering framework, other direct libraries also exist like pdf-to-image that can work directly, but it just doesn't make sense to add a massive library when we already have the tools right here.

packages/core/src/utils.js
RaghavsBrowserStack: yes, start_comparison_job.rb:360, there's an image regexp that needs to be met for the non rendering flow. This html is meets that criteria

RaghavsBrowserStack:
# Claude Code PR Review

**PR:** https://github.com/percy/cli/pull/2418  •  **Head:** 1e61fd4  •  **Reviewers:** stack:code-reviewer

> Diff taken against `origin/master` (44 files, ~2.3k insertions). The local `master` ref was 9 commits stale; reviewing against it would have folded in unrelated master-side changes.

## Summary
Adds native PDF visual testing: a new `@percy/cli-pdf` package (pdf.js assets, page-selection parsing, in-page render scripts), a `POST /percy/pdf/snapshot` endpoint in core that rasterizes a base64 PDF in the existing discovery browser and emits one Percy snapshot per page, a shared image-snapshot wrapper extracted out of `@percy/cli-upload`, a `postPdfSnapshot` sdk-utils helper, and a byte-exact PNG golden regression track.

## Review Table

| High | Security | Authentication/authorization checks present | Fail | The per-request PDF asset server inherits `Server`'s `::` bind default (`server.js:137`), serving the customer's document bytes unauthenticated on **all interfaces** for the duration of rasterization. Only the local discovery browser needs to reach it. See #5. |
| High | Security | Input validation and sanitization | Fail | Magic-byte, size, name, scale and page-selection validation are thorough — but the wrapper's HTML escaping corrupts the URL contract (#1) and the 50MB cap is applied only after full decode (#4). |
| High | Security | No IDOR — resource ownership validated | N/A | No multi-tenant resource addressing on this path. |
| High | Security | No SQL injection (parameterized queries) | N/A | No SQL in the CLI. |
| High | Correctness | Logic is correct, handles edge cases | Fail | Page selection, ranges, dedup, exclusion and scale clamping are excellent and exhaustively tested. #1 and #2 are defects on live paths. |
| High | Correctness | Error handling is explicit, no swallowed exceptions | Fail | Nothing is swallowed, but #2 discards the message entirely on the most likely production failure, and #6 reports every server-side failure as HTTP 400. |
| High | Correctness | No race conditions or concurrency issues | Pass | Per-request asset server and browser page are properly isolated; sync callbacks mirror the established `/percy/comparison` pattern. Caveat: no per-endpoint concurrency limit (#3). |
| Medium | Testing | New code has corresponding tests | Pass | `pages`, `browser-scripts`, `assets`, the endpoint, the wrapper contract and sdk-utils are all covered — including percy-api's extraction regex pinned verbatim as an assertion, which is the right instinct for a silently-degrading contract. |
| Medium | Testing | Error paths and edge cases tested | Fail | Validation rejections and sync failures are well covered; the in-page rasterization failure path (#2) and the apostrophe URL case (#1) are both untested, which is why CI is green. |
| Medium | Testing | Existing tests still pass (no regressions) | Fail | CI is green (49 pass, 2 pending, 0 fail) — but `@percy/cli-upload` genuinely regresses for filenames containing `'` (#1) and no test guards it. Green CI here reflects a coverage gap, not absence of regression. |
| Medium | Performance | No N+1 queries or unbounded data fetching | Fail | No cap on page count; every page PNG is retained in memory twice (render array + resource closures), each crossing CDP as a base64 data URL first. Request bodies are buffered with no limit (#3, #4). |
| Medium | Performance | Long-running tasks use background jobs | Fail | `page.eval(renderPage, …)` has no timeout — `Page.TIMEOUT` covers `goto` only (`page.js:187`). A PDF that wedges pdf.js hangs the HTTP request while holding a browser page and a listening asset server indefinitely (#3). |
| Medium | Quality | Follows existing codebase patterns | Pass | Route registration, `ServerError`, `handleSyncJob`, `percy.upload` + lazy `resources` thunk, sdk-utils helper shape, regression-track runner and `.semgrepignore` rationale format all match the repo. |
| Medium | Quality | Changes are focused (single concern) | Pass | The `cli-upload` refactor is genuinely required to keep one definition of the percy-api wrapper contract; the rest is on-topic. |
| Low | Quality | Meaningful names, no dead code | Pass | Nit: `_parseSelection` is exported from `pages.js:74` but imported nowhere. |
| Low | Quality | Comments explain why, not what | Pass | Among the strongest in the repo — the percy-api contract, the `isEvalSupported: false` rationale, the golden-platform scoping and the tolerance measurement are all "why". |
| Low | Quality | No unnecessary dependencies added | Pass | `pdfjs-dist` is necessary and `@napi-rs/canvas` was correctly dropped in favour of the browser already running. |

## Findings

- **File:** `packages/core/src/utils.js:533`
- **Severity:** High
- **Issue:** `escapeHtml` is applied to `imageUrl` in attribute context, but `encodeURIComponent` does **not** escape `'`. The apostrophe becomes `&#39;` in the `` while `createResource` registers the raw URL (`normalizeURL` preserves `'`), so the two diverge. Verified by direct reproduction:
  ```
  resource url : http://local/Jack's%20Resume/page-1.png
  extracted src: http://local/Jack&#39;s%20Resume/page-1.png   -> no match
  ```
  Per the file's own contract comment, a mismatch is **not** an error: percy-api rescues and silently falls back to rendering, turning the ~1s/page extraction path back into ~9-19s/page with nothing surfaced to the user. This is also a **behavioural regression in shipping code** — the previous `cli-upload/src/utils.js` interpolated `imageUrl` raw, so apostrophed filenames extracted correctly before this PR and will not after it.
- **Suggestion:** Split text-context from attribute-context escaping. The value is double-quoted, so only `&` and `"` can break out; `'` must be left literal so the src stays byte-identical to the registered resource URL.
  ```js
  // Attribute context: double-quoted, so only & and " can break out. `'` must NOT
  // be escaped -- percy-api compares this src against the registered resource URL
  // byte-for-byte, and encodeURIComponent leaves `'` literal.
  function escapeAttr(value) {
    return String(value).replace(/&/g, '&amp;').replace(/"/g, '&quot;');
  }
  ```
  Keep `escapeHtml` for `<title>${escapeHtml(name)}</title>`, use `escapeAttr` for the `src`. Add a test beside the existing `%20` one asserting `regex.exec(root.content)[1] === image.url` for a name containing `'`.

---

- **File:** `packages/core/src/pdf-snapshot.js:160`
- **Severity:** High
- **Issue:** `Page#eval` rejects with a **string**, not an `Error` — confirmed at `packages/core/src/page.js:216` (`throw exceptionDetails.exception.description`). Every failure originating inside the page (corrupt PDF, password-protected PDF, a pdf.js render throw, the `'pdf.js did not initialise in the page'` guard) therefore has no `.message`, and the caller receives literally `Could not rasterize PDF: undefined`. A malformed-but-`%PDF-`-prefixed document is the single most likely production failure. The existing test passes only because it stubs `percy.browser.page` rejecting with a real `Error`.
- **Suggestion:**
  ```js
  } catch (error) {
    // Page#eval rejects with the CDP exception description string, not an Error.
    let message = error?.message ?? String(error);
    log.error(`Failed to rasterize PDF "${name}": ${message}`);
    throw new ServerError(status, `Could not rasterize PDF: ${message}`);
  }
  ```
  Add a test posting a `%PDF-`-prefixed but structurally invalid buffer, asserting the message is non-empty.

---

- **File:** `packages/core/src/pdf-rasterize.js:60`
- **Severity:** High
- **Issue:** Three unbounded dimensions on a network-reachable endpoint. (a) Nothing caps page count — a 50MB PDF can hold thousands of pages, and `pages.push({ … png })` retains every PNG, which `queuePages` then holds again in resource closures; each also crosses CDP as a base64 data URL (+33%) before decode. (b) `page.eval(renderPage, …)` has **no timeout** — `Page.TIMEOUT` is applied only to `goto` (`page.js:187`), so `Runtime.callFunctionOn` with `awaitPromise: true` waits forever, hanging the request while holding a browser page and a listening asset server. (c) Nothing limits concurrent requests, so N in-flight PDF snapshots mean N asset servers and N browser pages.
- **Suggestion:**
  ```js
  const MAX_PAGES = 250;
  const PAGE_RENDER_TIMEOUT = 30_000;

  if (selected.length > MAX_PAGES) {
    throw new Error(`Requested ${selected.length} pages; the maximum is ${MAX_PAGES}. Use \`pages\` to narrow the selection.`);
  }

  let rendered = await Promise.race([
    page.eval(renderPage, { pageNumber, scale: effectiveScale }),
    new Promise((_, r) => setTimeout(() => r(new Error(
      `Timed out rendering page ${pageNumber} after ${PAGE_RENDER_TIMEOUT}ms`)), PAGE_RENDER_TIMEOUT).unref())
  ]);
  ```

---

- **File:** `packages/core/src/pdf-snapshot.js:51`
- **Severity:** Medium
- **Issue:** `Buffer.from(content, 'base64')` allocates before the size check, and `IncomingMessage` (`packages/core/src/server.js:39`) buffers the entire request body with no cap and JSON-parses it. A 1GB body is fully buffered, parsed, and decoded into a ~750MB Buffer before the 413 fires.
- **Suggestion:** Reject on the encoded length before allocating: `if (content.length > Math.ceil(MAX_PDF_BYTES / 3) * 4) throw new ServerError(413, …)`. A `Content-Length` guard in the route, and eventually a server-level body cap, would be the fuller fix.

---

- **File:** `packages/core/src/pdf-rasterize.js:6`
- **Severity:** Medium
- **Issue:** `Server.createServer({ port: 0 })` inherits `get host() { return process.env.PERCY_SERVER_HOST || '::' }` (`server.js:137`), so the customer's PDF — and confidential documents like policies and resumes are precisely this feature's target — is served unauthenticated on all interfaces while rasterization runs. Unlike the API server, nothing here needs off-loopback reachability; the only client is the local discovery browser.
- **Suggestion:** Thread a `host` option through `Server`/`createServer`, pass `'127.0.0.1'` here, and build `origin` from that host.

---

- **File:** `packages/core/src/pdf-snapshot.js:161`
- **Severity:** Medium
- **Issue:** Browser-launch failure, OOM, an asset-server bind error and a CDP disconnect all return `400 Bad Request`, telling the SDK the caller's input was wrong when it was not.
- **Suggestion:** Tag input-derived throws (page selection out of range, bad scale, unparseable PDF) with `status = 400` and default everything else to 500.

---

- **File:** `packages/core/src/pdf-rasterize.js:90`
- **Severity:** Medium
- **Issue:** In the `finally` block, `await page?.close()` precedes `await server.close()`, so a throwing `close()` leaks the asset server — a listening socket still holding the PDF buffer.
- **Suggestion:** `await Promise.allSettled([page?.close(), server.close()]);`. Moving `destroyDocument` (line 87) into the cleanup would also be tidier, though it is harmless today since the page is destroyed anyway.

---

- **File:** `test/regression/pdf-render.test.js:70`
- **Severity:** Medium
- **Issue:** Only the linux-x64 job asserts, so the committed `darwin-arm64` goldens (~2MB of PNGs) are verified by nothing in CI and will silently rot on the next Chromium bump — the first macOS contributor to run the track gets a failure unrelated to their change. Separately, any platform without a committed set (linux-arm64, win32, a future GitHub arm64 runner) **exits 1**, turning an environment mismatch into a red build. The assertion is also byte-exact against goldens produced on `ubuntu-latest`, a moving image: a freetype/harfbuzz/fontconfig bump changes glyph rasterization with no code change.
- **Suggestion:** Skip-with-warning on an unrecognised platform unless `CI`/`PDF_GOLDENS_REQUIRED=1` is set; either drop the darwin-arm64 set or document it as best-effort and unenforced; pin the job to `ubuntu-22.04` rather than `ubuntu-latest`. The tolerance mechanism for `jack-sparrow-resume` is well-reasoned and correctly refuses to tolerate size changes.

---

- **File:** `packages/core/test/pdf-snapshot.test.js:56`
- **Severity:** Medium
- **Issue:** `jasmine.DEFAULT_TIMEOUT_INTERVAL = 240000` plus real Chromium rasterization across ~25 specs turns part of core's **unit** job into an integration suite (the `filesystem.$bypass` hook for `pdfjs-dist` is a symptom of the same). The `afterAll` restore is correct and the comment is honest, but `Test @percy/core` is the one check still pending on this PR.
- **Suggestion:** Put the fan-out/sync specs behind a stubbed `rasterizePdf` and let `test/regression/pdf-render.test.js` own the real-browser assertions; the request-validation and `decodePdf` specs need no browser at all.

---

- **File:** `packages/core/src/utils.js:1001`
- **Severity:** Low
- **Issue:** `OPTION_MAPPINGS` has no entry for `excludepages`, and `normalizeOptions` falls back to the *original* key. A snake_case SDK caller sending `exclude_pages` therefore keeps that key, trips `unevaluatedProperties: false` into a log-only warning, and the handler's `excludePages` destructure yields `undefined` — the option is silently dropped. (`pages` and `scale` are unaffected: single lowercase words pass through intact.)
- **Suggestion:** Add `excludepages: 'excludePages'` to `OPTION_MAPPINGS`.

---

- **File:** `packages/core/src/config.js:1082`
- **Severity:** Low
- **Issue:** `maximum: 5` / `default: 2` duplicate `MAX_SCALE` / `DEFAULT_SCALE` (`cli-pdf/src/browser-scripts.js:4-5`), and `2000px` appears in the schema description, in `MAX_DIMENSION`, and in the warn string at `pdf-rasterize.js:70`. The schema `default` never applies, since the handler reads `scale` off `normalized` rather than the validated copy. Duplication is arguably forced here — `@percy/cli-pdf` is an optional dependency, so core cannot import it at schema-registration time — but nothing pins the two together.
- **Suggestion:** Add a test asserting the schema bounds match the `cli-pdf` constants, or note the constraint in a comment next to `maximum: 5`.

---

- **File:** `packages/core/src/pdf-snapshot.js:26`
- **Severity:** Low
- **Issue:** Declaring `@percy/cli-upload/<version>` to satisfy percy-api's substring check is a deliberate, well-commented hack, and pairing it with an honest `@percy/cli-pdf/…` entry is the right mitigation — but client-info telemetry will over-count the upload command for as long as it stands.
- **Suggestion:** File a linked ticket to add `@percy/cli-pdf` to percy-api's `upload_snapshot?` predicate, then drop the synthetic entry.

---

- **File:** `packages/cli-pdf/src/pages.js:74`
- **Severity:** Low
- **Issue:** Assorted polish. `_parseSelection` is exported but imported nowhere. `packages/cli-pdf/` has no `README.md` while every other `cli-*` package does, and it registers no CLI command despite the `cli-` prefix — `@percy/pdf` would describe it better, and renaming is far cheaper before the beta ships. `pdf-snapshot.js:151` awaits `loadPdfModule()` but discards the result, and `rasterizePdf` re-imports the same module. `sdk-utils/src/post-pdf-snapshot.js` lacks the explanatory comment its two siblings carry. The `.semgrepignore` entries are *path* exclusions removing the two files from **all** rules, not just `path-join-resolve-traversal` — acceptable for test-only files and consistent with repo convention, but the comment slightly overstates what is suppressed. Finally, the PR description contradicts itself on the pdfjs pin (4.8.69 in one section, 2.16.105 in another); the tree has `^2.16.105`, a caret range rather than the pin the description claims.
- **Suggestion:** Drop `_parseSelection` or add the test it was exported for; add a README; pass the loaded module into `rasterizePdf`; reconcile the PR description before merge.

---

> Note: `pdfjs-dist@2.x` is EOL, pinned for the Node 14 floor. `isEvalSupported: false` is the correct mitigation for CVE-2024-4367 and is asserted in a test — worth a tracking ticket to move to 4.x when the engine floor rises.

> CI at time of review: 49 checks passing, 2 pending (`Test @percy/core`), 0 failing.

> Security-specialist reviewers (`stack:security-review`, `stack:security-auditor`) are temporarily disabled in the orchestrator, so the Security rows above reflect only what the general code reviewer surfaced — they are not a substitute for a dedicated security pass.

---

**Verdict: FAIL**

RaghavsBrowserStack:
# Claude Code PR Review

**PR:** https://github.com/percy/cli/pull/2418  •  **Head:** 750a5f0  •  **Reviewers:** stack:code-reviewer

> Re-review of the fix commit `750a5f06`, which addresses 6 of the 13 findings from the review of `1e61fd4`. Diff taken against `origin/master` (local `master` is 9 commits stale).

## Summary
Adds native PDF visual testing (`POST /percy/pdf/snapshot`, new `@percy/cli-pdf`, in-browser rasterization, a shared image-snapshot wrapper); the new commit fixes the extraction-contract escape bug, makes `Page#eval` throw a real Error, adds page/time/size limits and corrects HTTP status classification.

## Review Table

| High | Security | No hardcoded secrets or credentials | Pass | None in the diff. |
| High | Security | Authentication/authorization checks present | Fail | Prior finding 5 unaddressed: `createAssetServer` (`pdf-rasterize.js:6`) inherits `Server`'s `'::'` default host (`server.js:137`), serving the customer's PDF unauthenticated on all interfaces for the life of the request. Only the local browser needs it. |
| High | Security | Input validation and sanitization | Pass | Scale, page selection, magic bytes and size (now pre- and post-decode) all validated; the `escapeHtml`/`escapeAttr` split is contextually correct. Edges in F10/F11. |
| High | Security | No IDOR — resource ownership validated | N/A | No user-scoped identifiers on this path. |
| High | Security | No SQL injection (parameterized queries) | N/A | No SQL in the CLI. |
| High | Correctness | Logic is correct, handles edge cases | Fail | F7 is not cosmetic — it shipped as a red `Test @percy/core`. Base64 arithmetic is exact, `MAX_PAGES` ordering is right, `page?.close()` under `allSettled` is right; F2/F3/F4 remain. |
| High | Correctness | Error handling is explicit, no swallowed exceptions | Fail | F6: `Promise.allSettled` now discards both cleanup rejections with no log. F13: a redundant `.catch()` whose comment describes a hazard that cannot occur. |
| High | Correctness | No race conditions or concurrency issues | Pass | `Promise.race` + `.finally(clearTimeout)` is sound on all paths; `timer.unref?.()` correct in Node, inert elsewhere; a timed-out page is closed, so the shared browser is not left holding a wedged target. |
| Medium | Testing | New code has corresponding tests | Fail | F1: `withTimeout` — the entire new timeout mechanism — has no test, and `@percy/core` enforces 100% branches/lines/functions/statements (`.nycrc`) via `test:coverage` in CI. `remoteError`, `MAX_PAGES`, `escapeAttr`, `asInputError` and the pre-decode cap are all well covered. |
| Medium | Testing | Error paths and edge cases tested | Fail | Good new coverage of 400-vs-500, non-Error throws, the undefined-message regression, cap-at-boundary and apostrophe/angle-bracket URLs. Untested: any timeout, cleanup failure, `&` in a URL. |
| Medium | Testing | Existing tests still pass (no regressions) | Fail | `Test @percy/core` is red on `snapshot.test.js:1754` "logs execute errors and does not snapshot" — deterministic across all 5 CI retries. See F7. |
| Medium | Performance | No N+1 queries or unbounded data fetching | Fail | F4: `MAX_PAGES` caps page *count*, not bytes — 250 pages at the 2000px ceiling is ~0.5-1.5GB resident, which is the failure mode the cap was meant to prevent. F12: the request body is still buffered and JSON-parsed with no cap. |
| Medium | Performance | Long-running tasks use background jobs | Fail | F2: two `page.eval` calls are still un-raced. F3: no aggregate deadline — 250 pages × 30s is ~2h on one held-open request without tripping a single timeout. |
| Medium | Quality | Follows existing codebase patterns | Pass | `asInputError` composes correctly with `server.js`'s existing `let { status = 500 } = error`; no `.status` collisions anywhere in `core`/`client`/`install` (audited). |
| Medium | Quality | Changes are focused (single concern) | Pass | Every hunk maps to a named prior finding. |
| Low | Quality | Meaningful names, no dead code | Fail | F13 (`promise.catch(() => {})`) plus the still-present `_parseSelection` export (`pages.js:86`). |
| Low | Quality | Comments explain why, not what | Fail | Unusually good overall, but two are inaccurate: F12's overstates what the pre-decode check closes, and F13's describes an unhandled-rejection hazard that cannot occur. |

## Findings

- **File:** `packages/core/src/page.js:97`
- **Severity:** High
- **Reviewer:** stack:code-reviewer (F7) — confirmed by CI
- **Issue:** `remoteError` takes the description's first line as the `message` of a `new Error`, whose `name` is `"Error"`. CDP's description already opens with the remote error's own name, so the prefix doubles. @percy/logger renders a thrown Error as `Error.prototype.toString.call(err)` and only falls back to `stack` at debug level (`logger.js:301-312`), so this also **drops the remote frames** from what the user sees. A thrown *string* was previously printed verbatim. `Test @percy/core` fails deterministically on `snapshot.test.js:1754`, which pins exactly those `at execute (<anonymous>:4:17)` frames — the lines a user needs to debug their own execute script.
- **Suggestion:** Blank the name so `Error.prototype.toString` returns the description verbatim, keeping logged output byte-identical to the string it replaced, and take `message.split('\n')[0]` where only a summary line is wanted (the HTTP body). Note the fix of parsing the name off the first line does *not* work: it yields `"Error: test error"` with the frames stripped, which still fails this spec.

---

- **File:** `packages/core/src/pdf-rasterize.js:34`
- **Severity:** High
- **Reviewer:** stack:code-reviewer (F1)
- **Issue:** `withTimeout` has no test. The `setTimeout(() => reject(...))` arrow, the `promise.catch(() => {})` arrow and the `timer.unref?.()` branch are never executed by any spec — the three fake-page specs reject on the *first* `page.eval`, which is the pdf.js injection at line 82 and is **not** wrapped. `.nycrc` requires 100% branches/lines/functions/statements and CI runs `test:coverage`. Beyond the gate: the fix for "no timeout on rendering" ships with no evidence it works.
- **Suggestion:** Export `withTimeout` and unit-test it at `ms: 1` — one spec for the timeout rejection, one for pass-through-and-clear, one asserting a late rejection does not surface as unhandled.

---

- **File:** `packages/core/src/pdf-rasterize.js:82`
- **Severity:** Medium
- **Reviewer:** stack:code-reviewer (F2)
- **Issue:** The comment claims *every* in-page call is raced against a timeout. Two are not: `page.eval(new Function(pdfjsSource))` at line 82 (injecting ~2MB of pdf.js into a page that may already be unresponsive) and `page.eval(destroyDocument)` at line 138. Either reproduces the original defect exactly — request hangs forever holding a browser page and a listening asset server.
- **Suggestion:** Wrap both in `withTimeout`, or move `destroyDocument` into the `finally` under `allSettled`.

---

- **File:** `packages/core/src/pdf-rasterize.js:110`
- **Severity:** Medium
- **Reviewer:** stack:code-reviewer (F3)
- **Issue:** The timeout is per-call, so a PDF whose pages each render in 29s runs 250 × 29s ≈ 2 hours on one held-open connection without tripping anything. A per-call timeout catches a wedge, not slowness at scale — which is the case the page cap was added for.
- **Suggestion:** Compute one deadline up front and shrink each per-call budget against it.

---

- **File:** `packages/cli-pdf/src/pages.js:1`
- **Severity:** Medium
- **Reviewer:** stack:code-reviewer (F4)
- **Issue:** `MAX_PAGES` caps page count, but the per-page ceiling is `MAX_DIMENSION = 2000`, so 250 content-heavy pages is ~0.5-1.5GB resident — the very failure mode the comment describes. The cap only helps against the "thousands of tiny pages" shape.
- **Suggestion:** Add a running byte budget in the render loop where `png.length` is known, or make the comment honest about the worst case.

---

- **File:** `packages/core/src/pdf-rasterize.js:141`
- **Severity:** Medium
- **Reviewer:** stack:code-reviewer (F6)
- **Issue:** Correct for the leak it fixes, but a `server.close()` failure used to propagate and now vanishes unlogged — a socket that fails to drain leaves a port holding the customer's PDF with nothing recording it.
- **Suggestion:** Iterate the settled results and `log.debug` any rejection.

---

- **File:** `packages/core/test/pdf-snapshot.test.js:234`
- **Severity:** Medium
- **Reviewer:** stack:code-reviewer (F5)
- **Issue:** The new oversized-PDF spec allocates a ~70MB string and round-trips it over HTTP — roughly 350-400MB transient, on a Node 14 runner, in a suite that already sets a 240s timeout and drives a real browser. The assertion itself is load-bearing (`expect(fromSpy).not.toHaveBeenCalledWith(...)` is the only thing distinguishing the new pre-decode check from the pre-existing post-decode 413).
- **Suggestion:** Move it into the existing `describe('decodePdf')` block and call `decodePdf` directly — same strength, a third of the memory, no server round trip. (`spyOn(Buffer, 'from')` itself is safe: writable/configurable, `callThrough` preserves behavior, `PDF_MAGIC` is built at module load before the spy installs.)

---

- **File:** `packages/core/src/pdf-rasterize.js:43`
- **Severity:** Low
- **Reviewer:** stack:code-reviewer (F13)
- **Issue:** `Promise.race` calls `.then(resolve, reject)` on every input, so a loser's later rejection is already handled and an unhandled-rejection warning cannot occur. Verified empirically. The line is dead code and its comment asserts a hazard that does not exist — and it is one of the uncovered functions in F1.
- **Suggestion:** Delete both the line and the comment.

---

- **File:** `packages/core/src/utils.js:520`
- **Severity:** Low
- **Reviewer:** stack:code-reviewer (F10)
- **Issue:** `&` is currently unreachable (both call sites use `encodeURIComponent`), but if one ever slips through, `escapeAttr` produces `src="…&amp;…"` against a resource URL of `…&…` — byte-for-byte the bug just fixed, with the same silence. A silent rewrite is the wrong backstop for a contract that fails silently.
- **Suggestion:** Warn when `[&"]` is seen, and add the `&` case to the property test so the escaping behavior itself is pinned.

---

- **File:** `packages/core/src/pdf-snapshot.js:14`
- **Severity:** Low
- **Reviewer:** stack:code-reviewer (F11, F12)
- **Issue:** The arithmetic is exact — `Math.ceil(52428800/3)*4 === 69905068`, matching `Buffer.alloc(50MB).toString('base64').length` with a `>` comparison, so an exactly-50MB PDF passes. But `Buffer.from(s, 'base64')` tolerates whitespace, and MIME-style 76-column wrapping inflates by ~1.3%: a 49.4MB line-wrapped PDF gets a spurious 413. Separately, the comment claims the check prevents a 1GB body being "buffered, JSON-parsed and decoded" — all three still happen except the decode.
- **Suggestion:** Compare `content.replace(/\s/g, '').length`, and either correct the comment or add the cap in `IncomingMessage`.

---

- **File:** `packages/cli-pdf/src/pages.js:80`
- **Severity:** Low
- **Reviewer:** stack:code-reviewer (F14)
- **Issue:** With no `pages` supplied, "Requested 251 pages but the maximum per request is 250" tells the caller they requested something they did not.
- **Suggestion:** Reword when `pages == null`.

---

- **File:** `packages/core/src/page.js:99`
- **Severity:** Low
- **Reviewer:** stack:code-reviewer (F8, F9)
- **Issue:** `error.stack = description` discards the Node-side frames entirely, so the stack names none of the eleven `page.eval` call sites. Separately, `JSON.stringify(value)` can throw on a cyclic thrown object, replacing the real page error with the serializer's.
- **Suggestion:** Append the local stack after the remote one; wrap the stringify in try/catch. (Verified correct: `0` and `false` take the `String(value)` branch, and both are pinned by tests.)

### Verified fixed by this commit

`escapeAttr` (prior finding 1) — correct scope, `'` preserved, good property test. `remoteError` (2) — no caller in the monorepo consumed the rejection as a string; `page.js:223` is genuinely a different path. `asInputError` / `error?.status ?? 500` (6) — audited, no `.status` collisions. Pre-decode base64 cap (4) — arithmetic exact. `Promise.allSettled` cleanup (7) — `page?.close()` → `undefined` handled correctly.

### Still open from the previous review

5 (`::` bind), 8 (darwin-arm64 goldens unverified; `ubuntu-latest` is a moving image for byte-exact goldens), 9 (240s timeout + real browser in the unit suite, worsened by F5), 10 (constants duplicated), 11 (now carries a justifying comment, which is a reasonable answer), 12 (`_parseSelection`, no README, `loadPdfModule()` result discarded then re-imported).

> CI at time of review: 48 pass, 2 fail, 1 pending. `Test @percy/core` fails on F7 above. `Test @percy/cli-exec` fails only on the Windows runner, on "handles terminating the child process when interrupted" — a SIGTERM/child-process spec that touches nothing in this PR; the same job passes on the Linux legs of the same commit.

> Security-specialist reviewers remain disabled in the orchestrator, so the Security rows reflect only what the general code reviewer surfaced.

---

**Verdict: FAIL**

RaghavsBrowserStack:
# Claude Code PR Review

**PR:** https://github.com/percy/cli/pull/2418  •  **Head:** e6b4108  •  **Reviewers:** stack:code-reviewer

> Third pass. `e6b41084` closes every High finding from the review of `750a5f06`. Verified against CI rather than asserted: `Test @percy/core` passes on the Linux workflow having run `test:coverage`, which enforces `.nycrc`'s 100% branches/lines/functions/statements.

## Summary
Native PDF visual testing via `POST /percy/pdf/snapshot` — a new `@percy/cli-pdf`, rasterization inside the existing discovery browser, a shared image-snapshot wrapper, an sdk-utils helper and a byte-exact PNG golden track — with the escaping, error-surfacing, bounding and status-classification defects from the prior two reviews now fixed.

## Review Table

| High | Security | No hardcoded secrets or credentials | Pass | None in the diff. |
| High | Security | Authentication/authorization checks present | Pass | Fixed. The asset server now binds `127.0.0.1` (`pdf-rasterize.js:9`) instead of inheriting `Server`'s `'::'`. An explicit host beats `PERCY_SERVER_HOST`, so widening the API server cannot widen the one serving the customer's document. |
| High | Security | Input validation and sanitization | Pass | Scale, page selection, magic bytes and size (pre- and post-decode) validated; the `escapeHtml`/`escapeAttr` split is contextually correct and property-tested. |
| High | Security | No IDOR — resource ownership validated | N/A | No user-scoped identifiers on this path. |
| High | Security | No SQL injection (parameterized queries) | N/A | No SQL in the CLI. |
| High | Correctness | Logic is correct, handles edge cases | Pass | F7 fixed and confirmed by CI — `snapshot.test.js:1754` passes again. F2 fixed: the pdf.js injection and `destroyDocument` are now raced too, so the "every in-page call" comment is finally true. |
| High | Correctness | Error handling is explicit, no swallowed exceptions | Pass | F6 fixed — cleanup rejections are logged rather than discarded. F13 fixed — the redundant `.catch()` and its incorrect rationale are gone. |
| High | Correctness | No race conditions or concurrency issues | Pass | `Promise.race` + `.finally(clearTimeout)` sound on all paths; a timed-out page is closed, so the shared browser is not left holding a wedged target. |
| Medium | Testing | New code has corresponding tests | Pass | `withTimeout` is exported and covered by 6 specs (timeout, naming, pass-through, early rejection, timer cleanup, no unhandled late rejection). The 100% gate passed in CI. |
| Medium | Testing | Error paths and edge cases tested | Pass | 400-vs-500, non-Error throws, the undefined-message regression, cap boundary, apostrophe/angle-bracket URLs, loopback bind, and now every `withTimeout` path. |
| Medium | Testing | Existing tests still pass (no regressions) | Pass | `Test @percy/core` green (19m38s, with coverage). `Test @percy/cli-exec` green on both legs including Windows. 50 pass / 1 pending at time of writing; the pending check is the long-running Windows workflow. |
| Medium | Performance | No N+1 queries or unbounded data fetching | Fail | Open, non-blocking. F4: `MAX_PAGES` bounds page *count*, not bytes — 250 pages at the 2000px ceiling is still ~0.5-1.5GB resident. F12: the request body is buffered and JSON-parsed with no cap before the route runs. |
| Medium | Performance | Long-running tasks use background jobs | Fail | Open, non-blocking. F3: the timeout is per-call, so 250 pages × 30s ≈ 2h on one held-open request without tripping anything. |
| Medium | Quality | Follows existing codebase patterns | Pass | `asInputError` composes with `server.js`'s existing `let { status = 500 } = error`; the new `host` option follows the existing constructor-options shape. |
| Medium | Quality | Changes are focused (single concern) | Pass | Every hunk maps to a named finding. |
| Low | Quality | Meaningful names, no dead code | Fail | Open, non-blocking. `_parseSelection` (`pages.js:86`) is still exported and imported nowhere. |
| Low | Quality | Comments explain why, not what | Fail | Open, non-blocking. F13's misleading comment is gone, but F12's still claims the pre-decode check stops a 1GB body being "buffered, JSON-parsed and decoded" — only the decode is actually prevented. |

## Findings

No Critical or High findings remain.

### Closed by `e6b41084`

- **`packages/core/src/page.js:97` (was High).** `remoteError` gave the Error a default `name`, so @percy/logger — which renders a thrown Error as `Error.prototype.toString` and only falls back to `stack` at debug level — printed `Error: Error: test error` and dropped the remote frames a user needs to debug their own execute script. `Test @percy/core` was red on this. Blanking `name` makes `toString` return the CDP description verbatim, byte-identical to the string this path used to throw. Note the obvious alternative fix — parsing the name off the first line — does *not* work: it strips the frames and leaves the spec red.
- **`packages/core/src/pdf-rasterize.js:42` (was High).** `withTimeout` had no test on a package with a 100% coverage gate, so the fix for "no render timeout" shipped with no evidence it worked. Now exported and covered. `timer.unref?.()` became `timer.unref()` — the optional call's false branch is unreachable in Node and would have failed the branch gate by itself.
- **`packages/core/src/pdf-rasterize.js:9` (was a High row).** Asset server binds loopback.
- **`packages/core/src/pdf-rasterize.js:86,145`.** The two remaining un-raced evals now go through `withTimeout`.
- **`packages/core/src/pdf-rasterize.js:152`.** Cleanup rejections logged.
- **`packages/core/src/pdf-rasterize.js`.** Dead `promise.catch(() => {})` removed — `Promise.race` attaches handlers to every input, verified empirically.

### Open, non-blocking

Medium: F3 (no aggregate request deadline), F4 (page-count cap does not bound bytes), F5 (the oversized-PDF spec allocates ~70MB and round-trips it over HTTP — it killed a local full-suite run for memory during this review, so it is not theoretical), F11 (whitespace-wrapped base64 near the limit gets a spurious 413), F12 (comment overstates what the pre-decode check closes), plus prior findings 8 (darwin-arm64 goldens unverified; `ubuntu-latest` is a moving image for byte-exact goldens) and 9 (240s timeout and a real browser in the unit suite).

Low: F8 (local stack discarded), F9 (`JSON.stringify` can throw on a cyclic thrown value), F10 (`&` would recur the contract break if it ever reached `escapeAttr` — currently unreachable), F14 (cap message misattributes the request), plus prior findings 10 (constants duplicated), 11 (`UPLOAD_CLIENT_INFO`, now carrying a justifying comment) and 12 (`_parseSelection`, no README, `loadPdfModule()` result discarded then re-imported).

> Security-specialist reviewers remain disabled in the orchestrator, so the Security rows reflect only what the general code reviewer surfaced — not a substitute for a dedicated security pass.

---

**Verdict: PASS**

packages/cli-pdf/src/pages.js
```diff
@@ -0,0 +1,89 @@
+// Ceiling on how many pages one request may rasterize. Every page is held in
+// memory as a PNG twice over -- once in the rasterizer's result array, once in
+// the resource closure the snapshot queue keeps -- and each additionally crosses
+// CDP as a base64 data URL. A 50MB PDF can carry thousands of pages, so without
+// a cap a single request can exhaust the heap. Callers who genuinely want more
+// can narrow with `pages` and issue several requests.
+export const MAX_PAGES = 250;
```
rishigupta1599, line 7: this should come from BE, else you will need to do release everytime. You can get it as part of build creation

rishigupta1599: maybe send it as follow up but create a ticket and link to epic

rishigupta1599 (approved):

packages/cli-pdf/src/pages.js
RaghavsBrowserStack: done https://browserstack.atlassian.net/browse/PPLT-6092

PR #2426, opened 2026-09-10
fix(core): allowlist the /percy/log level so request input never drives dynamic dispatch (PER-8625)

ninadbstack:
## Summary

- **Security fix for [PER-8625](https://browserstack.atlassian.net/browse/PER-8625)** (chain-breaker finding [PER-8606](https://browserstack.atlassian.net/browse/PER-8606), CWE-1321). `POST /percy/log` used the request-supplied `level` as a property key on the logger group object: `log[level](message, meta)`.
- On the unauthenticated local API server this let any caller invoke arbitrary logger methods. Confirmed side effects before the fix: `{"level":"loglevel","message":"debug"}` returned 200 and flipped the process-wide log level; `{"level":"deprecated",...}` emitted arbitrary deprecation warnings. `__proto__`/`constructor` produced a 500 rather than pollution, but the primitive is the same one the ticket asks us to close.
- `level` is now validated against `debug|info|warn|error` and rejected with **400 `{ error: 'Invalid log level' }`** otherwise. Dispatch is an explicit `switch`, so no request input reaches a dynamic property lookup.
- SDK behaviour is unchanged: `@percy/sdk-utils` only ever sends the four valid levels.

## Test plan

- [x] New spec: every invalid level (`__proto__`, `constructor`, `loglevel`, `deprecated`, `stdout`, `''`, `null`, `42`) gets a 400, the global loglevel is untouched, nothing is logged, `Object.prototype` is clean. Verified this spec **fails** against the unpatched handler.
- [x] New spec: all four valid levels dispatch and are recorded with the right level (keeps the `switch` at 100% coverage).
- [x] Existing `/log` spec and the full `API Server` describe block pass locally (39 specs); `eslint` clean.
- [ ] CI green (full suite + coverage threshold)

[PER-8625]: https://browserstack.atlassian.net/browse/PER-8625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PER-8606]: https://browserstack.atlassian.net/browse/PER-8606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/api.js
+++ b/packages/core/src/api.js
@@ -97,6 +97,9 @@ function assertNotCrossOrigin(req) {
   }
 }
 
+// Log levels an SDK may forward through POST /percy/log.
+const SDK_LOG_LEVELS = new Set(['debug', 'info', 'warn', 'error']);
+
 // Create a Percy CLI API server instance
 export function createPercyServer(percy, port) {
   let pkg = getPackageJSON(import.meta.url);
@@ -319,7 +322,20 @@ export function createPercyServer(percy, port) {
       const message = req.body.message;
       const meta = req.body.meta || {};
 
-      log[level](message, meta);
+      // `level` is attacker-controlled input on an unauthenticated endpoint, so
+      // it must never be used as a dynamic property key on the logger object
+      // (CWE-1321 / PER-8606, PER-8625). Only the four log levels are dispatched;
+      // anything else (`__proto__`, `constructor`, `loglevel`, ...) is rejected.
+      if (!SDK_LOG_LEVELS.has(level)) {
+        return res.json(400, { error: 'Invalid log level' });
+      }
+
+      switch (level) {
+        case 'debug': log.debug(message, meta); break;
+        case 'info': log.info(message, meta); break;
+        case 'warn': log.warn(message, meta); break;
+        default: log.error(message, meta);
+      }
 
       res.json(200, { success: true });
     })
--- a/packages/core/test/api.test.js
+++ b/packages/core/test/api.test.js
@@ -896,6 +896,41 @@ describe('API Server', () => {
     expect(sdkLogs[1].meta).toEqual(message2.meta);
   });
 
+  it('rejects /log requests whose level is not a real log level (PER-8606/8625)', async () => {
+    await percy.start();
+    let before = logger.loglevel();
+
+    for (let level of ['__proto__', 'constructor', 'loglevel', 'deprecated', 'stdout', '', null, 42]) {
+      let [data, res] = await request('/percy/log', {
+        body: { level, message: 'debug', meta: {} },
+        method: 'post'
+      }, true);
+
+      expect(res.statusCode).toBe(400);
+      expect(data).toEqual({ error: 'Invalid log level' });
+    }
+
+    // no dynamic dispatch side effects: global loglevel untouched, nothing logged
+    expect(logger.loglevel()).toBe(before);
+    expect(logger.instance.query(log => log.debug === 'sdk').length).toBe(0);
+    expect(Object.prototype.token).toBeUndefined();
+  });
+
+  it('dispatches every allowed /log level', async () => {
+    await percy.start();
+    logger.loglevel('debug');
+
+    for (let level of ['debug', 'info', 'warn', 'error']) {
+      await expectAsync(request('/percy/log', {
+        body: { level, message: `${level} message` },
+        method: 'post'
+      })).toBeResolvedTo({ success: true });
+    }
+
+    const sdkLogs = logger.instance.query(log => log.debug === 'sdk');
+    expect(sdkLogs.map(l => l.level)).toEqual(['debug', 'info', 'warn', 'error']);
+  });
+
   it('returns a 500 error when an endpoint throws', async () => {
     spyOn(percy, 'snapshot').and.rejectWith(new Error('test error'));
     await percy.start();


rishigupta1599 (approved):

packages/core/src/api.js
```diff
@@ -97,6 +97,9 @@ function assertNotCrossOrigin(req) {
   }
 }
 
+// Log levels an SDK may forward through POST /percy/log.
+const SDK_LOG_LEVELS = new Set(['debug', 'info', 'warn', 'error']);
```
ninadbstack, line 101: **[Low] Level allowlist duplicates the logger's own level list**

This set repeats the levels already defined as `LOG_LEVELS` in `packages/logger/src/logger.js`, which is not exported. If a level is ever added or renamed there without updating this copy, the two drift silently and this endpoint starts rejecting a valid level.

**Suggestion:** Export the level list from `@percy/logger` and import it here rather than keeping a second hardcoded copy. Fine as a follow-up.

packages/core/src/api.js
```diff
@@ -319,7 +322,20 @@ export function createPercyServer(percy, port) {
...
-      log[level](message, meta);
+      // `level` is attacker-controlled input on an unauthenticated endpoint, so
+      // it must never be used as a dynamic property key on the logger object
+      // (CWE-1321 / PER-8606, PER-8625). Only the four log levels are dispatched;
+      // anything else (`__proto__`, `constructor`, `loglevel`, ...) is rejected.
+      if (!SDK_LOG_LEVELS.has(level)) {
+        return res.json(400, { error: 'Invalid log level' });
+      }
+
+      switch (level) {
```
ninadbstack, line 333: **[Low] Switch is redundant once the guard has run**

After `SDK_LOG_LEVELS.has(level)` passes, `level` is provably one of four safe literals, so `log[level](message, meta)` would be equally safe and shorter. The comment above reads as though the guard alone is the mitigation, which leaves this switch looking redundant to a later reader.

**Suggestion:** Keep the switch if it is deliberate defense in depth against any dynamic property access, and say so in the comment. Otherwise collapse it back to the guarded dynamic call.

ninadbstack (commented):

ninadbstack:
# Claude Code PR Review

**PR:** https://github.com/percy/cli/pull/2426  •  **Head:** fec8de0  •  **Reviewers:** stack-code-reviewer

## Summary
Allowlists the four real log levels on the unauthenticated `POST /percy/log` route and replaces the dynamic `log[level](...)` dispatch with an explicit switch, so a request-supplied level can no longer invoke arbitrary logger or `Object.prototype` members (CWE-1321); adds regression tests for rejected levels and for all four allowed levels.

## Review Table

| High | Security | No hardcoded secrets or credentials | Pass | No secrets or credentials in the diff. |
| High | Security | Authentication/authorization checks present | Pass | Route stays unauthenticated by design (local SDK API); the change tightens what unauthenticated input can reach. Existing cross-origin assertion is untouched. |
| High | Security | Input validation and sanitization | Pass | `level` is now validated against a four-value allowlist before use; invalid input returns 400. |
| High | Security | No IDOR — resource ownership validated | N/A | No resource lookup by identifier in this change. |
| High | Security | No SQL injection (parameterized queries) | N/A | No database access in this change. |
| High | Correctness | Logic is correct, handles edge cases | Pass | Set membership rejects non-strings, empty string, `null` and numbers; the switch covers all four allowlisted values, with `error` as the default arm. |
| High | Correctness | Error handling is explicit, no swallowed exceptions | Pass | Invalid level returns an explicit 400 with an error body; nothing is swallowed. |
| High | Correctness | No race conditions or concurrency issues | Pass | Module-level `Set` is immutable in practice and read-only per request; no shared mutable state added. |
| Medium | Testing | New code has corresponding tests | Pass | Both new branches (reject, dispatch) have dedicated tests. |
| Medium | Testing | Error paths and edge cases tested | Pass | Rejection test covers `__proto__`, `constructor`, `loglevel`, `deprecated`, `stdout`, empty string, `null` and a number, and asserts no side effects. |
| Medium | Testing | Existing tests still pass (no regressions) | Pass | The pre-existing `/percy/log` test uses allowlisted levels and is unaffected by the stricter validation. |
| Medium | Performance | No N+1 queries or unbounded data fetching | N/A | No data fetching in this change. |
| Medium | Performance | Long-running tasks use background jobs | N/A | Handler remains a constant-time dispatch. |
| Medium | Quality | Follows existing codebase patterns | Pass | Matches the file's existing `res.json(<status>, { error })` convention for rejected input. |
| Medium | Quality | Changes are focused (single concern) | Pass | Two files, one concern: log-level dispatch hardening plus its tests. |
| Low | Quality | Meaningful names, no dead code | Pass | `SDK_LOG_LEVELS` is clear; no dead code introduced. |
| Low | Quality | Comments explain why, not what | Pass | The comment states the threat and cites the tickets. See the finding below on making the defense-in-depth intent explicit. |

## Findings

- **File:** `packages/core/src/api.js:101`
- **Severity:** Low
- **Issue:** The level allowlist duplicates the level set already defined as `LOG_LEVELS` in `packages/logger/src/logger.js`, which is not exported. If a level is added or renamed in the logger without updating this copy, the two lists drift silently and the endpoint rejects a valid level.
- **Suggestion:** Export the level list from `@percy/logger` and import it here instead of maintaining a second hardcoded copy. Reasonable as a follow-up rather than a change in this PR.

- **File:** `packages/core/src/api.js:333`
- **Severity:** Low
- **Issue:** Once the `Set` guard has run, `level` is provably one of four safe literals, so the switch is not required for correctness or safety. The comment reads as though the guard alone is the mitigation, which leaves the switch looking redundant to a future reader.
- **Suggestion:** Keep the switch if it is intended as defense in depth against all dynamic property access, and say so in the comment. Otherwise the guarded `log[level](message, meta)` is equivalent and shorter.

- **File:** `packages/core/src/api.js:322`
- **Severity:** Low
- **Issue:** Pre-existing and outside this PR's scope. `message` and `meta` from the same unauthenticated body are still passed through unvalidated, and the logger calls `.toString()` on `message`, so a `null` message can throw inside the handler.
- **Suggestion:** File a follow-up to validate the `message` and `meta` shapes on this route, since they cross the same trust boundary.

> Note: the review-threads lookup returned no data this run, so thread-resolution state was unavailable. Reconciliation used REST comment data only.

The other human reviewer approved this PR without inline comments, so there were no human concerns to confirm or carry forward.

---

**Verdict: PASS** — the security fix is correct and well covered by tests; all three findings are Low and non-gating.

PR #2427, opened 2026-09-10
fix(sdk-utils): bound the readiness eval on the Node side

rishigupta1599:
Fixes PER-10756 / percy/percy-playwright#655

## The problem

Using Playwright's mocked clock makes every interaction *after* `percySnapshot()` fail:

```js
test('something', async ({ page }) => {
  await page.clock.pauseAt(new Date())
  await page.goto('https://example.com/')
  await percySnapshot(page, 'test')
  await expect(page.locator('h1')).toHaveCount(1)   // ← fails, page is gone
})
```

```
[percy:playwright] waitForReady failed, proceeding to serialize: page.evaluate: Test timeout of 30000ms exceeded.
[percy:playwright] Could not take DOM snapshot "test"
[percy:playwright] page.evaluate: Target page, context or browser has been closed
```

## Root cause

`PercyDOM.waitForReady()` (`packages/dom/src/readiness.js:527`) is a `Promise.race` where **both** arms depend on in-page timers:

- every check settles on a `setTimeout`/`setInterval` — `checkDOMStability` resolves *only* from `setTimeout(settle, stabilityWindowMs)` (`readiness.js:188`), and network-idle (`:236`), image-ready (`:294`) and js-idle (`:332`/`:364`/`:393`) are the same shape;
- the timeout arm is itself `setTimeout(..., effectiveTimeout)` (`readiness.js:530`).

Playwright's `page.clock.pauseAt()` replaces and freezes the page's `setTimeout`, `setInterval`, `requestAnimationFrame`, `requestIdleCallback`, `Date` and `performance.now`. So neither the checks nor the gate's own 10s timeout can ever fire, and the promise that `page.evaluate()` is awaiting stays pending for the life of the page.

`runReadinessGate` awaits that eval unconditionally, so `percySnapshot()` blocks until the *test runner's* timeout kills the test — after which the page is torn down and every later assertion fails. The customer's stack is just the fallout: the `waitForReady failed` line is this function's own catch, and `index.js:246` is the serialize eval hitting an already-closed page.

`sinon.useFakeTimers()` and jest fake timers hit exactly the same wall.

### Reproduction

Against `@percy/dom` 1.32.9, injecting the real bundle and evaluating the exact script `waitForReadyScript()` emits:

| clock | result |
|---|---|
| real | resolves in **304ms**, all 5 checks pass |
| `page.clock.pauseAt(new Date())` | **still pending at 14s** — past the balanced preset's 10s in-page timeout, which cannot fire either |
| `preset: disabled` | resolves in 22ms, page stays usable |

## The fix

Node is the only side of the eval boundary guaranteed to have a real clock, so the backstop belongs there rather than in any one SDK. `runReadinessGate` now races the eval against a deadline of the effective in-page timeout plus a 3s grace. On expiry it logs at debug and returns `null` — the same graceful degradation the gate already had for a rejected eval, so `serialize()` still runs and the snapshot is captured ungated.

Verified that abandoning a hung readiness eval leaves the page fully usable: `PercyDOM.serialize()` returned a complete 104KB DOM in 10ms afterwards, and the post-snapshot `expect(page.locator('h1')).toHaveCount(1)` passed.

Two details worth reviewing:

- **The abandoned eval's later rejection is swallowed.** When the driver closes the page at end of test, the orphaned `page.evaluate` rejects; without the `.catch` that becomes an unhandled rejection.
- **`PRESET_TIMEOUT_MS` duplicates the preset timeouts** from `@percy/dom`'s `readiness.js`, which only exists in the browser bundle. A parity test reads the dom source and asserts they still match — same approach as the existing `serialize-frames` constant mirror, so drift fails loudly rather than silently producing a Node deadline shorter than the in-page one.

Fixing it here covers every SDK that routes through the gate — Playwright, Puppeteer, Selenium-js, WebdriverIO, Nightwatch, Ember — in one place.

## Scope note

`strict` (30s in-page → 33s deadline) still exceeds Playwright's default 30s test timeout, so a strict-preset gate that genuinely runs that long will fail the test first. That is pre-existing and independent of the frozen-clock bug.

## Testing

`packages/sdk-utils/test/index.test.js` — new cases:

- a never-settling eval returns `null` at the deadline instead of hanging, and the debug log names the deadline and points at the workaround
- an abandoned eval rejecting later (page closed) raises no unhandled rejection
- an eval that settles before the deadline still returns its diagnostics
- `readinessDeadlineMs`: preset defaults, unknown-preset fallback, explicit `timeoutMs`/`timeout_ms` precedence, `maxTimeoutMs` clamping
- parity with `@percy/dom`'s `PRESETS` timeouts

Existing `runReadinessGate` cases (disabled short-circuit, config shallow-merge, callback/promise script mode, rejection and sync-throw degradation) are unchanged and still pass.

## Workaround for users on a released CLI

```yaml
# .percy.yml
snapshot:
  readiness:
    preset: disabled
```

Or `page.clock.resume()` before `percySnapshot()` and re-pause after.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented DOM serialization from waiting indefinitely when readiness checks do not complete.
  - Serialization now proceeds after the configured readiness deadline, including when simulated timers are active.
  - Improved handling of delayed or failed readiness evaluations so late errors do not affect results.
  - Readiness diagnostics are retained when available.

- **New Features**
  - Exposed readiness deadline calculation through the SDK utilities for consistent timeout management across supported configurations.

### DIFF
--- a/packages/sdk-utils/src/serialize-dom.js
+++ b/packages/sdk-utils/src/serialize-dom.js
@@ -106,17 +106,72 @@ export function waitForReadyScript(readinessConfig = {}, { callback = false } = 
 //     options,
 //     { callback: true, log }
 //   );
+// Effective in-page timeout for each readiness preset, mirroring PRESETS in
+// @percy/dom's readiness.js. Duplicated here because the presets live in the
+// browser bundle, and the Node side needs the number to size its own deadline.
+const PRESET_TIMEOUT_MS = { balanced: 10000, strict: 30000, fast: 5000 };
+
+// Added to the in-page timeout before the Node side gives up, so a gate that is
+// merely slow (eval round-trip, check teardown) is never pre-empted.
+const READINESS_DEADLINE_GRACE_MS = 3000;
+
+// Wall-clock budget the Node side gives the readiness eval.
+//
+// PercyDOM.waitForReady() bounds itself with an in-page `setTimeout`, and every
+// individual check settles on a `setTimeout`/`setInterval`. On a page whose
+// timers are faked and paused -- Playwright's `page.clock.pauseAt()`, sinon
+// `useFakeTimers`, jest fake timers -- neither the checks nor the gate's own
+// timeout can ever fire, so the eval stays pending for the life of the page and
+// hangs the test that called percySnapshot(). Node is the only side of that
+// boundary guaranteed to have a real clock, so the backstop belongs here.
+export function readinessDeadlineMs(readinessConfig = {}) {
+  let timeout = readinessConfig.timeoutMs ?? readinessConfig.timeout_ms ??
+    PRESET_TIMEOUT_MS[readinessConfig.preset] ?? PRESET_TIMEOUT_MS.balanced;
+  let max = readinessConfig.maxTimeoutMs ?? readinessConfig.max_timeout_ms;
+  if (max != null) timeout = Math.min(timeout, max);
+  return timeout + READINESS_DEADLINE_GRACE_MS;
+}
+
+const READINESS_DEADLINE_HIT = Symbol('readiness-deadline');
+
 export async function runReadinessGate(evalScript, snapshotOptions = {}, { callback = false, log } = {}) {
   if (isReadinessDisabled(snapshotOptions)) return null;
   const config = getReadinessConfig(snapshotOptions);
   const script = waitForReadyScript(config, { callback });
+  const deadline = readinessDeadlineMs(config);
+  let timer;
+
   try {
-    return await evalScript(script);
+    const evaluation = Promise.resolve(evalScript(script));
+    // Hitting the deadline abandons `evaluation`, which may still reject much
+    // later -- when the driver tears the page down at end of test, say. Swallow
+    // that here so it never surfaces as an unhandled rejection.
+    evaluation.catch(() => {});
+
+    const result = await Promise.race([
+      evaluation,
+      new Promise(resolve => { timer = setTimeout(() => resolve(READINESS_DEADLINE_HIT), deadline); })
+    ]);
+
+    if (result === READINESS_DEADLINE_HIT) {
+      if (log && typeof log.debug === 'function') {
+        log.debug(
+          `waitForReady did not settle within ${deadline}ms, proceeding to serialize. ` +
+          'If this page fakes or pauses timers (e.g. Playwright page.clock), disable ' +
+          'the readiness gate with snapshot.readiness.preset: disabled.'
+        );
+      }
+      return null;
+    }
+
+    return result;
   } catch (err) {
     if (log && typeof log.debug === 'function') {
       log.debug(`waitForReady failed, proceeding to serialize: ${err?.message || err}`);
     }
     return null;
+  } finally {
+    clearTimeout(timer);
   }
 }
 
--- a/packages/sdk-utils/src/serialize-dom.js
+++ b/packages/sdk-utils/src/serialize-dom.js
@@ -132,6 +132,24 @@ export function readinessDeadlineMs(readinessConfig = {}) {
   return timeout + READINESS_DEADLINE_GRACE_MS;
 }
 
+// Captured once at module load, because the deadline below must not be the very
+// thing a faked clock disables. jest's and sinon's fake timers replace the
+// *global* `setTimeout` binding, and a bare `setTimeout(...)` call resolves that
+// global at call time -- so a consumer whose Node test process has fake timers
+// installed would schedule the deadline on a frozen clock and hang exactly as
+// before, one layer up from the in-page freeze this gate exists to survive.
+//
+// A module-scope capture holds because the SDK imports this module at require
+// time, before a test body reaches `useFakeTimers()`. It is deliberately not
+// `import { setTimeout } from 'node:timers'`, which would be immune to import
+// order but pulls a Node builtin into this file -- and this file is imported
+// statically by index.js, which is bundled for the browser too (see the package's
+// `browser` field). The one Node-only dependency in this package is reached by a
+// lazy `await import('./proxy.js')` in request.js precisely to keep builtins out
+// of that bundle; a static builtin import here would defeat it.
+const nativeSetTimeout = globalThis.setTimeout;
+const nativeClearTimeout = globalThis.clearTimeout;
+
 const READINESS_DEADLINE_HIT = Symbol('readiness-deadline');
 
 export async function runReadinessGate(evalScript, snapshotOptions = {}, { callback = false, log } = {}) {
@@ -150,7 +168,7 @@ export async function runReadinessGate(evalScript, snapshotOptions = {}, { callb
 
     const result = await Promise.race([
       evaluation,
-      new Promise(resolve => { timer = setTimeout(() => resolve(READINESS_DEADLINE_HIT), deadline); })
+      new Promise(resolve => { timer = nativeSetTimeout(() => resolve(READINESS_DEADLINE_HIT), deadline); })
     ]);
 
     if (result === READINESS_DEADLINE_HIT) {
@@ -171,7 +189,7 @@ export async function runReadinessGate(evalScript, snapshotOptions = {}, { callb
     }
     return null;
   } finally {
-    clearTimeout(timer);
+    nativeClearTimeout(timer);
   }
 }
 
--- a/packages/sdk-utils/src/serialize-dom.js
+++ b/packages/sdk-utils/src/serialize-dom.js
@@ -139,14 +139,23 @@ export function readinessDeadlineMs(readinessConfig = {}) {
 // installed would schedule the deadline on a frozen clock and hang exactly as
 // before, one layer up from the in-page freeze this gate exists to survive.
 //
-// A module-scope capture holds because the SDK imports this module at require
-// time, before a test body reaches `useFakeTimers()`. It is deliberately not
-// `import { setTimeout } from 'node:timers'`, which would be immune to import
-// order but pulls a Node builtin into this file -- and this file is imported
-// statically by index.js, which is bundled for the browser too (see the package's
-// `browser` field). The one Node-only dependency in this package is reached by a
-// lazy `await import('./proxy.js')` in request.js precisely to keep builtins out
-// of that bundle; a static builtin import here would defeat it.
+// The capture holds for the ordinary case: the SDK imports this module at
+// require time, before a test body or beforeEach reaches `useFakeTimers()`. It
+// is NOT a guarantee. Fake timers installed before this module is first
+// evaluated -- jest's `fakeTimers: { enableGlobally: true }`, a `useFakeTimers()`
+// call in `setupFiles`, or `resetModules()` + a fresh require under an already
+// faked clock -- capture the fake, and the hang returns. Those consumers need
+// `snapshot.readiness.preset: disabled`.
+//
+// Deliberately not `import { setTimeout } from 'node:timers'`, which would be
+// immune to import order too: this file is imported statically by index.js, and
+// index.js is the rollup entry for the browser bundle (see the package's
+// `browser` field). Node-only code in this package is always reached through a
+// lazy `await import(...)` -- `http`/`https` in request.js, `./proxy.js` and its
+// `net`/`tls` imports -- precisely to keep builtins out of that graph. A static
+// builtin import here would not even fail the build, since the rollup config
+// suppresses MISSING_NODE_BUILTINS; it would ship a browser bundle that breaks
+// at runtime, which is worse.
 const nativeSetTimeout = globalThis.setTimeout;
 const nativeClearTimeout = globalThis.clearTimeout;
 
--- a/packages/sdk-utils/src/index.js
+++ b/packages/sdk-utils/src/index.js
@@ -15,6 +15,7 @@ import {
   waitForReadyScript,
   getReadinessConfig,
   isReadinessDisabled,
+  readinessDeadlineMs,
   runReadinessGate
 } from './serialize-dom.js';
 
@@ -111,6 +112,7 @@ export {
   waitForReadyScript,
   getReadinessConfig,
   isReadinessDisabled,
+  readinessDeadlineMs,
   runReadinessGate
 };
 
--- a/packages/sdk-utils/test/index.test.js
+++ b/packages/sdk-utils/test/index.test.js
@@ -1087,10 +1087,24 @@ describe('SDK Utils', () => {
       globalThis.clearTimeout = () => {};
 
       try {
-        let result = await runReadinessGate(
-          () => new Promise(() => {}),
-          { readiness: { timeoutMs: 1000 } }
-        );
+        // Bound the wait on the REAL timer. Without it, a regression here (the
+        // deadline scheduled on the frozen clock) never settles the await, so
+        // the finally below never restores the globals and every later spec
+        // using a bare setTimeout hangs too — one regression becomes a
+        // suite-wide cascade instead of one clean failure. 6000ms is above the
+        // real 4000ms deadline (timeoutMs 1000 + 3000ms grace) so a healthy run
+        // resolves on its own, and below jasmine's 10s spec timeout so this
+        // rejection, not the runner, is what reports the regression.
+        let result = await Promise.race([
+          runReadinessGate(
+            () => new Promise(() => {}),
+            { readiness: { timeoutMs: 1000 } }
+          ),
+          new Promise((resolve, reject) => realSetTimeout(() => reject(new Error(
+            'readiness deadline never fired: it was scheduled on the frozen global clock ' +
+            'instead of the timer captured at module load'
+          )), 6000))
+        ]);
         expect(result).toBe(null);
         // Nothing was handed to the frozen clock — the capture was used.
         expect(scheduledOnFrozenClock).toBe(0);
--- a/packages/sdk-utils/test/index.test.js
+++ b/packages/sdk-utils/test/index.test.js
@@ -1043,6 +1043,108 @@ describe('SDK Utils', () => {
       let result = await runReadinessGate(() => Promise.reject(new Error('no log')), {});
       expect(result).toBe(null);
     });
+
+    // PER-10756: PercyDOM.waitForReady() enforces its own timeout with an
+    // in-page setTimeout, so a page with faked/paused timers (Playwright
+    // page.clock.pauseAt, sinon useFakeTimers) can never settle the gate. The
+    // Node-side deadline is what keeps percySnapshot() from hanging the test.
+    it('returns null when evalScript never settles, without blocking on it', async () => {
+      let logged;
+      let result = await runReadinessGate(
+        () => new Promise(() => {}),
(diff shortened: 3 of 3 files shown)


rishigupta1599 (commented):

packages/sdk-utils/src/serialize-dom.js
```diff
@@ -106,17 +106,72 @@ export function waitForReadyScript(readinessConfig = {}, { callback = false } =
...
-    return await evalScript(script);
+    const evaluation = Promise.resolve(evalScript(script));
+    // Hitting the deadline abandons `evaluation`, which may still reject much
+    // later -- when the driver tears the page down at end of test, say. Swallow
+    // that here so it never surfaces as an unhandled rejection.
+    evaluation.catch(() => {});
+
+    const result = await Promise.race([
+      evaluation,
+      new Promise(resolve => { timer = setTimeout(() => resolve(READINESS_DEADLINE_HIT), deadline); })
```
rishigupta1599: **[Medium] Node-side fake timers defeat this deadline**

This uses the bare global `setTimeout`, resolved at call time. If the consumer's *Node* test process has `jest.useFakeTimers()` or `sinon.useFakeTimers()` active (both patch the **global** binding), the deadline registers on the frozen fake clock and never fires — `Promise.race` stays pending and the hang returns one layer up. That undercuts this change's own rationale, which holds only if Node's *actual* timer implementation is what gets called.

Verified by emulating what jest/sinon do to the globals (replace `setTimeout`, never advance): current code **hangs**; a module-scope capture works only if the import precedes fake-timer installation; `node:timers` works in both orders (returns `null` at ~4s).

**Suggestion:** use the core-module timers — a separate binding from `globalThis`, so it is import-order-independent:

```js
import { setTimeout as nodeSetTimeout, clearTimeout as nodeClearTimeout } from 'node:timers';
// ...
new Promise(resolve => { timer = nodeSetTimeout(() => resolve(READINESS_DEADLINE_HIT), deadline); })
// ...
} finally { nodeClearTimeout(timer); }
```

Plus a regression test that installs Node-side fake timers and leaves `evalScript` pending. Rated Medium, not High: it does not regress anything (pre-PR had no protection) and the reported Playwright case is unaffected — `page.clock` fakes only the browser realm, not Node.

packages/sdk-utils/src/serialize-dom.js
```diff
@@ -106,17 +106,72 @@ export function waitForReadyScript(readinessConfig = {}, { callback = false } =
...
 //     { callback: true, log }
 //   );
+// Effective in-page timeout for each readiness preset, mirroring PRESETS in
+// @percy/dom's readiness.js. Duplicated here because the presets live in the
+// browser bundle, and the Node side needs the number to size its own deadline.
+const PRESET_TIMEOUT_MS = { balanced: 10000, strict: 30000, fast: 5000 };
+
+// Added to the in-page timeout before the Node side gives up, so a gate that is
+// merely slow (eval round-trip, check teardown) is never pre-empted.
+const READINESS_DEADLINE_GRACE_MS = 3000;
```
rishigupta1599, line 116: **[Medium] `strict` preset deadline exceeds Playwright's default test timeout**

`strict` is 30s in-page → 33s deadline here, above Playwright's default 30s per-test timeout. For a frozen clock under `strict`, the runner's own timeout fires first, so that test still fails with a Playwright timeout instead of degrading gracefully to `null`. The backstop still prevents an *indefinite* hang, but the surrounding comments imply a graceful save that does not hold for this tier.

**Suggestion:** state the tradeoff explicitly, or make the grace non-additive above a threshold so the deadline stays under common runner defaults.

rishigupta1599:
# Claude Code PR Review

**PR:** https://github.com/percy/cli/pull/2427  •  **Head:** 41ca4c0  •  **Reviewers:** stack-code-reviewer

## Summary

Adds a Node-side `Promise.race` deadline around the readiness `evalScript` call in `runReadinessGate`, sized from the same preset timeouts `PercyDOM.waitForReady()` uses in `@percy/dom` plus a 3s grace, so a page whose timers are faked and paused (Playwright `page.clock.pauseAt()`) can no longer hang `percySnapshot()` until the test runner's own timeout kills the test (PER-10756).

## Review Table

| High | Security | No hardcoded secrets or credentials | Pass | No secrets, tokens or URLs added. |
| High | Security | Input validation and sanitization | Pass | `readinessDeadlineMs` reads only numeric config already validated by the `snapshot.readiness` JSON schema; no new untrusted input path. |
| High | Correctness | Logic is correct, handles edge cases | Pass | Race, sentinel, `finally` cleanup and sync-throw path all verified correct. One narrower gap recorded as Medium (F1) — it does not regress existing behavior. |
| High | Correctness | Error handling is explicit, no swallowed exceptions | Pass | Deadline-hit and rejection both log at debug and degrade to the pre-existing "never block serialize" `null` path. The one deliberate swallow (`evaluation.catch(() => {})`) is required and documented. |
| High | Correctness | No race conditions or concurrency issues | Pass | `evaluation.catch` is attached in the same tick the promise is created, so an abandoned eval rejecting later (page teardown) cannot surface as an unhandled rejection. Confirmed by reading the code and by direct test. |
| Medium | Testing | New code has corresponding tests | Pass | 8 new cases; the never-settles case genuinely hangs against pre-PR code. |
| Medium | Testing | Error paths and edge cases tested | Pass | Deadline hit, late rejection of an abandoned eval, settle-before-deadline, sync throw, disabled short-circuit, and all `readinessDeadlineMs` branches. Gap noted in F1 (no Node-fake-timer test). |
| Medium | Testing | Existing tests still pass (no regressions) | Pass | CI green at this head: `Test @percy/sdk-utils` passed, `Lint` passed, 27 checks pass / 0 fail. |
| Medium | Performance | Long-running tasks use background jobs | N/A | Not applicable; the change *shortens* a pathological wait. |
| Medium | Quality | Follows existing codebase patterns | Pass | The mirrored-constant + parity-test approach copies the established `serialize-frames` precedent in the same test file; degradation reuses the gate's existing `null` semantics. |
| Medium | Quality | Changes are focused (single concern) | Pass | One concern, 3 files, additive only. |
| Low | Quality | Meaningful names, no dead code | Pass | No dead code; `READINESS_DEADLINE_HIT` is module-local and unexported. |
| Low | Quality | Comments explain why, not what | Pass | Comments carry the rationale (why the bound must live on the Node side), matching the file's existing style. |

## Findings

### F1 — Node-side fake timers defeat the new deadline

- **File:** `packages/sdk-utils/src/serialize-dom.js:153`
- **Severity:** Medium
- **Reviewer:** stack-code-reviewer (independently raised by CodeRabbit as Major; **confirmed empirically** by the orchestrator)
- **Issue:** The deadline uses the bare global `setTimeout`, resolved dynamically at call time. If the consumer's *Node* test process has `jest.useFakeTimers()` or `sinon.useFakeTimers()` active — both patch the **global** binding — the deadline itself registers on the frozen fake clock and never fires, so `Promise.race` stays pending and the hang returns one layer up. This directly undercuts the change's own stated rationale ("Node is the only side of that boundary guaranteed to have a real clock"), which holds only if Node's *actual* timer implementation is what gets called.
- **Verified, not asserted.** Emulating exactly what jest/sinon do to the globals (replace `setTimeout`, never advance):

  | Variant | globals faked *after* import | faked *before* import |
  |---|---|---|
  | current PR code (bare global) | **hangs** (deadline registered on the fake clock at 4000ms, never fires) | **hangs** |
  | module-scope capture of native | returns `null` at 4002ms | order-dependent — fails |
  | `import { setTimeout } from 'node:timers'` | returns `null` at 4002ms | **returns `null` at 4009ms** |

- **Suggestion:** use the core-module timers, which are a separate binding from `globalThis` and therefore import-order-independent (the module-scope capture only works if the import happens to precede fake-timer installation, and breaks under `jest.resetModules()` or a `setupFiles` install):

  ```js
  import { setTimeout as nodeSetTimeout, clearTimeout as nodeClearTimeout } from 'node:timers';
  // …
  new Promise(resolve => { timer = nodeSetTimeout(() => resolve(READINESS_DEADLINE_HIT), deadline); })
  // …
  } finally { nodeClearTimeout(timer); }
  ```

  Add a regression test that installs Node-side fake timers and leaves `evalScript` pending.
- **Why Medium and not High:** it does not regress anything — pre-PR code had no protection at all — and the reported customer scenario is unaffected, because Playwright's `page.clock` fakes only the browser realm, not Node (verified against real Chrome). It leaves a *different*, pre-existing hang unfixed rather than introducing one. The reviewer rated it "Medium-High" and explicitly declined to hard-block.

### F2 — `strict` preset deadline exceeds Playwright's default test timeout

- **File:** `packages/sdk-utils/src/serialize-dom.js:116`
- **Severity:** Medium
- **Issue:** `strict` is 30s in-page → 33s deadline, above Playwright's default 30s per-test timeout. For a frozen clock under `strict`, the runner's own timeout fires first, so that individual test still fails with a Playwright timeout rather than degrading gracefully to `null`. The backstop still prevents an *indefinite* hang, but the comments imply a graceful save that does not hold for this tier.
- **Suggestion:** state the tradeoff explicitly, or make the grace non-additive above a threshold so the deadline stays under common runner defaults.

### F3 — Deadline tests spend real wall-clock; grace is not injectable

- **File:** `packages/sdk-utils/test/index.test.js:112`
- **Severity:** Low
- **Issue:** The two deadline-hit tests use real timers with `timeoutMs: 1000`, so each waits the full grace-inclusive 4000ms — roughly 8s of unavoidable wall-clock added to the suite. `READINESS_DEADLINE_GRACE_MS` is a module constant with no test seam.
- **Suggestion:** drive these with mocked timers and explicit tick advancement, or make the grace injectable.

### Confirmed correct (no action)

The reviewer explicitly verified, by reading the code rather than assuming: the `Promise.race` / sentinel / `finally { clearTimeout }` mechanism including the sync-throw path; that `evaluation.catch(() => {})` genuinely prevents an unhandled rejection because it is attached in the creating tick; that `PRESET_TIMEOUT_MS` matches the current `PRESETS` in `packages/dom/src/readiness.js` and the parity-test regex actually extracts them correctly; and that the new export is purely additive and reachable via the package's `export * as default`, so SDKs pinned to older `sdk-utils` are unaffected.

---

**Verdict: PASS** — core fix is correct, tested and safely additive; F1 is a confirmed, cheap, non-regressing gap being addressed as a follow-up commit on this branch rather than shipped silently.

rishigupta1599:
# Claude Code PR Review

**PR:** https://github.com/percy/cli/pull/2427  •  **Head:** 07814bd  •  **Reviewers:** stack-code-reviewer

_Continues [the previous review](https://github.com/percy/cli/pull/2427#issuecomment-5624132762) — changes since `41ca4c0` (delta, two rounds: `8787e64` then `07814bd`)._

## Summary

Bounds the readiness `evalScript` call on the Node side so a page with frozen timers can no longer hang `percySnapshot()` (PER-10756), with the deadline itself scheduled on a timer captured at module load so a consumer's own fake timers cannot disable it, and a regression spec that fails cleanly rather than cascading.

## Review Table

| High | Security | No hardcoded secrets or credentials | Pass | None added. |
| High | Security | Input validation and sanitization | Pass | Only numeric config already validated by the `snapshot.readiness` schema. |
| High | Correctness | Logic is correct, handles edge cases | Pass | Race, sentinel, `finally` cleanup, sync-throw and captured-timer ordering all confirmed by reading the code. |
| High | Correctness | Error handling is explicit, no swallowed exceptions | Pass | Deadline hit and rejection both log at debug and degrade to the pre-existing "never block serialize" `null` path. The one deliberate swallow is required and documented. |
| High | Correctness | No race conditions or concurrency issues | Pass | `evaluation.catch` attached in the creating tick, so an abandoned eval rejecting at page teardown cannot surface as an unhandled rejection. |
| Medium | Testing | New code has corresponding tests | Pass | 9 cases. The never-settles case genuinely hangs against pre-PR code; the fake-timer case genuinely fails against the pre-capture commit. |
| Medium | Testing | Error paths and edge cases tested | Pass | Deadline hit, late rejection of an abandoned eval, settle-before-deadline, sync throw, disabled short-circuit, Node-side fake timers, and every `readinessDeadlineMs` branch. |
| Medium | Testing | Existing tests still pass (no regressions) | Pass | 48 of 50 checks green at review time (2 pending, 0 failures). Both `Test @percy/sdk-utils` jobs pass (2m18s and 3m34s — the Node and karma browser legs), and `Lint` passes. The karma pass also confirms the new spec's global-timer swap does not destabilise the browser runner. |
| Medium | Performance | Long-running tasks use background jobs | N/A | The change *shortens* a pathological wait. |
| Medium | Quality | Follows existing codebase patterns | Pass | Mirrored-constant + parity test copies the established `serialize-frames` precedent; the captured-timer pattern matches what jasmine-core itself does internally. |
| Medium | Quality | Changes are focused (single concern) | Pass | One concern, 3 files, additive only. |
| Low | Quality | Meaningful names, no dead code | Pass | `READINESS_DEADLINE_HIT` is module-local and unexported; no dead code. |
| Low | Quality | Comments explain why, not what | Pass | Rewritten in `07814bd` to state the capture's real precondition instead of implying a guarantee. |
| Low | Quality | No unnecessary dependencies added | Pass | No new dependencies — and deliberately no `node:timers` import, which would have put a Node builtin in the browser-bundle graph. |

## Findings

### Resolved since the last review

- ~~`packages/sdk-utils/src/serialize-dom.js:153` **Medium** — Node-side fake timers defeat the new deadline~~ **Resolved in `8787e64`** — the deadline now uses `globalThis.setTimeout`/`clearTimeout` captured at module load. Verified by emulating what jest/sinon do to the globals: before, the gate hung indefinitely and scheduled its deadline on the frozen clock; after, it returns `null` at the real 4s deadline and schedules nothing on the frozen clock. Residual risk (fake timers installed *before* this module is first evaluated — jest `fakeTimers.enableGlobally`, a `setupFiles` install, or `resetModules()` under an already-faked clock) is real, is not fixed by any import-order trick, and is now disclosed in the comment with `preset: disabled` named as the escape hatch.
- ~~`packages/sdk-utils/test/index.test.js` **Medium** — the fake-timer regression spec was unsound~~ **Resolved in `07814bd`** — raised in round 2. The spec replaced the global `setTimeout` then awaited the gate; against a real regression the gate never settles, so the `await` never returned, the `try/finally` never restored the globals, and every later spec using a bare `setTimeout` would hang — reporting one regression as a suite-wide cascade. It now races the gate against a rejection on the captured real timer at 6000ms: above the genuine 4000ms deadline so a healthy run resolves on its own, below jasmine's 10s `DEFAULT_TIMEOUT_INTERVAL` so this spec, not the runner, reports the failure. Confirmed by running the spec body against both module versions — pre-capture it now fails at 6003ms with a named error and the globals restored; post-capture it passes at 4002ms. All three `finally` paths verified.
- ~~`packages/sdk-utils/src/serialize-dom.js` **Low** — rationale comment overclaimed~~ **Resolved in `07814bd`** — "the one Node-only dependency" was wrong (`request.js` also imports `http`/`https` dynamically and `proxy.js` pulls `net`/`tls`), and the note now records that rollup suppresses `MISSING_NODE_BUILTINS`, so a static builtin import here would not fail the build — it would silently ship a browser bundle that breaks at runtime, which is a stronger justification than the original.

### Still open

#### F2 — `strict` preset deadline exceeds Playwright's default test timeout

- **File:** `packages/sdk-utils/src/serialize-dom.js:116`
- **Severity:** Medium (unchanged since `41ca4c0`)
- **Issue:** `strict` is 30s in-page → 33s deadline, above Playwright's default 30s per-test timeout. For a frozen clock under `strict`, the runner's own timeout fires first, so that individual test still fails with a Playwright timeout rather than degrading to `null`. The backstop still prevents an indefinite hang, but the comments imply a graceful save that does not hold for this tier.
- **Suggestion:** state the tradeoff explicitly, or make the grace non-additive above a threshold so the deadline stays under common runner defaults.
- **Status:** out of scope for this PR; not resolved by it. Worth a follow-up.

#### F3 — Deadline specs burn real wall-clock; grace has no test seam

- **File:** `packages/sdk-utils/test/index.test.js:112`
- **Severity:** Low (unchanged since `41ca4c0`)
- **Issue:** `READINESS_DEADLINE_GRACE_MS` is a bare module constant with no injection point, so every deadline spec must wait the full grace-inclusive 4000ms in real time. Round 3 adds one more such wait (the 6000ms bound, on the failure path only).
- **Suggestion:** give the grace a test seam and drive these specs with explicit tick advancement. Fixing this is what would let the fake-timer spec run in well under a second.
- **Status:** open, and mildly reinforced by this PR rather than worsened — the new spec follows the existing pattern rather than adding a new one.

---

**Verdict: PASS** — the two findings raised against earlier commits on this branch are confirmed resolved and independently re-verified; the two that remain are a Medium and a Low that predate this change's final form and are documented follow-ups, neither introduced nor worsened by it.

packages/sdk-utils/src/serialize-dom.js
rishigupta1599: Fixed in `8787e64` (capture) and `07814bd` (comment accuracy + a sound regression spec).

The deadline now uses `globalThis.setTimeout`/`clearTimeout` captured at module load. Verified by emulating what jest/sinon do to the globals (replace `setTimeout`, never advance): before, the gate hung indefinitely and scheduled its deadline on the frozen clock; after, it returns `null` at the real 4s deadline and schedules nothing on the frozen clock.

One correction to the suggestion in this thread: `import { setTimeout } from "node:timers"` would have been the wrong fix here. `serialize-dom.js` is imported statically by `src/index.js`, which is the rollup entry for the browser bundle (`rollup.config.js:111`), and the config suppresses `MISSING_NODE_BUILTINS` (`:39`) — so a static builtin import would not fail the build, it would ship a browser bundle that breaks at runtime. Node-only code in this package is always reached through a lazy `await import(...)` for exactly that reason.

Residual risk is real and now disclosed in the comment rather than papered over: fake timers installed *before* this module is first evaluated (jest `fakeTimers.enableGlobally`, a `setupFiles` install, or `resetModules()` under an already-faked clock) capture the fake, and no import-order trick fixes that — those consumers need `snapshot.readiness.preset: disabled`.

Resolving: the line this anchored to no longer exists in its original form.

packages/sdk-utils/src/serialize-dom.js
rishigupta1599: Leaving this one open deliberately — it is accurate and unaddressed.

`strict` is 30s in-page, so the deadline here is 33s, above Playwright's default 30s per-test timeout. Under `strict` with a genuinely frozen clock the runner's own timeout fires first, so that individual test still fails with a Playwright timeout rather than degrading to `null`. The backstop still prevents an *indefinite* hang, which is the part that matters for suite/CI health.

It predates this change and is not worsened by it, so it is called out in the PR description under "Scope note" rather than fixed here. Fixing it properly means either shrinking the grace for the `strict` tier or making it non-additive above a threshold — worth doing, but it changes behaviour for a tier nobody in this ticket is using, so it belongs in its own change.

Not resolving, so it stays visible to reviewers as a tracked follow-up.

ninadbstack (approved):

